PR #16, opened 2015-02-18
Add work in progress version to Composer's file

henriquemoody:
I have this process when I release some minor/major version:

1. Create a branch to version, like `x.y`
2. Create a tag to version, like `x.y.0`
3. Update the branch alias on `master` branch to the next minor version

It also makes easy to create fixes for older versions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,5 +32,10 @@
                 "test"
             ]
         }
+    },
+    "extra": {
+        "branch-alias": {
+            "dev-master": "2.1-dev"
+        }
     }
 }


henriquemoody:
BTW, I see no need to keep `master` and `develop` branches since you follow the process in the description - which IMHO is better for a library than GitFlow.

lcobucci:
I personally like GitFlow, but I'm open to sugestions.
Why is this better in your opinion?

henriquemoody:
Actually it depends of the support you give to older versions.

When you release `1.0.0` this version is the exact content of the `master` branch, then you will start the development of `1.1.0` (or `2.0.0`) on the `develop` branch.

If someone realizes there is a bug on `1.0.0` it's fine, the _pull request_ can be sent against `master`, you merge the _pull request_, release `1.0.0` and then merge `master` to `develop`.

Now, think you have `1.0.0`, `1.1.0` and `1.2.0` and someone realizes there is a bug on `1.0.0`. You first have to create a branch based on `1.0.0`, and no one can send a _pull request_ to fix this bug until you do that, then you have to tell the person what is the branch to send the _pull request_, merge the _pull request_, merge the branch you just created to `master` and merge `master` to `develop` branch and probably remove the branch you just created.

I think GitFlow works fine for applications which have just one stable version by time (in spite I think only `master` brach and _features branches_ or _pull requests_ is enough), but for a library the flow I recommend is better, IMHO.

henriquemoody:
Any way, you don't need to use this flow to accept this pull request, the flow was more a suggestion than other thing. With this change people can use `2.1.x-dev` instead of `dev-master` which is much better, again, IMHO.

lcobucci:
Ok, let's diccuss a little further =).

Why should I send a fix to ```1.0.0``` if I have ```1.2.0```? I mean, that's the meaning of semantic versioning (as I understand).

If one is using ```1.0.0``` he can just upgrade to ```1.2.0``` with no worries. If the bug persists he just send the PR to ```master``` (as it is the last stable version).

The catch here is when one is using ```1.0.0``` and we're working on ```2.1.0```. But I'm more inclined to don't offer updates on old major releases (wich I know is a very strict way to think).

The only thing I dislike on this change is having to change (on each major/minor release) the composer file and the branch aliases. Do you really think that this change is worth of the additional work?

henriquemoody:
> Why should I send a fix to `1.0.0` if I have `1.2.0`? I mean, that's the meaning of semantic versioning (as I understand).
> If one is using `1.0.0` he can just upgrade to `1.2.0` with no worries. If the bug persists he just send the PR to master (as it is the last stable version).

I don't think this is the meaning of Semantic Versioning but I won't go so deep at this.
New features are not always welcome; you can have a new feature that doesn't break API which is unnecessary for the developer, and still make the library slow (or faster, who knows?).
Also, a version is what it is: _just a number_; of course there are risks of API breaks in PATCH versions but even more on MINOR versions.

> The catch here is when one is using `1.0.0` and we're working on 2.1.0. But I'm more inclined to don't offer updates on old major releases (wich I know is a very strict way to think).

Not so strict, but giving I have `1.1.0`, `1.2.0` and then `2.0.0`, why should I remove support for `1.1.0` version? Of course it depends on my release cycle too.

> The only thing I dislike on this change is having to change (on each major/minor release) the composer file and the branch aliases. Do you really think that this change is worth of the additional work?

Actually that's what I do and I see many others doing. I just don't think if worth if you don't do the workflow I suggested.

lcobucci:
> Not so strict, but giving I have ```1.1.0```, ```1.2.0``` and then ```2.0.0```, why should I remove support for ```1.1.0``` version? Of course it depends on my release cycle too.

I believe that the proper question: why should we keep support for older releases? That's the strict POV :smile:

I think that the need of maintain old releases are applied on complex libs. The only reason that we have a ```2.x``` version here was the minor BC break when the ```Signer``` namespace was reorganized.

I understand your intent, but I believe that we would solve this issue having more opinions (now I'm not convinced heheh).

Let's hear the other contributors! Please @ppetermann and @jackmakiyama, can you share your thoughts on this?

We don't have a huge community yet, but I really wan't to try to do the best for everyone.

henriquemoody:
Despite the pull requests I've sent to this project I haven't made anything of high impact - I didn't change even one line of code on this project - so, to be honest, I don't know what was changed between the versions as well there are people who don't know too.

I'm wondering, if I don't want to fix the older versions why should I keep them? Why I don't use just the two branches instead? As I see, when a version has a bug it is not usable anymore, unless I give the proper support for that. That's why doesn't makes sense to me using versions and don't give support for the older ones, unless my intent is just to have a record of my releases. 

I think it's a good thing give support for some versions but not all, just like PHP does. (Actually I have to update the documentation of the projects I manage with the support information. :stuck_out_tongue_closed_eyes:)

Also, I've seen BC breaks between MINOR versions on many projects. Not everyone rely on Semantic Versioning, this project may probably follow it strictly but many others don't.

I like to use pinned versions but I use write my dependencies on "composer.json" file like `x.y.*` or `~x.y.z` and commit "composer.lock" file but often I update the dependencies in order to **avoid** problems. That's the beauty of PATCH versions (as I see) because I get only bug fixes, even if I was not experiencing this bug ~~yet~~, I won't have problems with this bug anymore.

So, according what was said, if the current version is `2.0.0` and I get a bug on `1.x.z` them I have to upgrade? It's really a way, but I don't know if is the best approach.

Any of the approaches works. I'm just clarifying my POV, whatever the choice you make it will be cool for me. :sunglasses:

ppetermann:
Hi. 

I believe git flow is just fine for libs. 

Now to address a few things said by you, henrique 

 * from what I understand the intention behind this is to have branch aliases for dev branches, so you can reference those in you software. This in most cases is bad practice, because it means you make going to production with your own software dependent on "we can't before OSS project xyz has not released" which is a nightmare for the planability of your own releases. Or worse, you end up going production with a dev reference - I think they have a special place in hell reserved for people who do that. 
 * you keep old tags around because someone might still be depending on those, and because it doesn't cost you anything. But you usually want a high adoption rate of new versions, to keep support efforts low, and your brain from being melted by the clusterfuck that back and forward porting of more and more versions tends to become. 
* the php project was mentioned - php and what happend with 5.3 is a good example on how harmful lts for old versions can be. As a lot of providers have not seen a reason to go to the next minor as long as there was support for 5.3 the PHP project extended the time to live on that version several times. Because they did so the providers had no reason to upgrade. The result? Horrible adoption rates of php 5.4, developers who can't use features that have been in PHP for years (because the software wouldn't run with the providers) and a lot of bug reports and backporting for things that where already fixed in later versions. Yay. 
 * today we see more and more people using semver, simply because it makes sense. It's a good and recommended praxis, and I try to avoid libraries which don't follow it. 
* putting a lib to 2.0 doesn't mean "go on use the old one" it's  to make sure stuff doesn't instantly break for you. But you should still adopt as fast as you can to profit from further fixes and keep the work load of those people giving you their work for free small. 


I hope my perspective is any help to you guys, 

Regards, 
PP. 

On February 25, 2015 11:07:42 PM GMT+01:00, Henrique Moody <notifications@github.com> wrote:
>Despite the pull requests I've sent to this project I haven't made
>anything of high impact - I didn't change even one line of code on this
>project - so, to be honest, I don't know what was changed between the
>versions as well there are people who don't know too.
>
>I'm wondering, if I don't want to fix the older versions why should I
>keep them? Why I don't use just the two branches instead? As I see,
>when a version has a bug it is not usable anymore, unless I give the
>proper support for that. That's why doesn't makes sense to me using
>versions and don't give support for the older ones, unless my intent is
>just to have a record of my releases. 
>
>I think it's a good thing give support for some versions but not all,
>just like PHP does. (Actually I have to update the documentation of the
>projects I manage with the support information.
>:stuck_out_tongue_closed_eyes:)
>
>Also, I've seen BC breaks between MINOR versions on many projects. Not
>everyone rely on Semantic Versioning, this project may probably follow
>it strictly but many others don't.
>
>I like to use pinned versions but I use write my dependencies on
>"composer.json" file like `x.y.*` or `~x.y.z` and commit
>"composer.lock" file but often I update the dependencies in order to
>**avoid** problems. That's the beauty of PATCH versions (as I see)
>because I get only bug fixes, even if I was not experiencing this bug
>~~yet~~, I won't have problems with this bug anymore.
>
>So, according what was said, if the current version is `2.0.0` and I
>get a bug on `1.x.z` them I have to upgrade? It's really a way, but I
>don't know if is the best approach.
>
>Any of the approaches works. I'm just clarifying my POV, whatever the
>choice you make it will be cool for me. :sunglasses: 
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/lcobucci/jwt/pull/16#issuecomment-76071149

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

henriquemoody:
I see no point on keeping this pull request opened since the maintainer doesn't need it.

Thanks guys, it was a great discussion!

PR #19, opened 2015-02-18
Remove unnecessary files from ignore list

henriquemoody:
The removed files must be in the global ignore list of developers who uses the IDE which creates these files.


henriquemoody:
Show I close this?

PR #24, opened 2015-06-09
Fixed hmac::verify

ptrofimov:
$expected could be not a string (namely false), when function base64_decode finds in the string invalid chars. In this case, function hash_equals triggers warning "hash_equals(): Expected known_string to be a string, boolean given".

This fix solves the problem.

### DIFF
--- a/src/Signer/Hmac.php
+++ b/src/Signer/Hmac.php
@@ -28,6 +28,10 @@ abstract class Hmac extends BaseSigner
      */
     public function verify($expected, $payload, $key)
     {
+        if (!is_string($expected)) {
+            return false;
+        }
+
         $callback = function_exists('hash_equals') ? 'hash_equals' : [$this, 'hashEquals'];
 
         return call_user_func($callback, $expected, $this->createHash($payload, $key));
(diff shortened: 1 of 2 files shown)


ptrofimov:
@lcobucci what do you think?

lcobucci:
Sorry @ptrofimov by the delay of the response and thanks for the collaboration.

ptrofimov:
Great! Thanks.

ptrofimov:
But could you, please, make a tag 2.1.2 for 2.1.* branch with this fix too? Because it will require time to change our project code for 3.* version.

lcobucci:
Ok, I'll create that tag.

On PR #16 @henriquemoody, @ppetermann and me discussed a little bit about maintaining older releases and the decision was to update only the last version, maybe we can review that with some new ideas.

lcobucci:
Done ;)

ptrofimov:
Super! Works. Thank you.

henriquemoody:
@lcobucci, perhaps you should create branches only for MAJOR versions, not for MINOR ones.

PR #46, opened 2015-10-09
Change comparison is empty.

AlefeVariani:
Change comparison is empty.

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -146,7 +146,7 @@ class Parser
      */
     protected function parseSignature(array $header, $data)
     {
-        if ($data == '' || !isset($header['alg']) || $header['alg'] == 'none') {
+        if (empty($data) || !isset($header['alg']) || $header['alg'] == 'none') {
             return null;
         }
 


Ocramius:
What's the reason for this change? Can you provide a test case?

AlefeVariani:
Sorry, this change is the [empty](http://php.net/manual/en/function.empty.php) function of php, in [Parse.php](https://github.com/lcobucci/jwt/blob/master/src/Parser.php) file.
> $data == ''
> empty($data) :+1:

Ocramius:
Yes, but why was it changed? Did you have a bug that needed this change?

AlefeVariani:
It's not a bug, just by existing empty function that does this :)

Ocramius:
empty is not the same of a string comparison...

lcobucci:
@AlefeVariani first of all thanks for your contribution :smile:.

What @Ocramius is saying is that the string comparison is good enough in this case. Did you found a bug that is solved after introducing the ```empty()``` function? If so please add test cases for that, otherwise I'll closing this PR.

AlefeVariani:
Ok, there's not bug.
Thanks @Ocramius and @lcobucci

PR #47, opened 2015-10-11
Add signer factory

### LINKED ISSUE #45: Signer factory?
I'm filing this issue mostly to feel out interest in and/or likelihood of acceptance of a simple factory implementation for `Signer` implementations if I were to submit a PR for one.

There's a [factory for Claims](https://github.com/lcobucci/jwt/blob/master/src/Claim/Factory.php), so it seems like there's some precedent for this.

This `Signer` factory would essentially resolve short algorithm names, such as [those used by the Firebase JWT implementation](https://github.com/firebase/php-jwt/blob/4a94b91434bf02395b4c2714ae19562116a80a10/src/JWT.php#L32-L37), to `Signer` instances, and allow users to add their own custom `Signer` implementations.

shadowhand:
Provides a standard interface for loading default and custom signers.

Fixes #45

cc @elazar

### DIFF
--- a/src/Signer/Factory.php
+++ b/src/Signer/Factory.php
@@ -1,0 +1,152 @@
+<?php
+/**
+ * This file is part of Lcobucci\JWT, a simple library to handle JWT and JWS
+ *
+ * @license http://opensource.org/licenses/BSD-3-Clause BSD-3-Clause
+ */
+
+namespace Lcobucci\JWT\Signer;
+
+use Lcobucci\JWT\Signer;
+
+/**
+ * Class that holds a registry of signers
+ *
+ * @author Woody Gilk <@shadowhand>
+ * @since 3.0.6
+ */
+class Factory
+{
+    /**
+     * The currently registered signers
+     *
+     * @var array
+     */
+    private $registry = [];
+
+    /**
+     * Prefixes used for namespaces
+     *
+     * @var array
+     */
+    private $prefixes = [
+        'ES' => 'Ecdsa',
+        'HS' => 'Hmac',
+        'RS' => 'Rsa',
+    ];
+
+    /**
+     * Initializes the factory, registering custom signers
+     *
+     * @param array $registry
+     * @param array $prefixes
+     */
+    public function __construct(array $registry = [], array $prefixes = [])
+    {
+        foreach ($registry as $signer) {
+            $this->register($signer);
+        }
+
+        $this->prefixes = array_replace($this->prefixes, $prefixes);
+    }
+
+    /**
+     * Register a signer
+     *
+     * @param Signer $signer
+     *
+     * @return self
+     */
+    public function register(Signer $signer)
+    {
+        $this->registry[$signer->getAlgorithmId()] = $signer;
+
+        return $this;
+    }
+
+    /**
+     * Get the namespace root for an algoritm id
+     *
+     * @param string $algoritm
+     *
+     * @return string
+     */
+    protected function getNamespace()
+    {
+        return __NAMESPACE__;
+    }
+
+    /**
+     * Get the namespace prefix for an algoritm id
+     *
+     * @param string $algoritm
+     *
+     * @return string|null
+     */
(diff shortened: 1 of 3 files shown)


src/Signer/Factory.php
```diff
@@ -0,0 +1,152 @@
...
+     *
+     * @param string $algoritm
+     *
+     * @return string|null
+     */
+    protected function getPrefix($algoritm)
+    {
+        $prefix = substr($algoritm, 0, 2);
+
+        if (empty($this->prefixes[$prefix])) {
```
shadowhand: This is slightly weird because the method is `protected` but `$prefixes` is private. The intent is that this method would be overloaded to do a custom prefix check and then use `parent::getPrefix($algorithm)` if necessary.

src/Signer/Factory.php
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Get the default class name for an algoritm id
+     *
+     * @param string $algoritm
+     *
+     * @return string
+     */
+    protected function getClass($algoritm)
+    {
+        return 'Sha' . substr($algoritm, 2);
```
shadowhand: Right now all of the classes are `ShaX` but I suppose in the future they could be different?

src/Signer/Factory.php
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Register a default signer by algorithm
+     *
+     * @throws \InvalidArgumentException if no signer can be identified
+     *
+     * @param string $id
+     *
+     * @return void
+     */
+    private function registerDefaultSigner($algoritm)
```
shadowhand: This is `private` to enforce the rules about namespacing.

test/unit/Signer/FactoryTest.php
```diff
@@ -0,0 +1,138 @@
...
+     * @uses Lcobucci\JWT\Signer\Ecdsa\Sha512::getAlgorithmId
+     * @uses Lcobucci\JWT\Signer\Ecdsa::__construct
+     * @uses Lcobucci\JWT\Signer\Ecdsa::createSignatureHash
+     * @uses Lcobucci\JWT\Signer\Ecdsa\KeyParser::__construct
+     * @uses Lcobucci\JWT\Signer\Hmac\Sha256::getAlgorithmId
+     * @uses Lcobucci\JWT\Signer\Hmac\Sha384::getAlgorithmId
+     * @uses Lcobucci\JWT\Signer\Hmac\Sha512::getAlgorithmId
+     * @uses Lcobucci\JWT\Signer\Rsa\Sha256::getAlgorithmId
+     * @uses Lcobucci\JWT\Signer\Rsa\Sha384::getAlgorithmId
+     * @uses Lcobucci\JWT\Signer\Rsa\Sha512::getAlgorithmId
```
shadowhand: I really dislike this huge `@uses` block... what's the reasoning for forcing the usage of `@covers` instead of allowing phpunit to determine the coverage?

test/unit/Signer/TestFixture/Sha256.php
```diff
@@ -0,0 +1,35 @@
+<?php
+/**
+ * This file is part of Lcobucci\JWT, a simple library to handle JWT and JWS
+ *
+ * @license http://opensource.org/licenses/BSD-3-Clause BSD-3-Clause
+ */
+
+namespace Lcobucci\JWT\Signer\TestFixture;
```
shadowhand: This class is pretty bogus but I wasn't sure how else to test custom prefixes.

lcobucci:
@shadowhand thanks for your contribution.

As said in #45 we're trying to avoid the creation of a factory for now. IMHO introducing it won't add value to the library itself and it's a lot better to rely in objects instead of a string convention (that's why I've removed it from v2 to v3).

I know that you're trying to remove this responsibility from sparkphp/auth#2 and I agree that you shouldn't do that in your lib. I'll try to suggest some ideas there so you'll be able to avoid that ;)

PR #58, opened 2015-12-03
Update Readme.md , use Key class and not KeyChain deprecated

psylo31:
Update Readme.md , use Key class and not KeyChain deprecated in v4 lib.


lcobucci:
Awesome @psylo31, thanks.

Can you just please name the following variables:

- $keyPrivate -> $privateKey
- $keyPublic -> $publicKey

I think it would be better :smile:

lcobucci:
Edited and merged as 60ce91bddac52f5637bfabc4a292b420cd599877

Thanks @psylo31 for your contribution!

Issue #49, opened 2015-10-11
Introduce configuration object


lcobucci:
As mentioned on #45 and #47 we're thinking a way to simplify the things by introducing a configuration object.

Ocramius:
@lcobucci I don't think this is better than the builder. (probably just provide the builder with better defaults and it's the same stuff).

It would be good to have the `Builder`, `Parser` and `Signer` to have in the config object. 

Token generation is already easy-peasy, but gluing everything together isn't really.

Ocramius:
I think it would be good to survey the possible use-case scenarios first, in order to see what can be simplified. I personally think the builder is fine, and very simple to use :-)

lcobucci:
Ok. I've edited to don't confuse anyone else.

PR #62, opened 2015-12-21
[WIP] Introducing a configuration object.

lcobucci:
@Ocramius that's was your idea with #49 right?


src/Configuration.php
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * @author Luís Otávio Cobucci Oblonczyk <lcobucci@gmail.com>
+ * @since 4.0.0
+ */
+final class Configuration
+{
+    /**
+     * @var array
+     */
+    private $data;
```
Ocramius: I suggest using properties instead of an array

lcobucci:
I'll create a new PR to the `master` branch

PR #63, opened 2015-12-31
Add functional tests to detect forged headers

paragonie-scott:
(Mostly to prevent regressions down the line.)


lcobucci:
Thanks again @paragonie-scott, you rock!

PR #64, opened 2015-12-31
Add security warning to README about signature algorithm selection.

paragonie-scott:
Related to #63, to prevent accidental misuse.


lcobucci:
That's a good reminder, thanks @paragonie-scott ;)

PR #68, opened 2016-01-14
use Lcobucci\JWT\Signer\Key

saltlakeryan:
For the code sample to work, must include 'use Lcobucci\JWT\Signer\Key;'
I also removed 'use Lcobucci\JWT\Signer\Keychain;' 
since Keychain is not used in the sample


lcobucci:
You're right, thanks! :beers:

saltlakeryan:
Thanks for the awesome library!

Issue #72, opened 2016-02-01
Token Validation API


schnittstabil:
In my opinion, the current Validation API can be greatly improved:

``` php
$data = new ValidationData();
…
$token->validate($data) // returns bool
```
1. returning `true` resp. `false` isn't very useful; returning a list of _constraint violations_ (i.e. Errors or similar) would be more helpful, e.g. like [Symfony and Java](https://github.com/symfony/validator) do.
2. token objects should not be responsible for validation; this violates many design principles (e.g.&nbsp;[SRP](https://en.wikipedia.org/wiki/Single_responsibility_principle)): _Changes_ in the validation process –like I suggest under item 1. – causes _changes_ in the `Token` class. IMO, using some `Validator` class would be a much better option.

lcobucci:
@schnittstabil some tradeoffs were made and when I built the `Token#validate()` I just need to know either if the token is valid or not (the violation list has not adding much value).

But you're right in your points and the next major release (v4) is a great opportunity to fix that. Could you send us a PR?

schnittstabil:
Sorry, I have no time for it, but I'll remember this issue when I'm less busy.

dannydorfel:
@lcobucci I have done some work on nr 2. Creating a validator class with a validate(Token $token, ValidationData $data). Will send in the PR this evening after fixing the tests.

I would also like to suggest the validator raises an Exception containing the errors. Since an invalid token could be potentially dangerous for the security of the system, the validate should not return false. The developer should be aware of the invalid token and be forced to act accordingly.

lcobucci:
@dannydorfel it would be awesome to see your PR. 

I left the decision of raising an exception to the user and I think it makes things simpler in some cases however I see the benefits of that. 

Some validation libraries offers different methods to handle that thing (ex: `validate` and `assert`). 

Right now I'm not too sure what would be the best way to go...

steverhoades:
I ran into a similar issue with wanting to validate the id token.  I've added my own token validator chain that seems to be working ok at the moment, still a WIP. 

https://github.com/steverhoades/oauth2-openid-connect-client/tree/master/src/Validator

I think @schnittstabil has a very good point about SRP here.  I'm happy to provide a PR if you see any value in the direction I have gone.

lcobucci:
@steverhoades we decided to go on [this direction](https://github.com/lcobucci/jwt/compare/refactor/validator-extraction).

It's also a WIP but I believe everything will be much easier.

[...]

Issue #29, opened 2015-08-31
Private/public claims validation


lcobucci:
The issue #28 remind me of that we actually don't have any way to validate private/public claims.

Of course this is an edge case because most of people will just "store things" in the token and are not expecting to validate a non-registered claim, but anyway this should probably be helpful for someone.

robksawyer:
Is it possible to use something like the following in validation? 

I'm still trying to wrap my head around the signing part of the token. (See my comment below.)
```
$data = new ValidationData();
$data->setIssuer($serverName);
$data->setAudience($serverName);
$data->setId($uid);
$data->setSigner(strval($my_secret_signer)); // <------- Is signer actually secret?
```

emilva:
+1 private/public claims validation would be very neat

jom:
+1 I'd especially like to be able to override just `exp` for adding a grace period when refreshing a [potentially a bit expired] JWT.

PR #73, opened 2016-02-04
Replaced Token->validate with Validator class handling token validation

dannydorfel:
@lcobucci I kept the responsibility for validating in the claim, since they are like value objects and know best how to validate themselves. I'm not sure where to put the error message. In the claim itself? 

Also documentation still needs update in this PR. 

Your comments please?

### DIFF
--- a/src/Claim/LesserOrEqualsTo.php
+++ b/src/Claim/LesserOrEqualsTo.php
@@ -10,7 +10,7 @@ declare(strict_types=1);
 namespace Lcobucci\JWT\Claim;
 
 use Lcobucci\JWT\Claim;
-use Lcobucci\JWT\Exception\InvalidClaimException;
+use Lcobucci\JWT\Exception\RangeException;
 use Lcobucci\JWT\ValidationData;
 
 /**
@@ -28,14 +28,7 @@ class LesserOrEqualsTo extends Basic implements Claim, Validatable
     {
         $name = $this->getName();
         if ($data->has($name) && ($this->getValue() > $data->get($name))) {
-            throw new InvalidClaimException(
-                sprintf(
-                    "The value of '%s' (%d) is not lesser than or equals the validation value (%d)",
-                    $name,
-                    $data->get($name),
-                    $this->getValue()
-                )
-            );
+            throw new RangeException($name, "The claim value should be less than the validation value");
         }
     }
 }
--- a/src/Claim/LesserOrEqualsTo.php
+++ b/src/Claim/LesserOrEqualsTo.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace Lcobucci\JWT\Claim;
 
 use Lcobucci\JWT\Claim;
+use Lcobucci\JWT\Exception\InvalidClaimException;
 use Lcobucci\JWT\ValidationData;
 
 /**
@@ -23,12 +24,18 @@ class LesserOrEqualsTo extends Basic implements Claim, Validatable
     /**
      * {@inheritdoc}
      */
-    public function validate(ValidationData $data): bool
+    public function validate(ValidationData $data)
     {
-        if ($data->has($this->getName())) {
-            return $this->getValue() <= $data->get($this->getName());
+        $name = $this->getName();
+        if ($data->has($name) && ($this->getValue() > $data->get($name))) {
+            throw new InvalidClaimException(
+                sprintf(
+                    "The value of '%s' (%d) is not lesser than or equals the claim value (%d)",
+                    $name,
+                    $data->get($name),
+                    $this->getValue()
+                )
+            );
         }
-
-        return true;
     }
 }
--- a/src/Claim/LesserOrEqualsTo.php
+++ b/src/Claim/LesserOrEqualsTo.php
@@ -30,7 +30,7 @@ class LesserOrEqualsTo extends Basic implements Claim, Validatable
         if ($data->has($name) && ($this->getValue() > $data->get($name))) {
             throw new InvalidClaimException(
                 sprintf(
-                    "The value of '%s' (%d) is not lesser than or equals the claim value (%d)",
+                    "The value of '%s' (%d) is not lesser than or equals the validation value (%d)",
                     $name,
                     $data->get($name),
                     $this->getValue()
--- a/src/Validation/Results.php
+++ b/src/Validation/Results.php
@@ -43,12 +43,4 @@ class Results implements ResultInterface
     {
         return $this->errors;
     }
-
-    /**
-     * @return bool
-     */
-    public function isExpired(): bool
-    {
-        return isset($this->errors['exp']);
-    }
 }
--- a/src/Validation/Results.php
+++ b/src/Validation/Results.php
@@ -8,6 +8,7 @@
 declare(strict_types=1);
 
 namespace Lcobucci\JWT\Validation;
+use Lcobucci\JWT\Exception\InvalidClaimException;
 
 /**
  * Results returned by validation
@@ -20,12 +21,11 @@ class Results implements ResultInterface
     private $errors = [];
 
     /**
-     * @param string $name
-     * @param string $message
+     * @param InvalidClaimException $exception
      */
-    public function addError(string $name, string $message)
+    public function addError(InvalidClaimException $exception)
     {
-        $this->errors[$name] = $message;
+        $this->errors[$exception->getClaim()] = $exception->getMessage();
     }
 
     /**
--- a/src/Validation/Results.php
+++ b/src/Validation/Results.php
@@ -12,7 +12,7 @@ namespace Lcobucci\JWT\Validation;
 /**
  * Results returned by validation
  */
-class Results implements ResultsInterface
+class Results implements ResultInterface
 {
     /**
      * @var array
@@ -31,7 +31,7 @@ class Results implements ResultsInterface
     /**
      * @inheritdoc
      */
-    public function valid(): bool
+    public function isValid(): bool
     {
         return empty($this->errors);
     }
@@ -39,7 +39,7 @@ class Results implements ResultsInterface
     /**
      * @inheritdoc
      */
-    public function errors(): array
+    public function getErrors(): array
     {
         return $this->errors;
     }
--- a/src/Validation/Results.php
+++ b/src/Validation/Results.php
@@ -1,0 +1,36 @@
+<?php
+namespace Lcobucci\JWT\Validation;
+
+class Results implements ResultsInterface
+{
+    private $errors = [];
+
+    /**
+     * @param string $name
+     * @param string $message
+     */
+    public function addError(string $name, string $message)
+    {
+        $this->errors[$name] = $message;
+    }
+
+    /** @inheritdoc */
+    public function valid(): bool
+    {
+        return empty($this->errors);
+    }
+
+    /** @inheritdoc */
+    public function errors(): array
+    {
+        return $this->errors;
+    }
+
+    /**
+     * @return bool
+     */
+    public function isExpired(): bool
+    {
+        return isset($this->errors['exp']);
+    }
+}
--- a/src/Validation/Results.php
+++ b/src/Validation/Results.php
@@ -1,8 +1,22 @@
 <?php
+/**
+ * This file is part of Lcobucci\JWT, a simple library to handle JWT and JWS
+ *
+ * @license http://opensource.org/licenses/BSD-3-Clause BSD-3-Clause
+ */
+
+declare(strict_types=1);
+
 namespace Lcobucci\JWT\Validation;
 
+/**
+ * Results returned by validation
+ */
 class Results implements ResultsInterface
 {
+    /**
+     * @var array
+     */
     private $errors = [];
 
     /**
@@ -14,13 +28,17 @@ class Results implements ResultsInterface
         $this->errors[$name] = $message;
     }
 
-    /** @inheritdoc */
+    /**
+     * @inheritdoc
+     */
     public function valid(): bool
     {
         return empty($this->errors);
     }
 
-    /** @inheritdoc */
+    /**
+     * @inheritdoc
+     */
     public function errors(): array
     {
         return $this->errors;
--- a/src/Validation/ResultsInterface.php
+++ b/src/Validation/ResultsInterface.php
@@ -1,9 +1,30 @@
 <?php
+/**
+ * This file is part of Lcobucci\JWT, a simple library to handle JWT and JWS
+ *
+ * @license http://opensource.org/licenses/BSD-3-Clause BSD-3-Clause
+ */
+
+declare(strict_types=1);
+
 namespace Lcobucci\JWT\Validation;
 
+/**
+ * Basic interface describing validation results
+ */
 interface ResultsInterface
 {
+    /**
+     * Returns true on valid results
+     *
+     * @return bool
+     */
     public function valid(): bool;
 
+    /**
+     * Returns an array with name => message
+     *
+     * @return array
+     */
     public function errors(): array;
 }
--- a/src/Validation/ResultsInterface.php
+++ b/src/Validation/ResultsInterface.php
@@ -1,0 +1,9 @@
+<?php
+namespace Lcobucci\JWT\Validation;
+
+interface ResultsInterface
+{
+    public function valid(): bool;
+
+    public function errors(): array;
+}
--- a/src/Validator.php
+++ b/src/Validator.php
@@ -1,0 +1,66 @@
+<?php
+/**
+ * This file is part of Lcobucci\JWT, a simple library to handle JWT and JWS
+ *
+ * @license http://opensource.org/licenses/BSD-3-Clause BSD-3-Clause
+ */
+
+namespace Lcobucci\JWT;
+
+use Generator;
+use Lcobucci\JWT\Claim\Validatable;
+
+/**
+ * Class that validates token with defined validationData
+ *
+ * @author Danny Dörfel <danny.dorfel@gmail.com>
+ */
+class Validator
+{
+    protected $errors = [];
+
+    /**
+     * @param Token $token
+     * @param ValidationData $data
+     * @return bool
+     */
+    public function validate(Token $token, ValidationData $data): bool
+    {
+        $this->errors = [];
+        foreach ($this->getValidatableClaims($token) as $claim) {
+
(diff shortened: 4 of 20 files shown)


KennedyTedesco:
Maybe the `validate()` method could be:

```
/**
 * @param Token $token
 * @param ValidationData $data
 * @return bool
 */
public function validate(Token $token, ValidationData $data): bool
{
    foreach ($this->getValidatableClaims($token) as $claim) {
        if ($data->has($claim->getName()) && ! $claim->validate($data)) {
            return false;
        }
    }

    return true;
}
```

KennedyTedesco:
Oops. Nevermind :smile: `getErrors()` explains everything.

dannydorfel:
Ah, I was already writing my explanation :-)

dannydorfel:
This PR is ment as a start to separate the validation concern from the token and add error messages. As mentioned in #72

KennedyTedesco:
Just another option:

```
/**
 * @param Token $token
 * @param ValidationData $data
 * @return bool
 */
public function validate(Token $token, ValidationData $data): bool
{
    $this->errors = [];

    foreach ($this->getValidatableClaims($token) as $claim) {
        if ($data->has($claim->getName()) && !$claim->validate($data)) {
            $this->errors[$claim->getName()] = false;
        }
    }

    return count($this->errors) === 0;
}
```

dannydorfel:
:+1: Ik like the improvement.

Op do 4 feb. 2016 19:53 schreef Kennedy Tedesco <notifications@github.com>:

lcobucci:
Awesome guys! I think we're going to a good way. 

After thinking a lot about those things I think that we can be a bit more radical :smile: : 

* get rid of the `Claim` objects (they are adding a bit of complexity that we may not need when removing this responsibility from the token);
* make the `Validator` a wrapper of an awesome validation library (like respect/Validation) 
* introduce a `ValidatorBuilder` to allow users to configure the values they want to use to create a validator (with a pre-configured set of constraints for the default claims) 

The benefits of that would be:

* simpler API (just use a validator instead of creating our own) 
* allowing the validation of private claims (on an easy way) #29

What you think?

lcobucci:
Maybe we could ping @henriquemoody or @alganet to get their input as well

src/Validator.php
```diff
@@ -0,0 +1,63 @@
...
+        $this->errors = [];
+
+        foreach ($this->getValidatableClaims($token) as $claim) {
+
+            if ($data->has($claim->getName()) && !$claim->validate($data)) {
+                $this->errors[$claim->getName()] = false;
+            }
+        }
+
+        return count($this->errors) === 0;
```
Ocramius: Can be simplified to `! $this->errors`

src/Validator.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * @return array
+     */
+    public function getErrors(): array
```
Ocramius: This API is kinda bad: stateful error storage. You should instead simply return a `ValidationResult` that encapsulates the coupled state of `valid + errors`

src/Validator.php
```diff
@@ -0,0 +1,63 @@
...
+use Lcobucci\JWT\Claim\Validatable;
+
+/**
+ * Class that validates token with defined validationData
+ *
+ * @author Danny Dörfel <danny.dorfel@gmail.com>
+ */
+class Validator
+{
+    protected $errors = [];
```
Ocramius: `private`

src/Validator.php
```diff
@@ -0,0 +1,63 @@
...
+
+use Generator;
+use Lcobucci\JWT\Claim\Validatable;
+
+/**
+ * Class that validates token with defined validationData
+ *
+ * @author Danny Dörfel <danny.dorfel@gmail.com>
+ */
+class Validator
```
Ocramius: Missing an interface for it?

src/Validator.php
```diff
@@ -0,0 +1,63 @@
...
+        return $this->errors;
+    }
+
+    /**
+     * Yields the validatable claims
+     *
+     * @param Token $token
+     * @return Generator
+     */
+    private function getValidatableClaims(Token $token)
```
Ocramius: Add ` : Generator` return hint

src/Validator.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * @return array
+     */
+    public function getErrors(): array
+    {
+        return $this->errors;
+    }
+
+    /**
+     * Yields the validatable claims
```
Ocramius: Docblock can be entirely skipped

Ocramius:
About @lcobucci's suggestions:

 > * get rid of the `Claim` objects (they are adding a bit of complexity that we may not need when removing this responsibility from the token);

Depends on whether they are exposed or not

 > * make the `Validator` a wrapper of an awesome validation library (like respect/Validation)

Sounds like overkill/implementation detail. The current validation is sufficient and very simple, IMO. The only thing to be fixed is the stateful validator (really bad!)

 > * introduce a `ValidatorBuilder` to allow users to configure the values they want to use to create a validator (with a pre-configured set of constraints for the default claims) 

In that case, having just a validator interface and letting the user implement the validator seems more reasonable for a first version, or else you might be getting out of the 80/20 use-cases. Don't overdo it, IMO.

src/Validator.php
dannydorfel: I would disagree. I rather be explicit and not use the convenience (flaws) of the PHP language.

src/Validator.php
dannydorfel: I agree. This is work in progress and minimal working version, not a final version. Do you have a link showing me how we could improve?

src/Validator.php
Ocramius: At this point, this is a local variable, and you can safely replace an `O(n)` with an `O(1)` operation.

src/Validator.php
Ocramius: at the end of `validate()`, `return new TokenValidationResult($token, $validationData, $success, $errors)`

src/Validator.php
```diff
@@ -0,0 +1,63 @@
...
+    protected $errors = [];
+
+    /**
+     * @param Token $token
+     * @param ValidationData $data
+     * @return bool
+     */
+    public function validate(Token $token, ValidationData $data): bool
+    {
+        $this->errors = [];
```
henriquemoody: I thought about creating a JWT rule on Respect\Validation sometimes. Of course we can do that but we need to write too much code from this library on our library then the chances of breaking the API is higher than using a single point of failure, that's why this feature is so useful. Also I don't thing it's fair to force someone to install one more library to have a fature that's simple for to have here.

About this implementation, IMHO, I would say that it can be better and the problem is on this line. If is there something I would get away from is to change the state of the object on each validation, and that what is happening here.

First of all, this method breaks SRP and secondly there is no need for such thing, not here. This method should only return `true` or `false` and that's all.

So, how can we get the errors and still not change the state of the object?

You can do what we do on Validation: Have a method to `assert()` the claims, throwing an exception when one claim or more are invalid then reuse the same method on `validate()` with a try/catch statement and return `true` if not exception is thrown or `false` if an exception is thrown, so you don't break DRY.

src/Validator.php
dannydorfel: You are right and I know very well it's possible to negate the array, but I would like to advocate being explicit. Using ! on a variable would normally imply a boolean, but in PHP you can use it on a lot of different situations. For me, that is more or less like goto: just because we have goto in php doesn't mean we have to use it. Considering, using ```return empty()``` would be the best choice since we know it's an array.

src/Validator.php
dannydorfel: :+1: Will start on that, hopefully this weekend

src/Validator.php
dannydorfel: :+1:

src/Validator.php
dannydorfel: Would be a good addition

src/Validator.php
Ocramius: `empty()` is also fine, yes. `! $arr` is perfectly fine in PHP semantics (explicit too, to be honest)

src/Validator.php
dannydorfel: :+1:

src/Validator.php
dannydorfel: :smile:

src/Validator.php
henriquemoody: I did not see @Ocramius' comment before my comment.

That would be nice as well :+1:

KennedyTedesco: Great! :+1:

lcobucci:
About @Ocramius and @henriquemoody comments:

> Depends on whether they are exposed or not

They're not exactly exposed. `Token#getHeaders()` and `Token#getClaims()` returns a list of them however it's way easier to just use `Token#getHeader($name)` and `Token#getClaim($name)` that hides the usage of `Claim`.

> Sounds like overkill/implementation detail. The current validation is sufficient and very simple, IMO. The only thing to be fixed is the stateful validator (really bad!)

> In that case, having just a validator interface and letting the user implement the validator seems more reasonable for a first version, or else you might be getting out of the 80/20 use-cases. Don't overdo it, IMO.

> I thought about creating a JWT rule on Respect\Validation sometimes. Of course we can do that but we need to write too much code from this library on our library then the chances of breaking the API is higher than using a single point of failure, that's why this feature is so useful. Also I don't thing it's fair to force someone to install one more library to have a fature that's simple for you to have here.

I think need to explain my suggestion better :smile:

The most important thing here (and I'm glad that we all agree on this HAHAHA) is to introduce an interface for the validator. In my mind it would be something like:

```php
namespace Lcobucci\JWT;

use Lcobucci\JWT\Validation\Result;

interface Validator
{
    public function validate(Token $token):Result;
}
```

Our default implementation would receive all the validation constraints (instances of `Respect\Validation\Validator`) and the `Validator#validate()` would just validate all the constraints (basically using the `Respect\Validation\Validator::allOf()`) and create a `Result` from that.

That means that we need an easy way to configure our validator's constraint list and that's where the builder enters. It would be the glue point with the chosen validation lib.

It's interface would be something like:

```php
namespace Lcobucci\JWT\Validation;

use Respect\Validation\Validator as Constraint;

class Builder
{
    public function assertId($id):Builder {}
    public function assertIssuer($issuer):Builder {}
    // the same for the other default claims
    public function assertThat($claim, Constraint $constraint):Builder {}
    public function getValidator():Validator {}
}
```

I'm just considering the idea of adding another dependency because we wouldn't have to deal here with the creation of error messages and also that we would allow users to validate the data in the token with some different validators.

Ocramius:
> deal here with the creation of error messages

Delegating this means less clear error messages (usually)

dannydorfel:
So we're clear on the interface (my time went 'poof' this weekend, so I'm in debt ;-))

How about keeping the assert in the claim, if you would consider a claim a value object. My preference is still to have the claim throw an exception. The validator can handle it and give a Result object back. Extracting it does feel a bit bloated.

As for the ```getValidator()``` in the builder: would not be my choice. The builder is a TokenBuilder and not a TokenValidatorBuilder, but I'm struggling with a better plan. Looks to me like the current builder class is already doing too much: building the headers, building the claims, building the token. So would it be an idea to extract the claims buildup in a separate class. That way we can reuse it in that object in the ```Validator::__construct(Claims $claims)```?

lcobucci:
> So we're clear on the interface (my time went 'poof' this weekend, so I'm in debt ;-))

Relax @dannydorfel, weekends are magical creatures that disappears sometimes =)

> How about keeping the assert in the claim, if you would consider a claim a value object. My preference is still to have the claim throw an exception. The validator can handle it and give a Result object back. Extracting it does feel a bit bloated.

I'm just suggesting to remove the claim objects because IMHO it's already a bit bloated heheh we're having to have a `ClaimFactory` inside of the `Builder` and inside of the `Parser` because of things related to the validation of each claim.

> As for the getValidator() in the builder: would not be my choice. The builder is a TokenBuilder and not a TokenValidatorBuilder, but I'm struggling with a better plan. Looks to me like the current builder class is already doing too much: building the headers, building the claims, building the token. So would it be an idea to extract the claims buildup in a separate class? That way we can reuse it in that object in the Validator::__construct(Claims $claims)?

If you look closely to my snippet you'll see that this `Builder` is inside other namespace and it's just related to validation :smiley:

dannydorfel:
A yes. But what if you can have a general ClaimBuilder for the claims. It's output is the same and can be used in the ```TokenBuilder``` and in the ```Validator```. Feels more SRP and decoupled than duplicate builders for token and validator. So extracting it from the builder makes it less bloated and still use the Claim as single value class that knows it's name and (if applicable) its validation.

And as I mentioned, I like the ValueObject-like setup. So removing the Claim object does not feel like an improvement. And I share @henriquemoody 's opinion on adding another library. You only have a few claim types, so throwing an Exception with correct message would suffice.

```php
namespace Lcobucci\JWT\Claim;

class Builder
{
    /** @var Collection */
    private $claims;
    private $factory;

    public function __construct(Factory $factory)
    { ... }

    public function setAudience(string $audience): Builder
    { ... }

    public function getClaims() : Collection
    {
        return $this->claims;
    }
        
}
```

```php
namespace Lcobucci\JWT;

use Lcobucci\JWT\Claim\Collection;

class Validator implements ValidatorInterface
{
    private $claims;

    public function __construct(Collection $claims)
    {
        $this->claims = $claims;
    }

    public function validate(Token $token): ValidateResult
    {
        $result = new ValidateResult();
        foreach ($this->claims as $claim) {
            .... try/catch, Claim with Validatable interface throws assert exception
        }

        return $result;
    }
}
```

```php
namespace Lcobucci\JWT;

use Lcobucci\JWT\Claim\Collection;

class Builder
{
    public function __construct(
        Collection $claims,
        Parsing\Encoder $encoder = null
    ) {
        ...
    }
}
```

dannydorfel:
@lcobucci @Ocramius @henriquemoody I changed the validator using interfaces and throwing exceptions. Please review ;-)

src/Validation/ResultsInterface.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Validation;
+
+/**
+ * Basic interface describing validation results
+ */
+interface ResultsInterface
```
henriquemoody: I would use `ResultInterface`.

src/Validation/ResultsInterface.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Returns true on valid results
+     *
+     * @return bool
+     */
+    public function valid(): bool;
```
henriquemoody: I would use `isValid()`.

src/Validation/ResultsInterface.php
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Returns an array with name => message
+     *
+     * @return array
+     */
+    public function errors(): array;
```
henriquemoody: I would use `getErrors()`.

src/Validation/Results.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * @var array
+     */
+    private $errors = [];
+
+    /**
+     * @param string $name
+     * @param string $message
+     */
+    public function addError(string $name, string $message)
```
henriquemoody: I think you should inject the exception here instead of defining these two arguments, you can also specialize the exception class.

src/Validation/Results.php
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public function errors(): array
+    {
+        return $this->errors;
+    }
+
+    /**
+     * @return bool
+     */
+    public function isExpired(): bool
```
henriquemoody: I don't thins it's necessary.

src/Claim/LesserOrEqualsTo.php
```diff
@@ -23,12 +24,18 @@ class LesserOrEqualsTo extends Basic implements Claim, Validatable
...
      */
-    public function validate(ValidationData $data): bool
+    public function validate(ValidationData $data)
     {
-        if ($data->has($this->getName())) {
-            return $this->getValue() <= $data->get($this->getName());
+        $name = $this->getName();
+        if ($data->has($name) && ($this->getValue() > $data->get($name))) {
+            throw new InvalidClaimException(
+                sprintf(
```
henriquemoody: You're repeating yourself with `sprintf($message, $name, $data->get($name), $this->getValue())`

dannydorfel:
So if have changed the exceptions for the claims after @henriquemoody 's review.
I specialised the exception class, replaced the name and message parameter in the Results class. And I also added the RangeException, extending the InvalidClaimException, defining a more specific exception for less/equal or greater/equal mismatch.

I hope I'm not asking too much, but am I on the right track here?

dannydorfel:
Just a reminder to keep things alive @lcobucci . No worries, I think this PR is not to be rushed. I'm actually expecting more refactor / review rounds.

lcobucci:
Thanks @dannydorfel for your patience! Since yesterday evening I finally have internet connection at my new home so things will be faster now :smile: 

About this PR, you were able to solve the SRP violation however it's way more complex than what I thought it would be. I was also hoping to solve the #29 with this extraction but I don't see this happening in the path we're going.

For me it doesn't matter if we're going to use a 3rd party lib or not, my idea is just to make everything simple and extensible. I just don't want you to think that I'm not considering the work you've done, I just think that the path I have in mind is easier to implement, removes some unnecessary complexity that we currently have and also solves the SRP problem - but I'm just unable to convert everything in words now =/.

I'm not sure if I should translate what I'm thinking into another PR and we could compare both paths and pick the best one or try to explain it here.

What's your suggestion guys?

henriquemoody:
I will be glad to review another pull request and discuss what's the best path to follow, @lcobucci. In spite of all the work that @dannydorfel has been done since then, I think it will not be a problem to have more options.

dannydorfel:
I'm actually glad to read that your looking for a simpler solution @lcobucci. Since Ocramius started with "having just a validator interface and letting the user implement the validator". I actually started my commits with only moving to the validator, removing the SRP issue. So it feels like a huge change and a lot the '80/20 use-cases' Ocramius talked about.

I wouldn't mind discussing a minimal code needed to remove the SRP and keeping the possibilities for #29 open? How about starting with only the ValidatorInterface and leaving the claims to returning false?

lcobucci:
Closing this in favor of #75.

PR #74, opened 2016-02-23
Introducing a configuration object

lcobucci:
Anything else @Ocramius?

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -1,0 +1,126 @@
+<?php
+/**
+ * This file is part of Lcobucci\JWT, a simple library to handle JWT and JWS
+ *
+ * @license http://opensource.org/licenses/BSD-3-Clause BSD-3-Clause
+ */
+
+declare(strict_types=1);
+
+namespace Lcobucci\JWT;
+
+use Lcobucci\Jose\Parsing;
+use Lcobucci\JWT\Claim\Factory as ClaimFactory;
+use Lcobucci\JWT\Signer\Hmac\Sha256;
+
+/**
+ * Configuration container for the JWT Builder and Parser
+ *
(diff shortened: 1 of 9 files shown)


src/Configuration.php
```diff
@@ -0,0 +1,123 @@
...
+
+    public function setClaimFactory(ClaimFactory $claimFactory)
+    {
+        $this->claimFactory = $claimFactory;
+    }
+
+    private function getEncoder(): Parsing\Encoder
+    {
+        if ($this->encoder === null) {
+            $this->encoder = new Parsing\Parser();
```
henriquemoody: typo `Parsing\Encoder`

henriquemoody: I just saw that `Parser` implements `Encoder`, sorry

src/Configuration.php
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * @author Luís Otávio Cobucci Oblonczyk <lcobucci@gmail.com>
+ * @since 4.0.0
+ */
+final class Configuration
+{
+    /**
+     * @var Parser
+     */
+    private $parser;
```
Ocramius: These values are potentially also `|null`

src/Configuration.php
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    public function setEncoder(Parsing\Encoder $encoder)
+    {
+        $this->encoder = $encoder;
+    }
+
+    private function getDecoder(): Parsing\Decoder
+    {
+        if ($this->decoder === null) {
```
Ocramius: Yoda your friend may be

src/Configuration.php
lcobucci: indeed

PR #75, opened 2016-02-24
[POC] Implementing a different path to the validator.

lcobucci:
@Ocramius, @henriquemoody and @dannydorfel.

I've tried to translate my thoughts into code so we can discuss the best approach to solve #72 (this or #73). I didn't fixed the tests, my idea was to quickly have something for you and then we can evolve.

There are some points to improve, but we can deal with them after choosing our destiny :smile:


src/Validation/Constraint/BaseConstraint.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        if ($this->isValid($claim)) {
+            return;
+        }
+
+        throw new ConstraintViolationException($this->getMessage($claim));
+    }
+
+    public abstract function isValid(Claim $claim): bool;
+    public abstract function getMessage(Claim $claim): string;
```
henriquemoody: I would consider having `isValid()` and `getMessage()` if they were `protected`, not `public`. However they are not necessary, IMO. You're not using them on other places and I don't see much use for them on another context.

I would probably rename `Constraint` to `Assertion`, `validate()` to `assert()` and just throw the exception when invalid, then you keep `Validator::validate()` as is.

src/Validation/Constraint/GreaterOrEqualsTo.php
```diff
@@ -0,0 +1,41 @@
...
+        return $claim->getValue() >= $this->value;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getMessage(Claim $claim)
+    {
+        return strpos(
+            'Claim %s expected to be greater or equals to %s, %s given',
```
henriquemoody: `sprint()`

src/Validation/Constraint/GreaterOrEqualsTo.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        return $claim->getValue() >= $this->value;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getMessage(Claim $claim)
+    {
+        return strpos(
```
henriquemoody: `sprint()`

src/Validation/Constraint/GreaterOrEqualsTo.php
henriquemoody: I don't feel much comfortable about the way messages have been created no this implementation because you're repeating yourself (`sprintf($format, $claim->getName(), $this->value, $claim->getValue())`) but I cannot thing about something simpler and I would rather stick with it than add more complexity to the code.

src/Validation/Constraint/BaseConstraint.php
dannydorfel: I agree with henriquemoody, going towards assertion and just throwing exceptions feels simpler. And removes the need for `isValid()` and `getMessage()`. Might even remove the need for this abstract class?

lcobucci: It's a good suggestion to rename it to `Assertion` and `assert()`.

I just tried to don't repeat code, the methods are public just to make them testable - but of course this could be done by calling the `assert()` method.

lcobucci:
Guys, I appreciate your comments about the naming and also the methods but I was expecting a bit more so we could choose the simplest path to solve the SRP violation, make the codebase a bit more cohesive and also fix the #29 (this or #73) :smile:.

Thanks!

dannydorfel:
I was not only referring to naming, but also to implementation. I think I was a little too simple in my reply ;-)

What I ment was: create a specific exceptions extending the `ConstraintViolationException` for the `LesserOrEqualsTo` and `GreaterOrEqualsTo` constraints. And simplify the constraint with only an `assert()`. For example the LesserOrEqualsTo constraint.

```
final class LesserOrEqualsTo implements Constraint
{
    /**
     * {@inheritdoc}
     */
    public function assert(Claim $claim)
    {
        if ($claim->getValue() > $this->value) {
            throw new ClaimNotLesserOrEqualException($this->createExceptionMessage($claim));
        }
    }

    private function createExceptionMessage(Claim $claim)
    {
        return sprintf(
            'Claim %s expected to be lesser or equals to %s, %s given',
            $claim->getName(),
            $this->value,
            $claim->getValue()
        );
    }
}
```

The `ClaimNotLesserOrEqualException` was a quick pick, might need improvement. But I hope you understand my point. It also remove the Abstract BaseConstraint.

An I know this does not solve the duplication of createExceptionMessage @henriquemoody. But changing it like this, I believe there is no real duplication. Only specific behaviour within this class.

lcobucci:
> I was not only referring to naming, but also to implementation. I think I was a little too simple in my reply ;-)

I was talking about the approach :smile: 

Personally I prefer this path over creating claims with different types that will serve just for validation concerns. This simplifies the `Builder`, the `Parser` and also removes the `ClaimFactory`. However, since you put effort to improving this lib, I also want to hear your opinion on that.

> What I meant was: create a specific exceptions extending the ConstraintViolationException for the LesserOrEqualsTo and GreaterOrEqualsTo constraints. And simplify the constraint with only an assert(). For example the LesserOrEqualsTo constraint.

Yes I understood :smile:. But I believe that we should just create those kind of exceptions if we really need them, I mean, why would be useful to differentiate each exception by the type on the validation result object? I can clearly see the benefit of that for a validation library, but our need here may be just know if the token has any problem and display list of those problems.

dannydorfel:
> Personally I prefer this path over creating claims with different types that will serve just for validation concerns. This simplifies the Builder, the Parser and also removes the ClaimFactory. However, since you put effort to improving this lib, I also want to hear your opinion on that.

I like your simplification. As I mentioned earlier, it has become bloated with extra wishes. I think we can close my PR?

> Yes I understood :smile:. But I believe that we should just create those kind of exceptions if we really need them, I mean, why would be useful to differentiate each exception by the type on the validation result object? I can clearly see the benefit of that for a validation library, but our need here may be just know if the token has any problem and display list of those problems.

My only thought there are to simplify the class, make the BaseClass obsolete and only build on the Interface. The `ClaimNotLesserOrEqualException` was an extra (since I often use such structures and catch a specific exception if needed). The `ConstraintViolationException` is already good enough for me.

henriquemoody:
The way I see, the messages here are not as important as I thought; the `Result` object is enough to known if there are violations or not and who are they. If someone whats to create messages they can just iterate over the results.

I believe that the `Constraint` interface can just return a boolean value on `validate()` method (which in this context would be better called `isValid()`), example:

```php
interface Constraint
{
    public function isValid(Claim $claim): bool;
}
```

Then the `Result` object is going to accept everything (violations or not):

```php
final class Result
{
    private $results;

    public function __construct(array $results)
    {
        $this->results = $results;
    }

    public function getResults(): array
    {
        return $this->results;
    }

    public function isValid()
    {
        $success = array_filter($this->results);

        return !empty($success);
    }
}
```

And finally the `validate()` method on the `Validator` will look like this:

```php
public function validate(Token $token): Result
{
    $results = [];
    $claims = $token->getClaims();

    foreach ($this->constraints as $claim => $constraint) {
        if (!array_key_exists($claim, $claims)) {
            continue;
        }

        $results[$claim] = $constraint->isValid($claims[$claim]);
    }

    return new Result($results);
}
```

It is simpler and provides more information for the client. As I said, if someone wants for create messages/exceptions the only thing one must do is to iterate over `Result::getResults()`.

***
In this context I think `Result` will be better called `ResultSet`.

lcobucci:
After a looooooong time away from OSS, I'm back!

But before jumping back into the code I would like to ask you guys one extra thing (YAY 😄): what do you think about making the assertions/constraints as mandatory or at least allow the user to mark it as optional instead of just ignoring the claims that are not in the token? I think that would be really helpful to ensure the token consistency.

lcobucci:
I'm changing this here and will close this PR and open a new one when it's done 😉

PR #76, opened 2016-02-28
feature/BetterClaimValidation - allow arrays for audience claim and issuer validation.

SamThePsychoticLeprechaun:
I have made changes that allow this JWT solution to follow more closely RFC7519 section 4.1.3 which allows for multiple audience members in the 'aud' claim via an array of stringOrUri values. Also follows the implications in section 4.1.7 of the same RFC which suggests that multiple issuers may be possible and acceptable for the same token - given the 'jti' claim can (should?) instead be used to differentiate tokens in circumstances where a set of issuers are considered equal.

The changes allow for uses like the following:

```php
$token = (new Builder())->setIssuer("http://example.com")
                        ->setAudience(["http://example.org", "http://example.com", "http://example.co.uk"])
                        ->setIssuedAt(time())
                        ->setNotBefore(time() + 60)
                        ->setExpiration(time() + 3600)
                        ->getToken();

$data = new ValidationData();
$data->setIssuer(["http://example.com", "http://example.org"]);
$data->setAudience("http://example.org");

var_dump($token->validate($data)); // True as the specified audience in ValidationData exists as a member of the 'aud' claim.
                                   // Also the 'iss' claim value is one of the values passed as allowed issuers to the ValidationData object.
```

However no backwards compatibility is broken, and the old uses of ```setIssuer ``` in ```ValidationData``` and ```setAudience``` in ```Builder``` are allowed.

One issue is that it in the current system ```setIssuer``` would no longer be correctly named given multiple issuers are possible.


SamThePsychoticLeprechaun:
I realise this may be the wrong branch, however I wasn't sure which to make the PR into.

lcobucci:
Hey @SamThePsychoticLeprechaun, thanks for your contribution! It's indeed very helpful :smile: 

I just believe that it breaks the BC since `$token->getClaim('aud')` will always return an array instead of a string, right? Given that we don't know how people are using the library we can't assume that they are not expecting a string from `$token->getClaim('aud')`.

Can you please squash your commits, sync your branch with `master` and send a new PR?

SamThePsychoticLeprechaun:
Ah, of course, you're quite right. Will do!

PR #83, opened 2016-04-15
Don't use generators

marco-c:
As far as I can tell, this is the only thing making the library incompatible with PHP 5.4.

Being compatible with PHP 5.4 would be important to make the library usable in WordPress plugins, see for example: https://wordpress.org/support/topic/svn-commit-fails-because-of-yield-keyword.

Also, since generators aren't widely used in the 3.1 branch (just once), I don't see the advantage in depending on them.

### DIFF
--- a/src/Token.php
+++ b/src/Token.php
@@ -224,11 +224,14 @@ class Token
      */
     private function getValidatableClaims()
     {
+        $validatableClaims = array();
         foreach ($this->claims as $claim) {
             if ($claim instanceof Validatable) {
-                yield $claim;
+                $validatableClaims[] = $claim;
             }
         }
+
+        return $validatableClaims;
     }
 
     /**


Ocramius:
This library requires [`"php": ">=5.5"`](https://github.com/lcobucci/jwt/blob/afea8e682e911a21574fd8519321b32522fa25b5/composer.json#L20) in its 3.1 version. Further versions will require PHP 7.0 or more.

Downgrading the requirement to a PHP version that is not receiving security patches anymore is out of question.

marco-c:
I appreciate you wanting to push users to use updated PHP versions, but I feel like it should be ultimately their decision (since the users of your library are developers, they can take decisions with awareness).

The composer requirements are supposed to be the minimum set of packages that the library strictly depends on, you could use the `suggest` field to suggest users to update to PHP 5.5.

Ocramius:
@marco-c the composer requirements are a strict requirement: not meeting them means that you can't install the library. If you install the library manually and ignore the requirements, then that is not up to the library maintainer to fix.

 > but I feel like it should be ultimately their decision

Lib maintainer decision. As far as I can see, this library has always been for PHP >= 5.5.

marco-c:
> @marco-c the composer requirements are a strict requirement: not meeting them means that you can't install the library. If you install the library manually and ignore the requirements, then that is not up to the library maintainer to fix.

I'm not asking the library maintainer to fix, I have proposed a fix which would make it available for PHP 5.4 at no cost for the maintainer.

> Lib maintainer decision. As far as I can see, this library has always been for PHP >= 5.5.

I understand that it's always been for PHP >= 5.5. I'm proposing to change the requirement via this simple fix. The library doesn't really need PHP 5.5, so why force the developers using this library to use PHP 5.5?

Ocramius:
> so why force the developers using this library to use PHP 5.5?

Mainly because [5.4 is not maintained anymore in first place](http://php.net/supported-versions.php), and we won't be able to fix bugs for those versions anyway. Also because there is no interest to maintain software for older versions when building a project for fun.

 > why force the developers using this library

This is my personal opinion, and has nothing to do with what @lcobucci thinks: in general, you are using free software written by someone for fun and for HIS/HER needs, not for YOUR needs. An OSS library is a free piece of code [given out at no cost for you](https://github.com/lcobucci/jwt/blob/397d8f51e3e12f9813d3f7ac7b77eeae6ac1af98/LICENSE), but with no responsibility from the author, so you should not consider yourself a "customer", and instead you should align to the requirements of the package, if you want to use it. Nobody is forcing you to do anything at all: you installed this library in first place, knowing its requirements. (as you can see, lots of "you" when it comes to OSS)

marco-c:
> Mainly because 5.4 is not maintained anymore in first place, and we won't be able to fix bugs for those versions anyway. Also because there is no interest to maintain software for older versions when building a project for fun.

Well, the 3.1 branch is in maintenance mode, isn't it? You're not going to introduce new features in this branch, or are you? So, it's unlikely you're going to need PHP 5.5 specific features (indeed, you already don't need them, apart from this one generator).

> This is my personal opinion, and has nothing to do with what @lcobucci thinks: in general, you are using free software written by someone for fun and for HIS/HER needs, not for YOUR needs. An OSS library is a free piece of code given out at no cost for you, but with no responsibility from the author. You should not consider yourself a "customer", and instead you should align to the requirements of the package, if you want to use it. Nobody is forcing you to do anything at all: you installed this library in first place, knowing its requirements. (as you can see, lots of "you" when it comes to OSS)

Well, in the full spirit of OSS, I was trying to bring my contribution to a project that I use.
I was just trying to convince you that the requirements could be relaxed, but I don't want to force you to do anything. I'll create my fork and use that.

Ocramius:
> You're not going to introduce new features in this branch, or are you? 

Correct, we're not introducing new features in 3.1. Support for 5.4 is also to be considered a feature though.

 > I'll create my fork and use that.

That is indeed viable in such case, thanks for your understanding :-)

lcobucci:
@marco-c I really appreciate the fact you're using and contributing to this library. You're absolutely right about not HAVING to depend on `php >= 5.5` (the only change is that method and all tests that uses `::class`) but the fact is that I really DECIDED to have that dependency because I think that we all should always rely on the last stable version (that's why v4.0 will require PHP 7).

That kind of decision is not so popular and I'm totally aware that it may keep people from using things I create, but I also see OSS as a way to express my beliefs and principles.

Thanks a lot for defending your vision and understanding mine (ours).
Please do use your fork and enjoy the amazing world of JWT 😄

marco-c:
Thank you for creating the library, it made my life much easier!

Actually, creating the fork also allowed me to remove the dependency on OpenSSL (since I don't need RSA), so there was also another benefit. I'll try to keep it up to date with your 3.1 branch, hopefully it shouldn't be too hard since you're not adding new features.

PR #87, opened 2016-05-24
Update required PHP version

Nyholm:
Since https://github.com/lcobucci/jwt/commit/cfd03d35d42a5895823ba28a3a91614199191861 we do no longer support php 5.5


Ocramius:
👍

Nyholm:
Thank you for merging

lcobucci:
Actually this is not fully correct 😄 on v3.x we still support PHP 5.5+.
The PHP 7 requirement is just for the next major release (`master` branch while is unstable).

The statement on https://github.com/lcobucci/jwt/blob/master/README.md#dependencies is correct so maybe we don't need that duplicated information.

lcobucci:
But thanks @Nyholm for your contribution

lcobucci:
I'll update that on #74 😉

Ocramius:
Argh, sorry for the over-eager merge then :(

lcobucci:
@Ocramius relax, no harm done

PR #107, opened 2016-08-04
Key might also be hosted on a different protocol than file://

frederikbosch:


lcobucci:
Hey @frederikbosch thanks for your contribution! Can you also create a test for that change please?

lcobucci:
@frederikbosch I was thinking about this PR... when we implement #32 we would already be able to use different protocols. So it seems that this a `wontfix` for me, do you agree?

lcobucci:
Closing here since had no response and it seems that implementing JWK will solve this.

Issue #106, opened 2016-07-27
Fatal error: Uncaught TypeError: Argument 1 passed to Lcobucci\JWT\Signer\Ecdsa\KeyParse


craigcarnell:
```
Fatal error: Uncaught TypeError: Argument 1 passed to Lcobucci\JWT\Signer\Ecdsa\KeyParser::__construct() must be an instance of Mdanter\Ecc\Math\MathAdapterInterface, instance of Mdanter\Ecc\Math\GmpMath given, called in vendor/lcobucci/jwt/src/Signer/Ecdsa.php on line 50 and defined in vendor/lcobucci/jwt/src/Signer/Ecdsa/KeyParser.php:43
```

lcobucci:
@craigcarnell some questions for you 😄:

- what version of `lcobucci/jwt` and `mdanter/ecc` are you using? 
- how did you installed the it?

craigcarnell:
@lcobucci jwt is master, mdanter/ecc is 0.4.0

It looks like in 0.4.0 they removed the class. If I use the last release 0.3.1 it works

lcobucci:
Thanks!

I'll fix the `0.3.1` on our stable version (3.1) and schedule the compatibility with `0.4.0` to our development version (4.0).

Please make sure you don't use our unstable/development version on production because it will change without any notice.

PR #109, opened 2016-08-08
Enable checking to see if tokens are expired

shadowhand:
Being able to check if a token is expired without checking other claims
is sometimes necessary, eg when doing token refreshing.

Refs #108

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -5,7 +5,7 @@ build:
         redis: false
         rabbitmq: false
         php:
-            version: 7.0
+            version: 7.0.8
 tools:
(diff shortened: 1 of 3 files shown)


shadowhand:
@lcobucci is this what you had in mind? It seems that `master` is the 4.0 branch. Should I also make a PR against `3.1`?

shadowhand:
@lcobucci build failures are not anything I changed...

lcobucci:
@shadowhand thanks for your contribution, don't worry about `master` I'm fixing build (it's related to #106) and I'll apply your changes there manually.

PR #115, opened 2016-09-08
Let isExpired() accept DateTimeInterface instead of just DateTime

DASPRiD:


lcobucci:
@DASPRiD  cool, thanks!

DASPRiD:
@lcobucci overall it would be nice if the entire token system would follow a more immutable approach, but that would be kind of a BC break :)

lcobucci:
@DASPRiD the `Token` class is meant to be immutable, if you find something that it's not please open a PR towards `master`.

We're developing the v4 in that branch, so it's not a big deal to have BC breaking changes 😃

PR #121, opened 2016-09-30
Just try to send the report when it exists

lcobucci:


Ocramius:
@lcobucci dang, too fast :D

lcobucci:
HAHAH

lcobucci:
@Ocramius btw we have to do the same on the storageless repo

PR #122, opened 2016-10-01
Simplify signer API

lcobucci:
This is part of a bigger refactoring which I'm working on that's basically creating a 1st class collection for the token claims and headers, improve the encapsulation and also remove things that are not really needed.


src/Signature.php
```diff
@@ -40,11 +42,11 @@ public function __construct(string $hash)
      *
      * @param Signer $signer
      * @param string $payload
-     * @param Key|string $key
+     * @param Key $key
      *
      * @return bool
      */
-    public function verify(Signer $signer, string $payload, $key): bool
+    public function verify(Signer $signer, string $payload, Key $key): bool
```
Ocramius: lqklotb

src/Signer.php
```diff
@@ -29,33 +29,31 @@ public function getAlgorithmId(): string;
 
     /**
      * Apply changes on headers according with algorithm
```
Ocramius: This interface method doesn't say anything to me (reader)

src/Signer.php
```diff
@@ -29,33 +29,31 @@ public function getAlgorithmId(): string;
...
-     * @param Key|string $key
+     * @param Key $key
      *
-     * @return Signature
+     * @return string
      *
      * @throws InvalidArgumentException When given key is invalid
      */
-    public function sign(string $payload, $key): Signature;
+    public function sign(string $payload, Key $key): string;
```
Ocramius: Don't like. What's your rationale for moving away from a `Signature` object?

src/Signature.php
Ocramius: Unrelated: does `Key` verify that a non-empty string is given?

src/Builder.php
```diff
@@ -39,7 +39,7 @@ class Builder
     /**
      * The token signature
      *
-     * @var Signature
+     * @var string
```
lcobucci: Yeap, you're. Thx

src/Builder.php
```diff
@@ -83,9 +83,8 @@ public function __construct(
      */
     public function canOnlyBeUsedBy(string $audience, bool $addHeader = false): Builder
     {
-        $audiences = isset($this->claims['aud'])
-            ? array_merge($this->claims['aud']->getValue(), [$audience])
-            : [$audience];
+        $audiences = isset($this->claims['aud']) ? $this->claims['aud']->getValue() : [];
```
lcobucci: Not yet due to `$this->claims['aud']->getValue()` but next PR will use it 😉

src/Signature.php
lcobucci: > Unrelated: does Key verify that a non-empty string is given?

Nope, it's a good thing to add thoug

src/Signer.php
lcobucci: Indeed, it can actually be removed. Will do it

lcobucci: Removed, basically it was an over engineering. We may have signers which changes other headers (rather than `alg`) but we don't have them so far...

Thanks

src/Signer.php
lcobucci: The idea is to keep the `Signature` object related only to the `Token`. I started this on https://github.com/lcobucci/jwt/compare/refactor/validator-extraction?expand=1

Ocramius: @lcobucci can this branch be merged into that one? I need to see the bigger picture here, because this change is a regression (worse API, from my PoV)

lcobucci: Sure, let's do it then. I'll work on it and open the PR when it's ready

Issue #128, opened 2016-10-21
Use library specific exceptions


lcobucci:
As pointed by @renatomefi we're relying on SPL exceptions and would be really good to have an exception tree for this library in order to simplify things while catching exceptions.

lcobucci:
Also adding named constructors where is applicable.

Issue #117, opened 2016-09-30
Improve the token interface


lcobucci:
In order to provide a better encapsulation for the token we should provide methods to:

- [x] check if token can already be used (`nbf` - not before - claim is <= now)
- [x] check if token was issued before a given time (`iat` claim - issued at - claim is <= now)
- [x] check if token was issued by one of a list of issuers (`iss` - issuer - claim)
- [x] check if token can be used by a client (`aud` - audience - claim)
- [x] check if token is related to a certain subject (`sub` - subject - claim)

Issue #113, opened 2016-08-14
Improve token builder interface


lcobucci:
The `Lcobucci\JWT\Builder` interface is not so fluent as it could be...

I'm thinking that this would be better:

```php
class Builder
{
    public function withId(string $id, bool $addHeader = false): Builder; // replace 'jti'
    public function relatedTo(string $subject, bool $addHeader = false): Builder; // replace 'sub'
    public function issuedBy(string $issuer, bool $addHeader = false): Builder; // replace 'iss'
    public function canOnlyBeUsedBy(string $audience, bool $addHeader = false): Builder; // append 'aud'

    public function canOnlyBeUsedAfter(int $notBefore, bool $addHeader = false): Builder; // replace 'nbf'
    public function expiresAt(int $expiration, bool $addHeader = false): Builder; // replace 'exp'
    public function issuedAt(int $issuedAt, bool $addHeader = false): Builder; // replace 'iat'

    public function withHeader(string $name, $value): Builder;
    public function with(string $name, $value): Builder;

    public function sign(Signer $signer, Key $key): Builder;
    public function unsign(): Builder;

    public function getToken(): Token;
}
```

Any suggestions?

alexbilbie:
I really like this (I'm using your library in `league/oauth2-server`). 👍 from me.

MarijnKoesen:
Working on it 👍

Issue #44, opened 2015-10-05
Ensure that claim content is json serializable 


lcobucci:

lcobucci:
I've decided not to do this. I mean, people have been using this library for years already and haven't reported issues. Encoding the claim and decoding just to validate stuff seems overkill for this library.

PR #129, opened 2016-10-24
Create new validation API

lcobucci:
Yes @Ocramius @henriquemoody @schnittstabil @dannydorfel, after a looong time we finally have something to solve the validation API 🎉 

Brace yourself because this is huge PR but after merging it we'll haz:
- interfaces for the things that the library exposes
- fix for the SRP violation
- private claim validation (custom constraints)


lcobucci:
There're things that can still be improved (especially naming) and still need to update the documentation, but would be amazing to have your comments!

src/Storage/DataSet.php
```diff
@@ -0,0 +1,64 @@
...
+ * This file is part of Lcobucci\JWT, a simple library to handle JWT and JWS
+ *
+ * @license http://opensource.org/licenses/BSD-3-Clause BSD-3-Clause
+ */
+
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Storage;
+
+use OutOfBoundsException;
```
Ocramius: Would suggest using library exceptions here

src/Storage/DataSet.php
```diff
@@ -0,0 +1,64 @@
...
+    private $data;
+
+    /**
+     * @var string
+     */
+    private $payload;
+
+    public function __construct(array $data, string $payload)
+    {
+        $this->data = $data;
```
Ocramius: There's no guarantee that `$data` is json-serializable here, right? What does `$data` have to respect?

src/Storage/DataSet.php
```diff
@@ -0,0 +1,64 @@
...
+namespace Lcobucci\JWT\Storage;
+
+use OutOfBoundsException;
+
+/**
+ * @author Luís Otávio Cobucci Oblonczyk <lcobucci@gmail.com>
+ *
+ * @since 4.0.0
+ */
+final class DataSet
```
Ocramius: Could you please describe the purpose of `DataSet` in the docblock?

src/Storage/Parser.php
```diff
@@ -0,0 +1,133 @@
...
+     * @return array
+     *
+     * @throws InvalidArgumentException When JWT don't have all parts
+     */
+    private function splitJwt(string $jwt): array
+    {
+        $data = explode('.', $jwt);
+
+        if (count($data) != 3) {
+            throw new InvalidArgumentException('The JWT string must have two dots');
```
Ocramius: Would suggest moving to library exceptions, as well as throwing the given `$jwt` as part of the message

src/Storage/Parser.php
```diff
@@ -0,0 +1,133 @@
...
+     * @return array
+     *
+     * @throws InvalidArgumentException When an invalid header is informed
+     */
+    private function parseHeader(string $data): array
+    {
+        $header = (array) $this->decoder->jsonDecode($this->decoder->base64UrlDecode($data));
+
+        if (isset($header['enc'])) {
+            throw new InvalidArgumentException('Encryption is not supported yet');
```
Ocramius: Is there an issue for this? Can you add a link to the exception message?

src/Storage/Parser.php
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * Parses the claim set from a string
+     *
+     * @param string $data
+     *
+     * @return array
+     */
+    private function parseClaims(string $data): array
+    {
+        return (array) $this->decoder->jsonDecode($this->decoder->base64UrlDecode($data));
```
Ocramius: Will the decoder return a recursive array? This `(array)` cast in front seems to be only ensuring that an `array` is returned at first level.

src/Storage/Signature.php
```diff
@@ -0,0 +1,48 @@
...
+        $this->payload = $payload;
+    }
+
+    public function hash(): string
+    {
+        return $this->hash;
+    }
+
+    /**
+     * Returns the signature payload as a string representation of the signature
```
Ocramius: This sentence is confusing. Is it returning the payload, or a signature? Am I misunderstanding what a payload is?

src/Storage/Token.php
```diff
@@ -0,0 +1,115 @@
+<?php
+/**
+ * This file is part of Lcobucci\JWT, a simple library to handle JWT and JWS
+ *
+ * @license http://opensource.org/licenses/BSD-3-Clause BSD-3-Clause
+ */
+
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Storage;
```
Ocramius: Why was `Token` moved down to `Storage`? There is no real storage stuff going on anywhere in this PR, as far as I can see...

Ocramius: Maybe the namespace itself should be `Lcobucci\JWT\Token`?

src/Storage/Token.php
```diff
@@ -0,0 +1,115 @@
...
+    private $signature;
+
+    /**
+     * Initializes the object
+     *
+     * @param DataSet $headers
+     * @param DataSet $claims
+     * @param Signature|null $signature
+     */
+    public function __construct(
```
Ocramius: Is this needed to be `public`? Would it be a BC break to change that? Can a named constructor be used instead?

src/Validation/Constraint/AllowedTo.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->audience = $audience;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function assert(Token $token)
+    {
+        if (!in_array($this->audience, $token->claims()->get('aud', []), true)) {
+            throw new ConstraintViolationException(
```
Ocramius: I suggest a named constructor here. Also, adding the actual values to the exception would be 👍

src/Validation/Constraint/IdentifiedBy.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function assert(Token $token)
+    {
+        $claimSet = $token->claims();
+
+        if (!$claimSet->has('jti') || $claimSet->get('jti') !== $this->id) {
+            throw new ConstraintViolationException(
```
Ocramius: Same as above

src/Validation/Constraint/IssuedBy.php
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function assert(Token $token)
+    {
+        $claimSet = $token->claims();
+
+        if (!$claimSet->has('iss')
+            || !in_array($token->claims()->get('iss'), $this->issuers, true)) {
```
Ocramius: What if the issuers count doesn't match?

src/Validation/Constraint/IssuedBy.php
```diff
@@ -0,0 +1,47 @@
...
+            throw new ConstraintViolationException(
```
Ocramius: Same remarks as above

src/Validation/Constraint/RelatedTo.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function assert(Token $token)
+    {
+        $claimSet = $token->claims();
+
+        if (!$claimSet->has('sub') || $claimSet->get('sub') !== $this->subject) {
+            throw new ConstraintViolationException(
```
Ocramius: Same as above

src/Storage/DataSet.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public function get(string $name, $default = null)
+    {
+        if ($this->has($name)) {
+            return $this->data[$name];
+        }
+
+        if ($default === null) {
+            throw new OutOfBoundsException();
```
Ocramius: Unsure why you prefer throwing an exception here, rather than returning the default value. I'd instead suggest adding a parameter to define whether an exception is preferrable. Separate method => even better.

src/Validation/Constraint/IdentifiedBy.php
Ocramius: You always repeat a `has` + `get` check here - I suggest having something like a `get` method that always returns `null` on no value instead (simpler in this scenario)

src/Validation/Constraint/IssuedBy.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function assert(Token $token)
+    {
+        $claimSet = $token->claims();
+
+        if (!$claimSet->has('iss')
```
Ocramius: Same as above on `has` + `get` check

src/Validation/Constraint/IssuedBy.php
Ocramius: There's no insurance that `'iss'` is an array here

src/Validation/Constraint/RelatedTo.php
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function assert(Token $token)
+    {
+        $claimSet = $token->claims();
+
+        if (!$claimSet->has('sub') || $claimSet->get('sub') !== $this->subject) {
```
Ocramius: Same as above on `has` + `get` check

src/Validation/Constraint/SignedWith.php
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function assert(Token $token)
+    {
+        $signature = $token->signature();
+
+        if ($signature === null) {
+            throw new ConstraintViolationException('The token is not signed');
```
Ocramius: Named constructor + throw with the actual values/context

src/Validation/Constraint/SignedWith.php
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        if ($token->headers()->get('alg', 'invalid') !== $this->signer->getAlgorithmId()) {
```
Ocramius: What guarantees that `invalid` cannot be used by a custom algo? Why not use `null`?

src/Validation/Constraint/SignedWith.php
```diff
@@ -0,0 +1,58 @@
...
+            throw new ConstraintViolationException('Token signer mismatch');
```
Ocramius: Named constructor + values

src/Validation/Constraint/SignedWith.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function assert(Token $token)
+    {
+        $signature = $token->signature();
+
+        if ($signature === null) {
```
Ocramius: I see that you do a lot of `=== null` checks in the codebase, even though you know that the type is `object|null`. I suggest dropping it, where the type is binary as in this scenario.

Ocramius: Simply `if (! $signature) {`, I mean.

src/Validation/Constraint/ValidAt.php
```diff
@@ -0,0 +1,62 @@
...
+    public function assert(Token $token)
+    {
+        $claimSet = $token->claims();
+
+        $this->assertExpiration($token);
+        $this->assertBeforeThanNow($claimSet, 'nbf', 'The token cannot be used yet');
+        $this->assertBeforeThanNow($claimSet, 'iat', 'The token was issued in the future');
+    }
+
+    private function assertExpiration(Token $token)
```
Ocramius: Document thrown exceptions in docblock

src/Validation/Constraint/ValidAt.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        if ($token->isExpired($this->now)) {
+            throw new ConstraintViolationException('The token is expired');
+        }
+    }
+
+    private function assertBeforeThanNow(
```
Ocramius: Document thrown exceptions in docblock

src/Validation/Constraint/ValidAt.php
```diff
@@ -0,0 +1,62 @@
...
+        DataSet $claims,
+        string $claimName,
+        string $message
+    ) {
+        if ($claims->has($claimName) && $claims->get($claimName) > $this->now->getTimestamp()) {
```
Ocramius: Same as above on `has` + `get` check

src/Validation/Constraint/ValidAt.php
```diff
@@ -0,0 +1,62 @@
...
+
+        $this->assertExpiration($token);
+        $this->assertBeforeThanNow($claimSet, 'nbf', 'The token cannot be used yet');
+        $this->assertBeforeThanNow($claimSet, 'iat', 'The token was issued in the future');
+    }
+
+    private function assertExpiration(Token $token)
+    {
+        if ($token->isExpired($this->now)) {
+            throw new ConstraintViolationException('The token is expired');
```
Ocramius: Same as above on named constructor + values

src/Validation/Constraint/ValidAt.php
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    private function assertBeforeThanNow(
+        DataSet $claims,
+        string $claimName,
+        string $message
+    ) {
+        if ($claims->has($claimName) && $claims->get($claimName) > $this->now->getTimestamp()) {
+            throw new ConstraintViolationException($message);
```
Ocramius: Same as above on named constructor + values

src/Validation/ConstraintViolationException.php
```diff
@@ -0,0 +1,23 @@
...
+use Lcobucci\JWT\Exception;
+
+/**
+ * @author Luís Otávio Cobucci Oblonczyk <lcobucci@gmail.com>
+ * @author Danny Dörfel <danny.dorfel@gmail.com>
+ * @author Marco Piveta <ocramius@gmail.com>
+ *
+ * @since 4.0.0
+ */
+final class ConstraintViolationException extends Exception
```
Ocramius: Is it really just `extends Exception`? Anything more specific that can be used?

src/Validation/ConstraintViolationException.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Validation;
+
+use Lcobucci\JWT\Exception;
+
+/**
+ * @author Luís Otávio Cobucci Oblonczyk <lcobucci@gmail.com>
+ * @author Danny Dörfel <danny.dorfel@gmail.com>
+ * @author Marco Piveta <ocramius@gmail.com>
```
Ocramius: `Pivetta`

Ocramius: Also: I didn't write this :-P

src/Validation/InvalidTokenException.php
```diff
@@ -0,0 +1,36 @@
...
+namespace Lcobucci\JWT\Validation;
+
+use Lcobucci\JWT\Exception;
+
+/**
+ * @author Luís Otávio Cobucci Oblonczyk <lcobucci@gmail.com>
+ *
+ * @since 4.0.0
+ */
+final class InvalidTokenException extends Exception
```
Ocramius: `extends InvalidArgumentException`? `extends UnexpectedValueException`?

src/Validation/InvalidTokenException.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * @var ConstraintViolationException[]
+     */
+    private $violations;
+
+    public static function fromViolations(ConstraintViolationException ...$violations): self
+    {
+        $exception = new self('The token violates some mandatory constraints');
+        $exception->violations = $violations;
```
Ocramius: Move the violation messages also into the message?

src/Validation/Validator.php
```diff
@@ -0,0 +1,50 @@
...
+namespace Lcobucci\JWT\Validation;
+
+use Lcobucci\JWT\Token;
+
+/**
+ * @author Luís Otávio Cobucci Oblonczyk <lcobucci@gmail.com>
+ *
+ * @since 4.0.0
+ */
+final class Validator implements \Lcobucci\JWT\Validator
```
Ocramius: More than a validator, this looks like a composite assertion. A validator simply returns true/false, this thing is actually throwing a `ConstraintViolationException` containing multiple previous `ConstraintViolationException` instances.

src/Validation/Validator.php
```diff
@@ -0,0 +1,50 @@
...
+     */
+    public function validate(Token $token, Constraint ...$constraints)
+    {
+        $violations = [];
+
+        foreach ($constraints as $constraint) {
+            $violations = $this->assert($violations, $constraint, $token);
+        }
+
+        if (!empty($violations)) {
```
Ocramius: `if ($violations)` ;-)

src/Validation/Validator.php
```diff
@@ -0,0 +1,50 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function validate(Token $token, Constraint ...$constraints)
+    {
+        $violations = [];
+
+        foreach ($constraints as $constraint) {
+            $violations = $this->assert($violations, $constraint, $token);
```
Ocramius: This is overwriting the previous `$violations`. Some confusion around this code.

src/Parser.php
```diff
@@ -10,146 +10,19 @@
...
-
-    /**
-     * Parses the header from a string
-     *
-     * @param string $data
-     *
-     * @return array
-     *
-     * @throws InvalidArgumentException When an invalid header is informed
+     * @throws InvalidArgumentException
```
lcobucci: Planned for #128

src/Storage/DataSet.php
lcobucci: Planned for #128

src/Storage/DataSet.php
lcobucci: Nothing now, should be addressed on #44

src/Storage/DataSet.php
lcobucci: I was just keeping the previous behavior but removing it simplifies a LOT indeed.

src/Storage/Parser.php
lcobucci: Planned for #128, adding `$jwt` to it is a good idea

src/Storage/Parser.php
lcobucci: #6 would solve that. It's a good ideia to add it there

src/Storage/Parser.php
lcobucci: It will return a recursive array however we would have a `TypeError` when an invalid claim set was given.

src/Storage/Signature.php
lcobucci: It's confusing indeed... it's not a payload, it's the encoded value. Will change it.

src/Storage/Token.php
lcobucci: I decided to extract an interface from the Token but definitely needs to be segregated (the only similarity between a nested token and a "normal" token is that they have headers).

`Storage` is a terrible name but I was out of option.
I thought about `Lcobucci\JWT\Token` but `Lcobucci\JWT\Token\Token` is weird.

Another option would be `Lcobucci\JWT\Token\Plain` (but it's also weird IMO) and `Lcobucci\JWT\Token\Encrypted`.

Any thoughts?

src/Storage/Token.php
lcobucci: Not caring about BC breaks here.
The object is only created by the `Builder`, people shouldn't instantiate it.
We can add named constructors here... `Token::unsigned()` and `Token::signed()`?

src/Validation/Constraint/IssuedBy.php
lcobucci: > There's no insurance that `'iss'` is an array here

Issuer is always a `stringOrUri`, so `'iss' is always a string.

The ideia of this constraint is to assert that the token was issued by issuer "a", "b" or "c".

src/Validation/Constraint/SignedWith.php
lcobucci: Stop raising the exception will solve that

src/Validation/ConstraintViolationException.php
lcobucci: `Exception` is `Lcobucci\JWT\Exception`... an initial work of #128.
Didn't want to come to a big inheritance tree now.

src/Validation/Validator.php
lcobucci: It's throwing an `InvalidTokenException` with multiple `ConstraintViolationException` instances.

What would be the best name in your opinion?

src/Validation/Validator.php
lcobucci: Didn't want to use `&$violations` there...

src/Storage/Builder.php
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function sign(Signer $signer, Key $key): BuilderInterface
+    {
+        $this->headers['alg'] = $signer->getAlgorithmId();
+
+        $this->signature = $signer->sign(
```
lcobucci: That's actually a good idea, we could split the `Builder#getToken()` method in two and drop `Builder#sign()` and `Builder#unsign()`

src/Storage/Builder.php
```diff
@@ -0,0 +1,231 @@
...
+
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function with(string $name, $value): BuilderInterface
+    {
+        if ($this->signature !== null) {
```
lcobucci: Signing the token on a late stage will remove the need of having that exception, which is good.

src/Validation/ConstraintViolationException.php
lcobucci: Sorry by the typo.

You helped on #75 discussion... IMO you're a co-author 😄 (but I can remove it if you want)

src/Storage/Token.php
lcobucci: > Could you explain to me what the criteria for grouping things in the Storage namespace is? We need to come up with a better name, IMO.

My idea was to group things related to token structure. Maybe `Builder` and `Parser` doesn't belong there.

> Is there a big difference between encrypted and plain tokens?

In terms of structure, plain tokens have `headers`, `claims` and `signature` (optionally). Encrypted tokens have `headers`, `encrypted key`, `initialization vector`, `ciphertext` and `authentication tag`.

I would say that they should be represented with different objects.

src/Storage/Token.php
lcobucci: As mentioned on https://github.com/lcobucci/jwt/pull/129#discussion_r84674968 I wouldn't add `::encrypted()` here.

src/Storage/Builder.php
```diff
@@ -0,0 +1,231 @@
...
+     */
+    private $encoder;
+
+    /**
+     * Initializes a new builder
+     */
+    public function __construct(Parsing\Encoder $encoder)
+    {
+        $this->encoder = $encoder;
+        $this->headers = ['typ'=> 'JWT', 'alg' => 'none'];
```
lcobucci: What you mean? Move to the property definition?

src/Storage/Builder.php
```diff
@@ -0,0 +1,231 @@
...
+     * {@inheritdoc}
+     */
+    public function canOnlyBeUsedBy(string $audience, bool $addHeader = false): BuilderInterface
+    {
+        $audiences = $this->claims['aud'] ?? [];
+        $audiences[] = $audience;
+
+        return $this->setRegisteredClaim(
+            'aud',
+            array_values(array_map('strval', $audiences)),
```
lcobucci: It could be added indeed (although 100% unrelated with the changeset AHAHAHH)

src/Storage/Builder.php
```diff
@@ -0,0 +1,231 @@
...
+        $this->headers = ['typ'=> 'JWT', 'alg' => 'none'];
+        $this->claims = [];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function canOnlyBeUsedBy(string $audience, bool $addHeader = false): BuilderInterface
+    {
+        $audiences = $this->claims['aud'] ?? [];
```
lcobucci: However I think we should add it on a different PR

src/Storage/Builder.php
Ocramius: Correct.

src/Storage/Builder.php
Ocramius: New issue then?

src/Storage/DataSet.php
Ocramius: If you can describe its purpose, we can look for a new name...

src/Storage/Token.php
Ocramius: Different objects, yes, but different namespaces? Not sure...

src/Validation/Validator.php
Ocramius: `AggregateAssertion` or `MultiAssertion` works too... It would still throw a subtype of `ConstraintViolationException`

src/Storage/DataSet.php
dannydorfel: The class is used for header and payload, right? Perhaps changing it to a Header and Payload class?

dannydorfel:
Nice! And now I understand the "case of the Mondays" tweet. 
I have got nothing to add 😁

src/Storage/Builder.php
lcobucci: #130

src/Storage/Builder.php
lcobucci: #131

src/Validation/Validator.php
lcobucci: `InvalidTokenException` is not a `ConstraintViolationException` and my idea is that users should only know about `InvalidTokenException`.

Of course we could turn these into a composite specification (which is closer to what you're describing IMO) but I don't think we need that complexity...

> A validator simply returns true/false

I'm a bit skeptical about it... not saying that's not true but it sounds over simplified.
Maybe we can offer both things:

```php
interface Validator
{
    /**
     * @throws InvalidTokenException
     */
    public function assert(Token $token, Constraint ...$constraints): void;
    public function validate(Token $token, Constraint ...$constraints): bool;
}
```

One would use `assert()` when is looking for a "detailed report" or `validate()` when is just caring about the end result.

src/Storage/DataSet.php
lcobucci: > The class is used for header and payload, right? Perhaps changing it to a Header and Payload class?

There's no difference in terms of behavior there... I wouldn't split them now

src/Storage/Token.php
lcobucci: I think it depends on how we want to tackle the encryption stuff. Based on what I know now I'd say that they should be in the same namespace but something might change after reading the specs again.

src/Storage/Builder.php
lcobucci: Actually I'll solve it here, that `array_map() + array_values()` is useless now...

lcobucci: (and it was hurting my eyes HAHAHA)

src/Validation/Constraint/ValidAt.php
```diff
@@ -0,0 +1,62 @@
...
+     * {@inheritdoc}
+     */
+    public function assert(Token $token)
+    {
+        $claimSet = $token->claims();
+
+        $this->assertExpiration($token);
+        $this->assertBeforeThanNow($claimSet, 'nbf', 'The token cannot be used yet');
+        $this->assertBeforeThanNow($claimSet, 'iat', 'The token was issued in the future');
+    }
```
schnittstabil: The first `ConstraintValidationException` wins – In my opinion, we should instead implement `validate`:
```php
    /**
     * @return ConstraintValidation[]
     */
    public function validate(Token $token)
    {
        $claimSet = $token->claims();
        return array_merge(
            $this->validateExpiration($token),
            $this->validateBeforeThanNow($claimSet, 'nbf', 'The token cannot be used yet'),
            $this->validateBeforeThanNow($claimSet, 'iat', 'The token was issued in the future')
        );
    }

    /**
     * @return ConstraintValidation[]
     */
    private function validateExpiration(Token $token)
    …
```

schnittstabil: This also means, we can drop the ugly `try`-`catch` at `Validator::assert`.

src/Validator.php
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @author Luís Otávio Cobucci Oblonczyk <lcobucci@gmail.com>
+ * @author Danny Dörfel <danny.dorfel@gmail.com>
+ * @author Marco Pivetta <ocramius@gmail.com>
+ * @author Henrique Moody <henriquemoody@gmail.com>
+ *
+ * @since 4.0.0
+ */
+interface Validator
```
schnittstabil: Sorry, I don't know if this was already discussed, but how do we want to use Validators?
Currently(?):
```php
$config = new Configuration();
$validator = $config->getValidator();

$validator->validate($token, new ValidAt(new DateTimeImmutable()));
$validator->validate($token, new IssuedBy('test.com'));
```

As `Configuration` *serves like a small DI container*, there are also other possibilities, e.g.:
```php
// during bootstrapping
$jwtConfig = new Configuration();
$jwtConfig->getValidator()
    ->addConstraint(new ValidAt(new DateTimeImmutable()))
    ->addConstraint(new IssuedBy('test.com'));

// somewhere else
$jwtConfig->getValidator()->validate($token);
```

lcobucci: > Sorry, I don't know if this was already discussed, but how do we want to use Validators?

```php
$config = new Configuration();
$validator = $config->getValidator();

$validator->validate(
     $token,
     new ValidAt(new DateTimeImmutable()),
     new IssuedBy('test.com')
);

// or simply
$constraints = [new ValidAt(new DateTimeImmutable()), new IssuedBy('test.com')];
$validator->validate($token, ...$constraints);
```

> As Configuration serves like a small DI container, there are also other possibilities, e.g.:

I don't see changing the state of the service as an option, IMO that's not a good practice. Also, `new ValidAt(new DateTimeImmutable())` might lead to some issues due to the creation and the validation time.

schnittstabil: That was some food for thought – changing the state is one of many possibilities, but you're right the splat operator offers great options for code reuse. The more I think about it, the more I love it.

src/Builder.php
```diff
@@ -11,277 +11,70 @@
...
-        $signer->modifyHeader($this->headers);
-
-        $this->signature = $signer->sign(
-            $this->getToken()->getPayload(),
-            $key
-        );
-
-        return $this;
-    }
+    public function with(string $name, $value): Builder;
```
henriquemoody: Only `with()` looks a little bit vague to me, wouldn't be better to call it `withClaim` instead?

src/Storage/Builder.php
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * Configures a registered claim
+     */
+    private function setRegisteredClaim(string $name, $value, bool $addHeader): BuilderInterface
+    {
+        $this->with($name, $value);
+
+        if ($addHeader) {
+            $this->headers[$name] = $this->claims[$name];
```
henriquemoody: Although they're private properties and _you know_ what `with()` does, I would create some method to get the claims by name, that you could also use on [`canOnlyBeUsedBy()`](https://github.com/lcobucci/jwt/pull/129/files#diff-343ce2f1a7cebf5e8711bc1f726796abR60), and I would also use `withHeader()` instead of setting `$this->headers` manually.

src/Storage/Builder.php
```diff
@@ -0,0 +1,189 @@
...
+     * {@inheritdoc}
+     */
+    public function with(string $name, $value): BuilderInterface
+    {
+        $this->claims[$name] = $value;
+
+        return $this;
+    }
+
+    private function encode(array $items): string
```
henriquemoody: Isn't this useful on the Encoder itself?

lcobucci: Not sure... sometimes you don't want to encode with JSON+Base64Url (binary data is just base64url for example)

src/Validation/Validator.php
lcobucci: @henriquemoody what I dislike about having a result object is the requiring users to do this when they just care about the result:

```php
if (!$validator->validate($token, ...$constraints)->isValid()) { // or assign a variable for the result ofc
}
```

> Split it into two methods may be a good idea but also looks too much responsibility for the `Validator`.

Both methods are related to the same thing, so the reason to change is the same.

@dannydorfel @henriquemoody @schnittstabil @Ocramius thanks for your suggestions regarding this. I think the best option for now is splitting in two methods.

schnittstabil:
```php
$validator = new Validator();
$validator->validate($token, ...$constraints)->isValid()
```
**Eek, 3 times *valid* on a single line. Wait – 3 times?**

The `Validator` class is final and doesn't have any state, thus it is only a *container* or the *namespace* for the `validate` and `assert` functions, isn't it? For consumers, this is very inconvenient, I think they should use those functions directly:
```php
use function Lcobucci\JWT\assert;
use function Lcobucci\JWT\validate;

assert($token, ...$constraints);
$result = validate($token, ...$constraints); // whatever result will be
```

Btw, because of this, I don't know if the `Validator` class and the `Validator` interface are need at all. Interop? – Questionable. Of course, they can be kept for internal usage, but I think providing only functions is much more appropriate.

schnittstabil: Another suggestion:
```php
use function Lcobucci\JWT\assert;   // clearly throws Exceptions
use function Lcobucci\JWT\isValid;  // clearly returns boolean values
use function Lcobucci\JWT\validate; // returns ValidationResult objects, IMO acceptable
```

src/Storage/Token.php
lcobucci: After thinking a bit about this I've decide to:

- remove `claims()`, `signature()` and `payload()` from the interface so we can reuse it for encrypted tokens in the future
- rename `Lcobucci\JWT\Storage` to `Lcobucci\JWT\Token`
- rename `Lcobucci\JWT\Token\Token` to `Lcobucci\JWT\Token\Plain`
- make builder return an instance of `Lcobucci\JWT\Token\Plain` instead of the interface
- implement #117 here

src/Builder.php
```diff
@@ -11,277 +11,70 @@
...
-            ? array_merge($this->claims['aud']->getValue(), [$audience])
-            : [$audience];
-
-        return $this->setRegisteredClaim(
-            'aud',
-            array_values(array_map('strval', $audiences)),
-            $addHeader
-        );
-    }
+    public function canOnlyBeUsedBy(string $audience, bool $addHeader = false): Builder;
```
lcobucci: Not really, `audience` is about who can consume the token.
I wouldn't change it.

src/Validation/InvalidTokenException.php
```diff
@@ -0,0 +1,36 @@
...
+
+        return $exception;
+    }
+
+    public function violations(): array
+    {
+        return $this->violations;
```
lcobucci: Solved, thanks 😉

src/Configuration.php
```diff
@@ -123,4 +110,18 @@ public function setDecoder(Parsing\Decoder $decoder)
...
+    public function getValidator(): Validator
+    {
+        if ($this->validator === null) {
+            $this->validator = new Validation\Validator();
+        }
+
+        return $this->validator;
+    }
+
+    public function setValidator(Validator $validator)
```
Ocramius: Should have a `@return void` docblock

src/Token.php
```diff
@@ -9,267 +9,59 @@
...
     /**
-     * Returns the value of a token claim
-     *
-     * @param string $name
-     * @param mixed $default
-     *
-     * @return mixed
-     *
-     * @throws OutOfBoundsException
+     * Returns if the token minimum time is before than given time
```
Ocramius: Minimum time of what?

src/Token/DataSet.php
```diff
@@ -0,0 +1,54 @@
...
+    private $data;
+
+    /**
+     * @var string
+     */
+    private $encoded;
+
+    public function __construct(array $data, string $encoded)
+    {
+        $this->data = $data;
```
Ocramius: This is super-weird.

 1. Parallel data structure: argh!
 2. No control over its validity

src/Token/Parser.php
```diff
@@ -0,0 +1,145 @@
...
+    {
+        $this->decoder = $decoder;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function parse(string $jwt): TokenInterface
+    {
+        $data = $this->splitJwt($jwt);
```
Ocramius: I suggest using `list() =` here

src/Token/Parser.php
```diff
@@ -0,0 +1,145 @@
...
+     *
+     * @return array
+     *
+     * @throws InvalidArgumentException When JWT don't have all parts
+     */
+    private function splitJwt(string $jwt): array
+    {
+        $data = explode('.', $jwt);
+
+        if (count($data) != 3) {
```
Ocramius: Isn't the signature optional?

src/Token/Parser.php
```diff
@@ -0,0 +1,145 @@
...
+     * Returns the signature from given data
+     *
+     * @param array $header
+     * @param string $data
+     *
+     * @return Signature|null
+     */
+    private function parseSignature(array $header, string $data)
+    {
+        if ($data === '' || !isset($header['alg']) || $header['alg'] === 'none') {
```
Ocramius: This plays well with the idea of having a `NoneSigner` in the builder

src/Token/Plain.php
```diff
@@ -0,0 +1,166 @@
...
+
+        return $now->getTimestamp() > $this->claims->get('exp');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __toString(): string
+    {
+        return implode('.', get_object_vars($this));
```
Ocramius: Consider getting rid of @`get_object_vars()` here, as no order is guaranteed

src/Validation/Constraint.php
```diff
@@ -0,0 +1,27 @@
...
+ * @since 4.0.0
+ */
+interface Constraint
+{
+    /**
+     * @param Token $token
+     *
+     * @throws ConstraintViolationException
+     */
+    public function assert(Token $token);
```
Ocramius: :+1:

src/Validation/Constraint.php
```diff
@@ -0,0 +1,27 @@
...
+ * @author Luís Otávio Cobucci Oblonczyk <lcobucci@gmail.com>
+ *
+ * @since 4.0.0
+ */
+interface Constraint
+{
+    /**
+     * @param Token $token
+     *
+     * @throws ConstraintViolationException
```
Ocramius: Add `@return void`

src/Validation/InvalidTokenException.php
```diff
@@ -0,0 +1,51 @@
...
+    private static function buildMessage(array $violations): string
+    {
+        $violations = array_map(
+            function (ConstraintViolationException $violation): string {
+                return '- ' . $violation->getMessage();
+            },
+            $violations
+        );
+
+        $message = 'The token violates some mandatory constraints, details:' . PHP_EOL;
```
Ocramius: I'd just use `\n`

src/Validation/InvalidTokenException.php
```diff
@@ -0,0 +1,51 @@
...
+        $message .= implode(PHP_EOL, $violations);
```
Ocramius: Samee: `\n`

src/Token/Parser.php
lcobucci: Yeap, but the last `.` is not so it would be an empty string.

src/Builder.php
lcobucci: Those names were defined on #113 we're just extracting an interface. Let's define on #150 and change them on a different PR.

src/Builder.php
lcobucci: Those names were defined on #113 we're just extracting an interface. Let's define on #150 and change them on a different PR.

src/Builder.php
```diff
@@ -10,278 +10,71 @@
...
-
-        return $this;
-    }
+    public function getToken(Signer $signer, Key $key): Plain;
 
     /**
-     * Returns the resultant token
-     *
-     * @return Token
+     * Returns an unsecured token (not recommended)
```
lcobucci: What you mean here?

src/Configuration.php
lcobucci: Wouldn't bother with that now since I want to merge it and work on #146

src/Builder.php
Ocramius: Have always a token, let a `NoneSigner`deal with it (valifates only empty signature, generates only empty signature)

src/Token.php
lcobucci: It validates the "not before" claim, so in theory it's the minimum time of usage but I didn't have a better name... any suggestion?

src/Builder.php
lcobucci: That's a good idea, it would simplify things indeed.

src/Validation/Constraint.php
lcobucci: Wouldn't bother with that now since I want to merge it and work on #146

src/Token.php
Ocramius: Ah, that's confusing because of the `not` in `not before` :-\

I guess that's what the spec is - can't fix the naming then.

src/Token/DataSet.php
lcobucci: The intent here is to encapsulate both values in a common structure for headers and claims (and reduce the repeated methods in the `Token`)...

It simplified things a lot IMO but it's indeed a parallel data structure and there's no way to ensure that `$encoded` is the string representation of the `$data` and of course the validity but I'd say it's an acceptable trade off.

Do you have any suggestion here?

src/Builder.php
lcobucci: But it can be done in a different PR... I really want to have this merged instead of adding things to the stack

lcobucci: #152 is there for that.

src/Token/DataSet.php
lcobucci: Going to merge it, could you please open an issue with your suggestion?

PR #153, opened 2017-01-08
Require PHP 7.1

lcobucci:


src/Signer/Rsa.php
```diff
@@ -52,7 +52,7 @@ public function verify(string $expected, string $payload, Key $key): bool
      *
      * @expectedException InvalidArgumentException
```
Ocramius: This is probably a mistake/leftover

.travis.yml
```diff
@@ -2,12 +2,11 @@ language: php
 sudo: false
 
 php:
-  - 7.0
   - 7.1
   - nightly
 
 before_script:
-  - composer install --prefer-dist -o
+  - composer install -o
```
Ocramius: Don't use `-o` for tests, as it will cause the test suite to succeed even if there is a misnamed class

composer.json
```diff
@@ -47,6 +47,9 @@
             ]
         }
     },
+    "config": {
+        "preferred-install": "dist"
```
Ocramius: Oh, that's new :O

lcobucci: AHAHAHA I don't think it's THAAAAT new 😜

src/Signer/Rsa.php
lcobucci: Definitely

manchuck:
Just out of curiosity, why php 7.1 and not 7.0 ?   I do not see any php 7.1 features being used.  I only ask because I am currently running 7.0 using 4.0.0.  This is now causing some issues with our build process

manchuck:
NVM found the nullable type for getType on the builder

Issue #130, opened 2016-10-26
Extract the registered claim names to constants


lcobucci:
In order to remove the strings from all over the library and avoid issues

PR #159, opened 2017-02-01
Move string references of registered claims to constants

lcobucci:
Fixes: https://github.com/lcobucci/jwt/issues/130


Ocramius:
👍

PR #160, opened 2017-02-04
Create none signer

### LINKED ISSUE #152: Create the `Signer\None` to simplify things
As per @Ocramius we could have a `None` signer that would handle empty signatures.

lcobucci:
Fixes #152

### DIFF
--- a/test/unit/Validation/Constraint/ValidAtTest.php
+++ b/test/unit/Validation/Constraint/ValidAtTest.php
@@ -38,6 +38,7 @@ final class ValidAtTest extends ConstraintTestCase
      *
      * @uses \Lcobucci\JWT\Token\DataSet
      * @uses \Lcobucci\JWT\Token\Plain
+     * @uses \Lcobucci\JWT\Token\Signature
      */
     public function assertShouldRaiseExceptionWhenTokenIsExpired(): void
     {
@@ -66,6 +67,7 @@ final class ValidAtTest extends ConstraintTestCase
      *
      * @uses \Lcobucci\JWT\Token\DataSet
      * @uses \Lcobucci\JWT\Token\Plain
+     * @uses \Lcobucci\JWT\Token\Signature
      */
     public function assertShouldRaiseExceptionWhenMinimumTimeIsNotMet(): void
     {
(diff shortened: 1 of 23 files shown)


lcobucci:
@Ocramius I like the final result but still have mixed feelings regarding creating "useless" objects (`Key` and `Signature`).

I can also convert the `new Signature('', '')` to a named constructor. I don't think it makes sense to use inheritance there (to have a special case object)...

test/unit/Validation/Constraint/ValidAtTest.php
```diff
@@ -38,6 +38,7 @@ public function createDependencies(): void
      *
      * @uses \Lcobucci\JWT\Token\DataSet
      * @uses \Lcobucci\JWT\Token\Plain
+     * @uses \Lcobucci\JWT\Token\Signature
```
Ocramius: Just noticed all these `@uses`: since I didn't notice before, what is the conveyed intent?

lcobucci: Although the objects are used I don't want to make PHPUnit mark them as covered on the coverage report.

`@covers` and `@uses` are a requirement since I configured PHPUnit to be strict about these annotations.

lcobucci:
> That's indeed something I'd suggest as well. 

@Ocramius any suggestion regarding the method name? Not so creative right now 😜 

---

btw I was also thinking about adding `$signingKey` and `$verificationKey` (maybe with a different name) on the `Configuration` and change the method that configure the signer to `configureSigner(Signer $signer, Key $signingKey, ?Key $verificationKey): void` so people can configure these things in a single operation (since they're related), do you have any opinion regarding this?

Ocramius:
> @Ocramius any suggestion regarding the method name?

Start from a sentence explaining what you are doing when creating an empty signature.

 > so people can configure these things in a single operation (since they're related), do you have any opinion regarding this?

Seems like a good idea, since you'd never want just one or two of the three.

lcobucci:
@Ocramius I've added the named constructor, can you review the PR again please?

> Seems like a good idea, since you'd never want just one or two of the three.

I'll create an issue to do that 😉

PR #170, opened 2017-02-17
Enhancement: Keep packages sorted without specifying --sort-packages

localheinz:
This PR

* [x] keeps packages in `composer.json` sorted (without the need to specify the `--sort-packages` option)

💁‍♂️ For reference, see https://getcomposer.org/doc/06-config.md#sort-packages.


Ocramius:
LGTM 👍

localheinz:
Thank you, @Ocramius!

PR #171, opened 2017-02-17
Handle claims conversion

### LINKED ISSUE #43: Use DateTime for date claims

### LINKED ISSUE #144: Support non-integer timestamps
RFC 7519 clearly states (page 5) that for NumericDate properties, "non-integer values can be represented", but your version 3.2.1 code appears to cast all timestamps to (int)!  For instance, Builder's setExpiration() and setNotBefore(); and ValidationData's setCurrentTime().  A whole second is an eternity in CPU time.  Can we please have decimal seconds?

lcobucci:
There're some weird stuff to improve @Ocramius (and I couldn't resist adding the last commit 😂).

Fixes #144 
Fixes #43

### DIFF
--- a/src/Token/Builder.php
+++ b/src/Token/Builder.php
@@ -173,6 +173,10 @@ final class Builder implements BuilderInterface
 
     private function formatClaims(array $claims): array
     {
+        if (isset($claims[RegisteredClaims::AUDIENCE][0]) && !isset($claims[RegisteredClaims::AUDIENCE][1])) {
+            $claims[RegisteredClaims::AUDIENCE] = $claims[RegisteredClaims::AUDIENCE][0];
+        }
+
         foreach (array_intersect(RegisteredClaims::DATE_CLAIMS, array_keys($claims)) as $claim) {
             $claims[$claim] = $this->convertDate($claims[$claim]);
         }
--- a/src/Token/Builder.php
+++ b/src/Token/Builder.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 
 namespace Lcobucci\JWT\Token;
 
+use DateTimeImmutable;
 use Lcobucci\Jose\Parsing;
 use Lcobucci\JWT\Builder as BuilderInterface;
 use Lcobucci\JWT\Signer;
@@ -68,7 +69,7 @@ final class Builder implements BuilderInterface
     /**
      * {@inheritdoc}
      */
-    public function expiresAt(int $expiration): BuilderInterface
+    public function expiresAt(DateTimeImmutable $expiration): BuilderInterface
     {
         return $this->setClaim(RegisteredClaims::EXPIRATION_TIME, $expiration);
     }
@@ -84,7 +85,7 @@ final class Builder implements BuilderInterface
     /**
      * {@inheritdoc}
      */
-    public function issuedAt(int $issuedAt): BuilderInterface
+    public function issuedAt(DateTimeImmutable $issuedAt): BuilderInterface
     {
         return $this->setClaim(RegisteredClaims::ISSUED_AT, $issuedAt);
     }
@@ -100,7 +101,7 @@ final class Builder implements BuilderInterface
     /**
      * {@inheritdoc}
      */
-    public function canOnlyBeUsedAfter(int $notBefore): BuilderInterface
+    public function canOnlyBeUsedAfter(DateTimeImmutable $notBefore): BuilderInterface
     {
         return $this->setClaim(RegisteredClaims::NOT_BEFORE, $notBefore);
     }
@@ -158,7 +159,7 @@ final class Builder implements BuilderInterface
         $headers['alg'] = $signer->getAlgorithmId();
 
         $encodedHeaders = $this->encode($headers);
-        $encodedClaims = $this->encode($this->claims);
+        $encodedClaims = $this->encode($this->formatClaims($this->claims));
 
         $signature = $signer->sign($encodedHeaders . '.' . $encodedClaims, $key);
         $encodedSignature = $this->encoder->base64UrlEncode($signature);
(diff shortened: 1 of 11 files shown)


src/Token/Builder.php
```diff
@@ -169,4 +170,29 @@ public function getToken(Signer $signer, Key $key): Plain
             new Signature($signature, $encodedSignature)
         );
     }
+
+    private function formatClaims(array $claims): array
+    {
+        if (isset($claims[RegisteredClaims::AUDIENCE]) && count($claims[RegisteredClaims::AUDIENCE]) === 1) {
```
Ocramius: Please don't sneak in this optimisation here :-P

Ocramius: Also, if you assume that it always is a packed array, simply use an `isset()` on index `0` and another one on index `1`

src/Token/Builder.php
```diff
@@ -169,4 +170,29 @@ public function getToken(Signer $signer, Key $key): Plain
...
+                continue;
+            }
+
+            $claims[$claim] = $this->convertDate($claims[$claim]);
+        }
+
+        return $claims;
+    }
+
+    private function convertDate(DateTimeImmutable $date)
```
Ocramius: Return type declaration needed

src/Token/Builder.php
```diff
@@ -169,4 +170,29 @@ public function getToken(Signer $signer, Key $key): Plain
...
+    {
+        $seconds = $date->format('U');
+        $microseconds = $date->format('u');
+
+        return $microseconds === '000000' ? (int) $seconds : $seconds . '.' . $microseconds;
```
Ocramius: Big assumption on the number of decimal digits. Instead, use `0 === (int) $microseconds`

src/Token/Builder.php
Ocramius: I see it's `int|string`. Can't it be `float`?

Ocramius: Note that `json_encode()` does some trickery around floats, making them integers when no part is after the comma

Ocramius: Ref: https://3v4l.org/Uq1CL

src/Token/Parser.php
```diff
@@ -90,7 +91,30 @@ private function parseHeader(string $data): array
      */
     private function parseClaims(string $data): array
     {
-        return (array) $this->decoder->jsonDecode($this->decoder->base64UrlDecode($data));
+        $claims = (array) $this->decoder->jsonDecode($this->decoder->base64UrlDecode($data));
+
+        if (isset($claims[RegisteredClaims::AUDIENCE])) {
```
Ocramius: Note that casting `null` to `(array)` yields an empty array. I don't know if that's valid for `$claims`, but worth considering

src/Token/Parser.php
```diff
@@ -90,7 +91,30 @@ private function parseHeader(string $data): array
...
+            $claims[RegisteredClaims::AUDIENCE] = (array) $claims[RegisteredClaims::AUDIENCE];
+        }
+
+        foreach (RegisteredClaims::DATE_CLAIMS as $claim) {
+            if (!isset($claims[$claim])) {
```
Ocramius: `areay_intersect_key()`?

src/Token/Parser.php
```diff
@@ -90,7 +91,30 @@ private function parseHeader(string $data): array
...
+        return $claims;
+    }
+
+    private function convertDate(string $value): DateTimeImmutable
+    {
+        if (strpos($value, '.') === false) {
+            $value .= '.0';
+        }
+
+        return DateTimeImmutable::createFromFormat('U.u', $value);
```
Ocramius: I'd use two different expressions for when a microtime is set or not: this way you avoid overwriting `$value`

src/Token/RegisteredClaims.php
```diff
@@ -27,6 +27,8 @@
         self::SUBJECT
     ];
 
+    const DATE_CLAIMS = [self::ISSUED_AT, self::NOT_BEFORE, self::EXPIRATION_TIME];
```
Ocramius: One per line, please

src/Token/Builder.php
lcobucci: `DateTimeInterface#format('u')` always returns a 6 chars string. It can be compared to 0 though, there's no much difference

src/Token/Parser.php
lcobucci: I don't want to set if it doesn't exist in the original token

src/Token/Builder.php
lcobucci: I was with the old behaviour in my mind: https://3v4l.org/1Q7hh
Then when we convert it back to a `DateTime` we would have `DateTimeImmutable::createFromFormat('U.u', '1487372450.0231')` instead of `DateTimeImmutable::createFromFormat('U.u', '1487372450.023126')`, what can cause some weird errors.

src/Token/Builder.php
lcobucci: Yeah it's definitely better then `count()` in that case

src/Token/Builder.php
Ocramius: Ah, I see what you mean. If you can, you should cover the edge case you just mentioned in the test suite

src/Token/Builder.php
Ocramius: Yeah, but nothing prevents the engine from making this 7 chars, or similar, in future versions. This code is very fragile

src/Token/Parser.php
Ocramius: Makes sense

src/Token/Builder.php
lcobucci: Indeed

src/Token/Parser.php
lcobucci: In order to use it I would also have to do an `array_flip(RegisteredClaims::DATE_CLAIMS)` (or an `array_keys($claims)` and use `array_intersect()` instead) not sure if will help us...

src/Token/Builder.php
lcobucci: It's being covered on https://github.com/lcobucci/jwt/pull/171/files#diff-89e66398726c97dea8259f9c0d31b5f5R417

lcobucci: But can be more explicit, ofc

src/Token/Parser.php
Ocramius: Or define the constants in the keys ;-)

lcobucci:
@Ocramius could you please review it again?

PR #180, opened 2017-05-07
Use a time provider instead on `ValidAt` constraint

lcobucci:
So that we can separate the constraint initialisation from the validation itself and then be able to map validation constraints in a dependency injection container.


lcobucci:
@Ocramius @henriquemoody sorry guys but I'll merge this one... I want to release an alpha version so I can send a PR to fix psr7-sessions/storageless#61 😉

PR #185, opened 2017-06-02
fix two dot token string returned by getPayload() function according to RFC

belaz:
Current function getPayload() in the Token class return 1 dotted string.

See p8
[https://tools.ietf.org/html/rfc7519](https://tools.ietf.org/html/rfc7519)

### DIFF
--- a/src/Token.php
+++ b/src/Token.php
@@ -263,7 +263,7 @@ class Token
      */
     public function getPayload()
     {
-        return $this->payload[0] . '.' . $this->payload[1];
+        return implode('.', $this->payload);
     }
 
     /**


Ocramius:
Requires a test case

Ocramius:
@belaz what's the resolution of this? Is the issue invalid?

belaz:
Hello, the payload must return 3 sections and two dots according to p8 of the RFC, you are currently unit testing that getPayload()  returns 2 sections and one dot.
The issue for me is only to rewrite your unit tests but i think you're unit testing this by design, i will resubmit this when i'm ready to read all the code.

Ocramius:
Re-opening until it's clear if this is a bug or the path is not affected.

lcobucci:
This issue is invalid, this is the payload for signature validation (that should contain only the encoded headers and claims). 

Cast the object to string if you want to have the string representation of the token.

qlixes:
I have some problem while using version 3.3.1
all output already I've checked and tested on JWT.io, it's said valid JWT String. but while try to parse, it raise exception  the jwt string must  have 2 dot.
my code on :
https://pastebin.com/1q6Z4CVZ
Please let me know If I've did wrong way.
Thank you a lot

Issue #198, opened 2017-10-04
Add PHPCS as a dev dependency


lcobucci:
Also adding it as a Travis-CI stage.

This is the config I'm using for my projects and can be used as sample: https://github.com/lcobucci/chimera-foundation/blob/b125eb9f07509ee977298d599655c2e1d391cf0e/phpcs.xml.dist

PR #199, opened 2017-10-04
Add CS checking to build

vanbrabantf:
This is a PR for issue #198.
It includes the https://github.com/lcobucci/chimera-foundation/blob/b125eb9f07509ee977298d599655c2e1d391cf0e/phpcs.xml.dist cs rules.

They seem to break for the current code base (mostly the `=` alignment).
I can fix those as well if you like.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ composer.lock
 humbuglog.txt
 coverage
 phpcs.xml
+/.phpcs.cache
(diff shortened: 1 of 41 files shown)


lcobucci:
@vanbrabantf 🚢

lcobucci:
@vanbrabantf Thanks again for your contribution 🎉 

> I can fix those as well if you like.

I've just executed `phpcbf` to fix everything and added a new commit to your PR. Automate ALL THE THINGS o/

Issue #207, opened 2017-10-20
Errors while reading key from file are not caught on the correct place


lcobucci:
No exception is raised when `file_get_contents()` returns `false` in https://github.com/lcobucci/jwt/blob/3.2.2/src/Signer/Key.php#L67, which can lead to errors in other places.

lcobucci:
Possible cause of #206

lcobucci:
Fixed by #209

PR #209, opened 2017-11-01
Add content check after key file reading

misantron:
Issue #207

### DIFF
--- a/src/Signer/Key.php
+++ b/src/Signer/Key.php
@@ -7,7 +7,9 @@
 
 namespace Lcobucci\JWT\Signer;
 
+use Exception;
 use InvalidArgumentException;
+use SplFileObject;
 
 /**
  * @author Luís Otávio Cobucci Oblonczyk <lcobucci@gmail.com>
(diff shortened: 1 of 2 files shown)


lcobucci:
@misantron sorry about the huge delay to get back to you, but can you please add tests to your changes?

misantron:
I have no idea how to test this case. I think the only way is to mock internal php function but I don't think it's good idea. Any suggestion?

lcobucci:
Maybe locking the file with `flock()`?

misantron:
Ok I'll try

lcobucci:
@misantron thanks!

misantron:
Test is ready but the only way to prevent warnings is to use @ operator. I know it's a bad practice but it's exceptional case.

lcobucci:
@misantron I was thinking about this... maybe using `SplFileObject` should be simpler and remove the `@`.

misantron:
It's not a solution I think. SPL object using same PHP SAPI inside that's why behavour will be the same. And additionaly we have problems with testing because of this: https://github.com/mikey179/vfsStream/wiki/Known-Issues

misantron:
@lcobucci if you decide to use ```SplFileObject``` I can do it in this patch.

PR #210, opened 2017-12-01
Fixed bug where audience claim could not be an array

Zwartpet:
According to https://tools.ietf.org/html/rfc7519#section-4.1.3 audience claim should be an array but MAY be a string

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -74,14 +74,14 @@ class Builder
     /**
      * Configures the audience
      *
-     * @param string $audience
+     * @param string|array $audience
      * @param boolean $replicateAsHeader
      *
      * @return Builder
      */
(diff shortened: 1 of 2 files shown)


src/Builder.php
```diff
@@ -74,14 +74,14 @@ public function __construct(
...
-     * @param string $audience
+     * @param string|array $audience
      * @param boolean $replicateAsHeader
      *
      * @return Builder
      */
     public function setAudience($audience, $replicateAsHeader = false)
     {
-        return $this->setRegisteredClaim('aud', (string) $audience, $replicateAsHeader);
+        return $this->setRegisteredClaim('aud', $audience, $replicateAsHeader);
```
lcobucci: Removing the type cast allows people to send anything to there which is not ideal.

Issue #206, opened 2017-10-19
Random error


yohkun:
Hi, i have unexplained random error : 
`<br />
<b>Fatal error</b>:  Uncaught TypeError: Argument 3 passed to Lcobucci\JWT\Signer\Hmac::doVerify() must be an instance of Lcobucci\JWT\Signer\Key, boolean given, called in C:\Users\admin\programmation\application-tr\application-TR-BO\vendor\lcobucci\jwt\src\Signer\BaseSigner.php on line 42 and defined in C:\Users\admin\programmation\application-tr\application-TR-BO\vendor\lcobucci\jwt\src\Signer\Hmac.php:29
Stack trace:
#0 C:\Users\admin\programmation\application-tr\application-TR-BO\vendor\lcobucci\jwt\src\Signer\BaseSigner.php(42): Lcobucci\JWT\Signer\Hmac-&gt;doVerify('\xA0&lt;\x81D\x07\x084R\xF7R\x03r\xBD}L...', 'eyJ0eXAiOiJKV1Q...', false)
#1 C:\Users\admin\programmation\application-tr\application-TR-BO\vendor\lcobucci\jwt\src\Signature.php(47): Lcobucci\JWT\Signer\BaseSigner-&gt;verify('\xA0&lt;\x81D\x07\x084R\xF7R\x03r\xBD}L...', 'eyJ0eXAiOiJKV1Q...', false)
#2 C:\Users\admin\programmation\application-tr\application-TR-BO\vendor\lcobucci\jwt\src\Token.php(201): Lcobucci\JWT\Signature-&gt;verify(Object(Lcobucci\JWT\Signer\Hmac\Sha256), 'e in <b>C:\Users\admin\programmation\application-tr\application-TR-BO\vendor\lcobucci\jwt\src\Signer\Hmac.php</b> on line <b>29</b><br />
`

Do you ave an idea ?

lcobucci:
@yohkun it looks like you're sending `false` to `Token#verify()`, you should check that

yohkun:
For the key or the secret ?
All my verify call look like : $token->verify(new Sha256(), "secret"))
That will work 40 time and the 41 i will get the error

lcobucci:
> That will work 40 time and the 41 i will get the error

lcobucci:
> For the key or the secret ?

The 2nd argument of `Token#verify()` is your verification key.

> That will work 40 time and the 41 i will get the error

[...]

PR #216, opened 2017-12-26
Check if file_get_contents returns false - #206 #207

Rkallenkoot:
Not sure how to test this, `is_readable` seems to catch most of the cases.


Rkallenkoot:
Any suggestions on how to test this?

Ocramius:
Not really, but it looks like a race condition or a directory fopen attempt

PR #225, opened 2018-03-06
Add missing docblock argument

Quezler:


src/Signer/BaseSigner.php
```diff
@@ -69,6 +69,7 @@ private function getKey($key)
     /**
      * Creates a hash with the given data
      *
+     * @param mixed $expected
```
lcobucci: Nice catch! But this should rather be `string` as you can see on the new API: https://github.com/lcobucci/jwt/blob/9d9479b95103e9dc851f0f7b7d63f012d416a29f/src/Signer.php#L28

Quezler: Ah, i looked through the code i had local and noticed that in a test it also called null, which is 'technically' not a string, but i'll change it :)

Quezler:
👌

Issue #229, opened 2018-04-17
Exp in microseconds


VincentSurelle:
Hi guys,

For informations the 4.0 builder create the exp claim with microseconds which is not valid (see: https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#page-6). 
There should be only second.
It cause token validity troubles with some libraries.
I can a merge request if needed.

lcobucci:
> For informations the 4.0 builder create the exp claim with microseconds which is not valid (see: https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#page-6).
There should be only second.

@VincentSurelle the RCF clearly states that non-integer formats can be used:

> This is equivalent to the IEEE Std 1003.1,
      2013 Edition [POSIX.1] definition "Seconds Since the Epoch", in
      which each day is accounted for by exactly 86400 seconds, **other
      than that non-integer values can be represented.**  See RFC 3339
      [RFC3339] for details regarding date/times in general and UTC in
      particular.

In #171 we've decided to use decimal seconds to make things a bit more precise, we can make the DateTimeImmutable format a bit more configureable but for now you can simply do:

```php
$now = new DateTimeImmutable('@' . time());

$builder->issuedAt($now)
        ->expiresAt($now->modify('+15 minutes'));
```

And builder will use integers.

> It cause token validity troubles with some libraries.

Those libraries should be changed to adheres to what the RFC says 😄

lcobucci:
With that said I'll be closing this as an invalid bug, please reopen it if you feel that we need to discuss things further 👍

PR #230, opened 2018-04-17
Removed microseconds from timestamps

VincentSurelle:
In order to be valid with https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#page-6


lcobucci:
Closing as `wont fix` as explained in #229

PR #239, opened 2018-05-21
Don't require ext-openssl

chalasr:
As it is not necessary when using Hmac signers. Already reflected by Rsa related tests that use the `@requires ext-openssl` annotation.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,8 +17,7 @@
         "BSD-3-Clause"
     ],
     "require": {
-        "php": ">=5.5",
-        "ext-openssl": "*"
+        "php": ">=5.5"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.5",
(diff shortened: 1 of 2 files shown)


lcobucci:
@chalasr I have mixed feelings with this but it does make sense... The best way to handle this IMO would be splitting the package. My concern is more related to testing, there's no good way now to ensure that this will still work as time goes by...

chalasr:
Right, I understand your feeling. 
Btw it looks like @Spomky did the split you're talking about on its jwt-framework, see e.g. https://github.com/web-token/jwt-signature-algorithm-eddsa.
Closing here for now

Spomky:
Hi,

This behaviour is still in development but it is need to limit the dependencies.

Splitting it in several packages is a hard work as there are more and more packages to manage.
But the mono/many repo methodology can help. gitsplit saved my life.
*One [config](https://github.com/web-token/jwt-framework/blob/v1.2/.gitsplit.yml) to rule them, One [app](https://github.com/splitsh/lite) to split them all, One [command](https://github.com/jderusse/docker-gitsplit/#split-your-repo-manualy) to bring them all and in the repo commit them*.
Tests are not so difficult to manage. The package you mention is not a good example because I have not moved them to that sub-repo. This way to proceed offers a good opportunity to write better unit tests, completely isolated and based on the interfaces, not concrete implementations.
Functional tests can be done in the main repo. This is one of my biggest tasks for the next weeks (... months).

Issue #191, opened 2017-08-08
Validation of iat breaks on systems with a clock skew.


viviivanov:
Hi, i am using the packagist 3.2.1 version of the library and noticed a problem which occurs by validating 'iat' by default.

Scenario: 
Issuing server has a skew of +1 second
Communication between server and client is sub-second speed.
Resulting in example timestamps.: $jwt['iat'] = 1001, $clientNow = 1000. 
Which results in.: $token->validate() === false

I did not find a good solution to disable the validation. I don't have control over server and i would rather not adjust reference timestamp in ValidationData::__construct as it also affects nbf and exp checks. There is also no guaranty that the skew will be not more then 5 or 10 seconds and i have no time limit besides the timeframe of exp-nbf but this seems absolutely hacky to start such adjustments with variable time frames based on expand nbf. To stay compatible the change of course should not break existing behavior but a flag which disables 'iat' validation explicitly in ValidationData would suffice imho. Would you accept a PR or be willing to extend the ValidationData::_construct yourself with something like this?

```php
    public function __construct($currentTime = null, $ignoreIAT = false)
    {
        $currentTime = $currentTime ?: time();

        $this->items = [
            'jti' => null,
            'iss' => null,
            'aud' => null,
            'sub' => null,
            'iat' => $currentTime,
            'nbf' => $currentTime,
            'exp' => $currentTime
        ];

    if ($ignoreIAT)
        $this->items['iat'] = null;
    }
```

I am open for any clean solution.
Regards.

Ocramius:
@viviivanov the 4.x branch uses a `Clock` object to handle time issues - can a test be written against that, and then we try backporting it to 3.x?

viviivanov:
Referencing the function from: https://github.com/lcobucci/jwt/blob/master/src/Validation/Constraint/ValidAt.php#L37 
```php
    public function assert(Token $token): void
    {
        $now = $this->clock->now();
        $this->assertIssueTime($token, $now);
        $this->assertMinimumTime($token, $now);
        $this->assertExpiration($token, $now);
    }
```

Changing the clock would result in the same undesired behavior of changing the validation of all three (iat, nbf, exp) parameters. What i actually think would be cleaner is disabling 'iat' validation completely and leave everything else as is.

A test case would test if a switch will turn of 'iat' validation but if there is no switch for it, what should it test :) ?

A testcase would be create a token with 'iat' in the future.

lcobucci:
[...]

PR #248, opened 2018-06-30
Issue #191: Allow leeway to handle clock skew

m777z:
Not sure if I'm doing this right since it's my first real code contribution to an open-source project. I'm sending the PR to 3.3 per @lcobucci's comment on this issue on May 17. 

Note that all the whitespace changes were made by my editor (PHPStorm) when set to the "PSR1/PSR2" predefined style.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,12 +32,13 @@ Just use the builder to create a new JWT/JWS tokens:
 ```php
 use Lcobucci\JWT\Builder;
 
+$time = time();
 $token = (new Builder())->issuedBy('http://example.com') // Configures the issuer (iss claim)
                         ->canOnlyBeUsedBy('http://example.org') // Configures the audience (aud claim)
                         ->identifiedBy('4f1g23a12aa', true) // Configures the id (jti claim), replicating as a header item
-                        ->issuedAt(time()) // Configures the time that the token was issue (iat claim)
-                        ->canOnlyBeUsedAfter(time() + 60) // Configures the time that the token can be used (nbf claim)
-                        ->expiresAt(time() + 3600) // Configures the expiration time of the token (exp claim)
+                        ->issuedAt($time) // Configures the time that the token was issue (iat claim)
+                        ->canOnlyBeUsedAfter($time + 60) // Configures the time that the token can be used (nbf claim)
+                        ->expiresAt($time + 3600) // Configures the expiration time of the token (exp claim)
                         ->with('uid', 1) // Configures a new claim, called "uid"
                         ->getToken(); // Retrieves the generated token
 
@@ -69,7 +70,7 @@ echo $token->getClaim('uid'); // will print "1"
 
 ### Validating
 
-We can easily validate if the token is valid (using the previous token as example):
+We can easily validate if the token is valid (using the previous token and time as example):
 
 ```php
 use Lcobucci\JWT\ValidationData;
@@ -79,15 +80,37 @@ $data->setIssuer('http://example.com');
 $data->setAudience('http://example.org');
 $data->setId('4f1g23a12aa');
 
-var_dump($token->validate($data)); // false, because token cannot be used before of now() + 60
+var_dump($token->validate($data)); // false, because token cannot be used before now() + 60
 
-$data->setCurrentTime(time() + 61); // changing the validation time to future
+$data->setCurrentTime($time + 61); // changing the validation time to future
 
 var_dump($token->validate($data)); // true, because current time is between "nbf" and "exp" claims
 
-$data->setCurrentTime(time() + 4000); // changing the validation time to future
+$data->setCurrentTime($time + 4000); // changing the validation time to future
 
 var_dump($token->validate($data)); // false, because token is expired since current time is greater than exp
+
+// We can also use the $leeway parameter to deal with clock skew (see notes below)
+// If token's claimed time is invalid but the difference between that and the validation time is less than $leeway, 
+// then token is still considered valid
+$dataWithLeeway = new ValidationData($time, 20); 
+$dataWithLeeway->setIssuer('http://example.com');
+$dataWithLeeway->setAudience('http://example.org');
+$dataWithLeeway->setId('4f1g23a12aa');
+
+var_dump($token->validate($dataWithLeeway)); // false, because token can't be used before now() + 60, not within leeway
+
+$dataWithLeeway->setCurrentTime($time + 51); // changing the validation time to future
+
+var_dump($token->validate($dataWithLeeway)); // true, because current time plus leeway is between "nbf" and "exp" claims
+
+$dataWithLeeway->setCurrentTime($time + 3610); // changing the validation time to future but within leeway
+
+var_dump($token->validate($dataWithLeeway)); // true, because current time - 20 seconds leeway is less than exp
+
+$dataWithLeeway->setCurrentTime($time + 4000); // changing the validation time to future outside of leeway
+
+var_dump($token->validate($dataWithLeeway)); // false, because token is expired since current time is greater than exp
 ```
 
 #### Important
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -88,6 +88,27 @@ var_dump($token->validate($data)); // true, because current time is between "nbf
 $data->setCurrentTime(time() + 4000); // changing the validation time to future
 
 var_dump($token->validate($data)); // false, because token is expired since current time is greater than exp
+
+// We can also use leeway to deal with clock skew.
```
Ocramius: I think most readers of the `README.md` will unlikely know what `leeway` or `clock skew` mean in this context. Also, `iat`, `nbf` and `exp`, while in the spec, need some expansion here.

Is this maybe for a separate documentation page? (`README.md` should be short and to the point)

README.md
```diff
@@ -88,6 +88,27 @@ var_dump($token->validate($data)); // true, because current time is between "nbf
...
+// It uses the current time to validate, except that iat will add 5 seconds, nbf will add 10, and exp will subtract 20
+$dataWithLeeway = new ValidationData(null, ['iat' => 5, 'nbf' => 14, 'exp' => -20]); 
+$dataWithLeeway->setIssuer('http://example.com');
+$dataWithLeeway->setAudience('http://example.org');
+$dataWithLeeway->setId('4f1g23a12aa');
+
+var_dump($token->validate($dataWithLeeway)); // false, because token can't be used before now() + 60 and it's now() + 14
+
+$dataWithLeeway->setCurrentTime(time() + 51); // changing the validation time to future
```
Ocramius: Use `$time = time()`, assigned only once to make it clear that we're working within the boundaries of fixed time throughout the example

README.md
```diff
@@ -97,6 +118,11 @@ var_dump($token->validate($data)); // false, because token is expired since curr
 configured in ```ValidationData``` they will be ignored by ```Token::validate()```.
 - ```exp```, ```nbf``` and ```iat``` claims are configured by default in ```ValidationData::__construct()```
 with the current UNIX time (```time()```).
+- The values for the keys in the  optional ```$leeway``` array of ```ValidationData``` will add that number of seconds
```
Ocramius: Same here about `$leeway`. A link is required

src/ValidationData.php
```diff
@@ -24,24 +24,28 @@ class ValidationData
      */
     private $items;
 
+    private $timeFields = ['iat', 'nbf', 'exp'];
```
Ocramius: Docblock with `@var string[]` needed

src/ValidationData.php
```diff
@@ -24,24 +24,28 @@ class ValidationData
...
+
+    private $leeways = [];
```
Ocramius: Docblock with types needed

src/ValidationData.php
```diff
@@ -24,24 +24,28 @@ class ValidationData
...
+
     /**
      * Initializes the object
      *
      * @param int $currentTime
+     * @param array $leeways
```
Ocramius: Please use `type[]` (with the correct type) here

src/ValidationData.php
```diff
@@ -91,9 +95,13 @@ public function setSubject($subject)
      */
     public function setCurrentTime($currentTime)
     {
-        $this->items['iat'] = (int) $currentTime;
-        $this->items['nbf'] = (int) $currentTime;
-        $this->items['exp'] = (int) $currentTime;
+        foreach ($this->timeFields as $field) {
+            if (isset($this->leeways[$field]) && is_int($this->leeways[$field])) {
```
Ocramius: The `is_int()` is to be avoided here - use an assertion where `$this->leeways` is being modified instead, and throw if a non-int is passed in.

src/ValidationData.php
Ocramius: This kind of definition of leeways is a bit squishy and hard to keep in mind, as the user needs to know the keys of the `$leeways` parameters. An explicit parameter for each of the fields supporting `$leeways` is more explicit and avoids basic mistakes.

test/unit/ValidationDataTest.php
```diff
@@ -209,16 +279,18 @@ private function createExpectedData(
...
-            'aud' => $aud !== null ? (string) $aud : null,
-            'sub' => $sub !== null ? (string) $sub : null,
-            'iat' => $time,
-            'nbf' => $time,
-            'exp' => $time
+            'jti' => $id !== null ? (string)$id : null,
+            'iss' => $iss !== null ? (string)$iss : null,
+            'aud' => $aud !== null ? (string)$aud : null,
+            'sub' => $sub !== null ? (string)$sub : null,
+            'iat' => (int)$iat,
```
Ocramius: This cast is likely wrong - should never come in as non-integer here

test/unit/ValidationDataTest.php
```diff
@@ -200,7 +268,9 @@ public function claimValues()
      * @param string $sub
      * @param string $iss
      * @param string $aud
-     * @param int $time
+     * @param int $iat
+     * @param int $nbf
```
Ocramius: `int|null`, I suppose?

test/unit/ValidationDataTest.php
```diff
@@ -200,7 +268,9 @@ public function claimValues()
...
+     * @param int $exp
```
Ocramius: `int|null`, I suppose?

README.md
lcobucci: > Is this maybe for a separate documentation page? (README.md should be short and to the point)

I agree with you @Ocramius but let's do that as a separate PR, we can keep it on the `README.md` for now and modify it later.

m777z:
Thanks for the feedback, guys! Since some pretty big changes are needed, should I continue to send commits to this pull request or just close it and open a new PR when the code is ready?

Ocramius:
Sending patches here would give us an idea if thr comments are addressed:
incremental reviews FTW!

m777z:
@lcobucci I updated the code to only take one integer parameter for the leeway (which is then applied to each of the time-based checks).

@Ocramius The code style changes were reverted. I made some changes to README.md (used your $time = time() suggestion), though as you and @lcobucci said we may eventually want to link to a good explanation elsewhere since I am having trouble succinctly explaining leeway and clock skew. I believe all your comments on src/ValidationData.php have been indirectly addressed by the simplification I made on @lcobucci's suggestion. I also added the |null in test/unit/ValidationDataTest.php and took out the unneeded cast.

Please let me know what further changes need to be made.

lcobucci:
@m777z awesome, thanks for your work!

This week I'll be releasing the last patch on 3.2.x and focus on 3.3, thus your code will be reviewed ASAP

src/ValidationData.php
```diff
@@ -24,24 +24,30 @@ class ValidationData
      */
     private $items;
 
+    /**
+     * The leeway (in seconds) to use when validating time claims
+     * @var int
+     */
+    private $leeway;
```
lcobucci: I was wondering if it makes sense to have this as a property... I mean, it's only used for configuring the current time, maybe it would be better to simply pass it to `ValidationData#setCurrentTime()`... @Ocramius any thoughts?

test/unit/ValidationDataTest.php
```diff
@@ -209,16 +238,18 @@ private function createExpectedData(
...
         return [
             'jti' => $id !== null ? (string) $id : null,
             'iss' => $iss !== null ? (string) $iss : null,
             'aud' => $aud !== null ? (string) $aud : null,
             'sub' => $sub !== null ? (string) $sub : null,
-            'iat' => $time,
-            'nbf' => $time,
-            'exp' => $time
+            'iat' => $iat,
+            'nbf' => $nbf !== null ? $nbf: $iat,
```
lcobucci: It's sad that we can't use `??` to simplify this due to PHP 5.x compatibility :sob:

src/ValidationData.php
m777z: I'm completely fine with doing it either way. Leaving it as it is makes sense if it's fairly common to use the `setCurrentTime()` method 1 or more times after initial construction of the object while keeping the same leeway. However, if users would want to keep changing the leeway without constructing a new object, or if they rarely use `setCurrentTime()` directly at all, then passing leeway as a parameter to `setCurrentTime()` makes more sense.

lcobucci: @m777z I'm have no clue on how users are using this object, and since it will be deprecated very soon I don't think we should debate this much. Let's go with your suggested approach and if needed we change it. Thanks!

m777z: @lcobucci I am assuming you mean it is okay to just leave things as they are for this part. If so, are there other changes I need to make before this pull request is merged?

lcobucci: @m777z that's correct. There're some small CS changes (which I'll handle ASAP), I just didn't manage to work on this. Plan to work on it tomorrow

lcobucci:
@m777z thanks for your contribution (and sorry about the delay!) :smile:

PR #250, opened 2018-07-15
Add newer PHP versions (allowing some to fail)

lcobucci:


.travis.yml
```diff
@@ -3,11 +3,17 @@ php:
...
   - 7.0
+  - 7.1
+  - 7.2
+  - nightly
   - hhvm
   - hhvm-nightly
 
 matrix:
   allow_failures:
+    - php: 7.2
```
Ocramius: This one really shouldn't be allowed to fail

PR #254, opened 2018-07-31
should it change the order of headers? 

fatihwk:
in builder . the heads sequence : 
$this->headers = ['typ'=> 'JWT', 'alg' => 'none'];

should change to : 
$this->headers = [ 'alg' => 'none','typ'=> 'JWT'];

I think it should by alphabetical order,   it is right token, but the token will be different with other libs .


lcobucci:
@fatihwk there's nothing in the specs mentioning the order of the elements, so I'd say this is not really relevant.

lcobucci:
@fatihwk would you mind moving this to an issue instead of a PR? Specially because this PR here doesn't have anything about your suggestion, so I'll be closing it as invalid.

PR #260, opened 2018-08-21
Fixes memory leak in RSA-Signer

Fahrenholz:
A SSL-Key-resource was created and never freed, thus creating a memory leak. This pull request fixes it by freeing the resource.

### DIFF
--- a/src/Signer/Rsa.php
+++ b/src/Signer/Rsa.php
@@ -24,17 +24,19 @@ abstract class Rsa implements Signer
         $key = openssl_get_privatekey($key->getContent(), $key->getPassphrase());
         $this->validateKey($key);
 
-        $signature = '';
+        try {
+            $signature = '';
 
-        if (! openssl_sign($payload, $signature, $key, $this->getAlgorithm())) {
-            throw new InvalidArgumentException(
-                'There was an error while creating the signature: ' . openssl_error_string()
-            );
+            if (! openssl_sign($payload, $signature, $key, $this->getAlgorithm())) {
+                throw new InvalidArgumentException(
+                    'There was an error while creating the signature: ' . openssl_error_string()
+                );
+            }
+
+            return $signature;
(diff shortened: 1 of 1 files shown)


src/Signer/Rsa.php
```diff
@@ -52,7 +55,7 @@
      *
      * @throws InvalidArgumentException
      */
-    private function validateKey($key): void
+    protected function validateKey($key): void
```
Ocramius: Please revert this to `private`

Ocramius: You can revert https://github.com/lcobucci/jwt/pull/260/commits/235d4870b7977d9b29f3fa9fd756411dc294eaf6 to achieve that

Ocramius:
This class is not designed for inheritance: another interface implementation is the correct approach here

Fahrenholz:
@lcobucci Hm, es, that would be good as it creates a new request. Should I open a new PR or should I do it in this one?

lcobucci:
@Fahrenholz sorry about my delay, I have just one more question =)

Do you know of any reliable way to test this?

lcobucci:
@Fahrenholz :ship: 

We can try to add tests in another PR, I'll be backporting this to 3.3 now. Thanks for your help!

Fahrenholz:
@lcobucci : I don't really know of a way to test it unfortunately. I can tell you how I found out: We had a long running process, with reactPHP as a server. After every request, we verified the included JWT in order to authenticate the user. And so after every 100 requests we saw that the PHP process had slightly grown, eventually getting so big every one or two weeks that it killed the container. 

Maybe a strategy would be to performance test it with jmeter or gatling.

Ocramius:
A test that instantiates the stack a thousand times and measures memory at
0, 500 and 1000 cycles (and expexts 1000===500 in memory usage) would be
enough

Issue #204, opened 2017-10-13
Add validation constraints to configuration object


lcobucci:
In order to simplify things when working with the library.

I thought about simply adding:

```php
public function getValidationConstraints(): array;
public function setValidationConstraints(array $constraints): void;
```

daniruizcamacho:
Hi @lcobucci, I have some questions.

What type of value is each element of the constraints array? I mean, I found a class called "Constraint" and I suppose that each element must be an object of Constraint class but I am not sure.

And if this first question is affirmative, Would you validate that each element be "Constraint" and throw an exception when it's not true? (\InvalidArgumentException)

Other question is that I can't run all tests in this moment because the library is not stable in PHP7.2 but I can run the tests related to this issue. Is this enough?

Thanks!

PR #264, opened 2018-09-10
Add validation constraints to configuration object

daniruizcamacho:
This feature is to add validation constraints to configuration object. This feature is related to this opened issue. https://github.com/lcobucci/jwt/issues/204

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -6,6 +6,7 @@ namespace Lcobucci\JWT;
 use Lcobucci\Jose\Parsing;
 use Lcobucci\JWT\Signer\Key;
 use Lcobucci\JWT\Signer\None;
+use Lcobucci\JWT\Validation\Constraint;
 
 /**
  * Configuration container for the JWT Builder and Parser
@@ -49,6 +50,11 @@ final class Configuration
      * @var Validator|null
      */
     private $validator;
+
+    /**
+     * @var Constraint[]
+     */
+    private $validationConstraints = [];
 
     public static function forAsymmetricSigner(
         Signer $signer,
@@ -154,5 +160,18 @@ final class Configuration
     public function setValidator(Validator $validator): void
     {
         $this->validator = $validator;
+    }
+
+    /**
+     * @return Constraint[]
+     */
+    public function getValidationConstraints(): array
+    {
+        return $this->validationConstraints;
+    }
+
+    public function setValidationConstraints(Constraint ...$validationConstraints): void
+    {
+        $this->validationConstraints = $validationConstraints;
     }
 }
--- a/test/unit/ConfigurationTest.php
+++ b/test/unit/ConfigurationTest.php
@@ -7,6 +7,7 @@ use Lcobucci\Jose\Parsing;
 use Lcobucci\JWT\Signer\Key;
 use Lcobucci\JWT\Signer\None;
 use Lcobucci\JWT\Token\Builder as BuilderImpl;
+use Lcobucci\JWT\Validation\Constraint;
 use PHPUnit\Framework\TestCase;
 
 final class ConfigurationTest extends TestCase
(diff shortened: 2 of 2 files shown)


src/Configuration.php
```diff
@@ -50,6 +51,11 @@
      */
     private $validator;
 
+    /**
+     * @var Constraint[]|null
+     */
+    private $constraints;
```
lcobucci: We could configure the default value already here, so that the property would never be `null`.

src/Configuration.php
```diff
@@ -155,4 +161,24 @@ public function setValidator(Validator $validator): void
...
         $this->validator = $validator;
     }
+
+    public function getValidationConstraints(): array
+    {
+        if ($this->constraints === null) {
+            return [];
+        }
+
+        return $this->constraints;
```
lcobucci: This would be simplified if we set the default value. Another way to simplify this method would be:

```php
public function getValidationConstraints(): array
{
    return $this->constraints ?? [];
}
```

I think I'd have the former but am open to discuss possibilities here :)

src/Configuration.php
```diff
@@ -155,4 +161,24 @@ public function setValidator(Validator $validator): void
...
+    }
+
+    public function setValidationConstraints(array $constraints): void
```
lcobucci: I'd rather have a variadic function for this, which would leave this validation up to PHP. Something like:

```php
public function setValidationConstraints(Constraint ...$constraints): void
{
    $this->constraints = $constraints;
}
```

src/Configuration.php
lcobucci: What do you think about renaming the property in order to align to the accessors' names (getter/setter) - rename to `validationConstraints`?

test/unit/ConfigurationTest.php
```diff
@@ -306,4 +313,60 @@ public function getValidatorShouldReturnTheConfiguredValidator(): void
...
+     * @uses \Lcobucci\JWT\Signer\None
+     * @uses \Lcobucci\JWT\Signer\Key
+     */
+    public function getValidationConstraintsShouldReturnTheDefaultWhenItWasNotConfigured(): void
+    {
+        $config                = Configuration::forUnsecuredSigner();
+        $validationConstraints = $config->getValidationConstraints();
+
+        self::assertSame([], $validationConstraints);
+        self::assertNotSame($this->validationConstraints, $validationConstraints);
```
lcobucci: This assertion is not necessary since the one above already validates if the value is exactly `[]`, so I'd remove it.

lcobucci: Which would allow us to inline the variable `$ValidationConstraints`, since it has a single usage.

test/unit/ConfigurationTest.php
```diff
@@ -306,4 +313,60 @@ public function getValidatorShouldReturnTheConfiguredValidator(): void
...
+     * @test
+     *
+     * @covers \Lcobucci\JWT\Configuration::getValidationConstraints
+     *
+     * @uses \Lcobucci\JWT\Configuration::forUnsecuredSigner
+     * @uses \Lcobucci\JWT\Configuration::__construct
+     * @uses \Lcobucci\JWT\Signer\None
+     * @uses \Lcobucci\JWT\Signer\Key
+     */
+    public function getValidationConstraintsShouldReturnTheDefaultWhenItWasNotConfigured(): void
```
lcobucci: I'd suggest to rename this method to `getValidationConstraintsShouldReturnAnEmptyArrayWhenItWasNotConfigured`, since it makes our test intention a bit more explicit

test/unit/ConfigurationTest.php
```diff
@@ -306,4 +313,60 @@ public function getValidatorShouldReturnTheConfiguredValidator(): void
...
+     * @covers \Lcobucci\JWT\Configuration::setValidationConstraints
+     *
+     * @uses \Lcobucci\JWT\Configuration::forUnsecuredSigner
+     * @uses \Lcobucci\JWT\Configuration::__construct
+     * @uses \Lcobucci\JWT\Signer\None
+     * @uses \Lcobucci\JWT\Signer\Key
+     */
+    public function setValidationConstraintsShouldThrowExceptionWhenConstraintsArrayIsNotValid(): void
+    {
+        self::expectException(\InvalidArgumentException::class);
```
lcobucci: This would not be an `InvalidArgumentException` but rather a type error if we use the variadic form.

src/Configuration.php
daniruizcamacho: Yes, totally agree. I will rename this var and the param in the setter function.

src/Configuration.php
daniruizcamacho: Yes, it's simpler. I use here the guard clauses pattern sometimes it's more clear. https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

src/Configuration.php
daniruizcamacho: Very interesting. I never use variadic on this way. Thanks for the tip.

test/unit/ConfigurationTest.php
daniruizcamacho: Sorry, this is an error from my part.

daniruizcamacho:
@lcobucci Don't worry about the issue, it's very difficult to have some time to answer all things. All comments are addressed.

PR #270, opened 2018-10-14
Remove assertion count manipulation

lcobucci:
Using `@doesNotPerformAssertions` is preferred since it doesn't rely on PHPUnit's internals.


lcobucci:
Coverage decrease happened due to PHPUnit's bug fixed in: https://github.com/sebastianbergmann/phpunit/pull/3348

lcobucci:
I was wrong about what's "preferred" and this must be reverted

Issue #133, opened 2016-10-28
Add deprecation notes on methods


lcobucci:
To help users migrate to v4

lcobucci:
Even adding `Builder` methods

MasterOdin:
Is there plans to backport the changes in v4 to avoid the now deprecated sign/unsign functions?

lcobucci:
@MasterOdin definitely, I'm working on that already. I decided to release v3.3 to solve the incompatibility with newer PHP versions but v3.4 isn't far away.

PR #295, opened 2019-05-23
Fix deprecation notes

lcobucci:
As pointed out by @MasterOdin in #133 we were a bit too eager to add the deprecation notes.

This solves that, allowing users to migrate in a smooth way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,7 @@ Hmac signatures are really simple to be used:
 
 ```php
 use Lcobucci\JWT\Builder;
+use Lcobucci\JWT\Signer\Key;
 use Lcobucci\JWT\Signer\Hmac\Sha256;
 
 $signer = new Sha256();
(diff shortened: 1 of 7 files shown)


lcobucci:
@MasterOdin can you please let me know if this solves the issue you raised? I don't want for this lib to annoy people with useless deprecation notes.

MasterOdin:
Yeah, this looks like it would solve my issue in that there's a way to use non-deprecated functions for generating a token! 👍 

One thing I noticed while looking through this is that the 3.3 release uses [Builder::with](https://github.com/lcobucci/jwt/blob/3.3/src/Builder.php#L338-L347) while 4.0 uses [Builder::withClaim](https://github.com/lcobucci/jwt/blob/master/src/Builder.php#L62).

lcobucci:
@MasterOdin well spotted! I'll fix it too. Thanks!

lcobucci:
@MasterOdin released :+1:

MasterOdin:
Thanks so much! :+1:

PR #319, opened 2019-09-26
adds tests for ES512 algorithm

ffflabs:
This PR adds keys and test to check compatibility with `ecdsa-sha2-nistp521` signature algorithm

```sh
openssl ecparam -genkey -name secp521r1 -noout -out test/_keys/ecdsa/private_ec512.key
openssl pkey -in test/_keys/ecdsa/private_ec512.key -pubout -out test/_keys/ecdsa/public_ec512.key
```


test/functional/ES512TokenTest.php
```diff
@@ -0,0 +1,280 @@
...
+            Sha512::create(),
+            static::$ecdsaKeys['private_ec512'],
+            static::$ecdsaKeys['public_ec512']
+        );
+    }
+
+    /**
+     * @test
+     *
+     * @covers \Lcobucci\JWT\Configuration
```
Ocramius: Hmm, I'd leave this at class level, and only put coverage on the ecdsa classes

test/functional/ES512TokenTest.php
```diff
@@ -0,0 +1,280 @@
...
+            ['http://client.abc.com', 'http://client2.abc.com'],
+            $token->claims()->get(Token\RegisteredClaims::AUDIENCE)
+        );
+
+        return $token;
+    }
+
+    /**
+     * @test
+     * @depends builderCanGenerateAToken
```
Ocramius: `@depends` can likely be skipped overall

Ocramius (approved):

test/functional/ES512TokenTest.php
lcobucci: Not really since it depends on the value returned by it :)

test/functional/ES512TokenTest.php
lcobucci: We should do this for all functional tests indeed, it would simplify everything. Sounds like a different PR, though :+1:

lcobucci (approved):
@amenadiel thanks for your contribution! Out of curiosity, what motivated you to create the PR? Were you comparing features with another implementation or just checking if things would work?

ffflabs:
@lcobucci hey guys, I'm glad my PR was accepted.

I detected a huge flaw in JWT verification in a webapp that handles sensitive data. The assessment also raised the need to switch the current symmetric implementation with a more secure one.

The app provider agreed to patch the "none" issue but he didn't see the need to ditch HS256 considering  RS256 was breached as well on [Crown Sterling demonstration](https://www.crownsterling.io/2019/09/crown-sterling-decrypts-rsa-asymmetric-public-keys-in-live-demonstration/) so it was pointless. I spent the next hours creating a POC demonstrating how to embrace RS512 given the EC algorithms aren't just a matter of bits, for which I decided to write the missing test and here it is.

(I'd love to do the same for ed25529 if you guys ever implement it and I find out how to generate the pubout key for it)

Issue #312, opened 2019-09-19
Question: how to get the whole JWT string to send back over HTTP


tommasodargenio:
Hi,

this could be a very stupid question but I'm struggling a bit to understand how can I get out of the object the whole JWT string that I need to send back to the HTTP post request from the client.

I've used the sample code as provided in your readme to create the token, but the getToken method doesn't give me the plain text but return the whole object. I can see the token string and the signature in the object payload but shouldn't the getToken method return the plain text version I can send back to the client?

Thanks

Ocramius:
`$token->__toString()` or `(string) $token`:

https://github.com/lcobucci/jwt/blob/a11ec5f4b4d75d1fcd04e133dede4c317aac9e18/src/Token.php#L269-L283

lcobucci:
@tommasodargenio the `Token` class has the `__toString()` method, which means that you get the string representation by casting it to string.

PR #320, opened 2019-09-26
Add method getToken on Token.php

EdouardTack:
Add method to return the token and not the entire object

### DIFF
--- a/src/Token.php
+++ b/src/Token.php
@@ -271,7 +271,7 @@ class Token
      *
      * @return string
      */
-    public function __toString()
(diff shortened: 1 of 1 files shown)


hizmarck:
Hello @EdouardTack , I want to understand what happens with this issue, was closed because it's not useful? or how can I track the correct path to know whether it is in the next release or not?

Thanks for your support,

hizmarck:
I've noticed that answer https://github.com/lcobucci/jwt/issues/312, thanks and sorry for that.

PR #329, opened 2019-10-16
Added multiple audiences support

AlvaroLarumbe:
Version 3.3 does not support multiple audiences for validating a token as RFC says (https://tools.ietf.org/html/rfc7519#section-4.1.3), it should support more than one.
I noticed new version will support multiple audiences, but meanwhile, here it is a PR fixing it on branch 3.4.
I created a new method `setAudiences()` in order to not break current API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ use Lcobucci\JWT\ValidationData;
 $data = new ValidationData(); // It will use the current time to validate (iat, nbf and exp)
 $data->setIssuer('http://example.com');
 $data->setAudience('http://example.org');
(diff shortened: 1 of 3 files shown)


lcobucci:
@AlvaroLarumbe thanks for your contribution! The only issue is that the current validation API is being deprecated and replaced by the [one we have in `master`](https://lcobucci-jwt.readthedocs.io/en/latest/validating-tokens/) - which already supports that btw.

This means that I'd really avoid adding anything to the current one. I hope you understand this decision :+1:

lcobucci:
Closing as `won't fix` as per my comment.

PR #331, opened 2019-10-22
Update Builder.php

gouguoyin:
add  a function withClaims which can configure some claim items in bulk.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -372,6 +372,25 @@ class Builder
 
         return $this;
     }
+    
+    /**
+     * Configures some claim items
+     *
+     * @param array $claims
+     *
+     * @return Builder
+     */
+    public function withClaims($claims)
+    {
+        if(!is_array($claims) || empty($claims)){
+            return $this;
+        }
+        foreach ($claims as $name => $value) {
+            $this->claims[(string) $name] = $this->claimFactory->create($name, $value);
+        }
+
+        return $this;
+    }
 
     /**
      * Configures a claim item


lcobucci:
I think that we don't need to have such method in the builder's interface because **I prefer** to have everything explicitly defined (types, intention). Passing an array of claims may open the interface too much, forcing us to do some validation inside such method that might be not very useful.

gouguoyin:
Thank you for your reply. How to batch assign the data what query from the database  is usually an array？

gouguoyin:
Thank you for your reply. How to batch assign an array of data queried from database into claims?

gouguoyin:
Thank you for your reply. How to batch assign the data  into claims?The data which is queried from the database is usually an array.

lcobucci:
> How to batch assign the data into claims? The data which is queried from the database is usually an array.

@gouguoyin thanks, that does make things clearer but I'm curious about which claims you're setting from DB. Could you elaborate on that?

src/Builder.php
```diff
@@ -372,6 +372,25 @@ public function withClaim($name, $value)
...
+     * Configures some claim items
+     *
+     * @param array $claims
+     *
+     * @return Builder
+     */
+    public function withClaims($claims)
+    {
+        if(!is_array($claims) || empty($claims)){
+            return $this;
```
jakubboucek: This is not a fail-safe way to handle invalid input. It will cause silent error, user has no any notification about invalid input. Please use method type hint for argument `$claims` like this:
```php
public function withClaims(array $claims)
```

lcobucci:
Closing this for since we didn't have any update and I fail to see the value here. Thanks for the input anyway :smile:

PR #337, opened 2019-11-21
Add more files to .gitattributes

reedy:


.gitattributes
```diff
@@ -1,10 +1,15 @@
...
 /.gitattributes   export-ignore
+/.github          export-ignore
 /.gitignore       export-ignore
 /.scrutinizer.yml export-ignore
 /.travis.yml      export-ignore
 /CONTRIBUTING.md  export-ignore
 /*.dist           export-ignore
+/mkdocs.yml       export-ignore
 /phpbench.json    export-ignore
+/phpcs.xml.dist   export-ignore
```
lcobucci: These ones should be covered by `*.dist`, aren't them?

reedy: Haha. Yeah. I think I got confused looking at what was coming in via composer from version 3.3.1 (which didn't have a .gitattributes at all)

PR #351, opened 2020-03-31
Validator: raise exception when no constraint is given

Slamdunk:
Hi, today I expected a token to fail, but the assertion made as described in https://lcobucci-jwt.readthedocs.io/en/latest/validating-tokens/ didn't raise any error.
After a while I found no build-in Constraint is provided (and that's ok for me), but also no error is raised when this happen.

I'm scared that many users would fall into my mistake, a pretty dangerous mistake!

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -12,6 +12,10 @@ final class Validator implements \Lcobucci\JWT\Validator
      */
     public function assert(Token $token, Constraint ...$constraints): void
     {
+        if ($constraints === []) {
+            throw new NoConstraintsGiven('No constraint given.');
+        }
+
         $violations = [];
 
         foreach ($constraints as $constraint) {
@@ -40,6 +44,10 @@ final class Validator implements \Lcobucci\JWT\Validator
 
     public function validate(Token $token, Constraint ...$constraints): bool
     {
+        if ($constraints === []) {
+            throw new NoConstraintsGiven('No constraint given.');
+        }
+
         try {
             foreach ($constraints as $constraint) {
                 $constraint->assert($token);
--- a/test/unit/Validation/ValidatorTest.php
+++ b/test/unit/Validation/ValidatorTest.php
@@ -26,6 +26,20 @@ final class ValidatorTest extends TestCase
      * @test
      *
      * @covers \Lcobucci\JWT\Validation\Validator::assert
+     */
+    public function assertShouldRaiseExceptionWhenNoConstraintIsGiven(): void
+    {
+        $validator = new Validator();
+
+        $this->expectException(NoConstraintsGiven::class);
+
+        $validator->assert($this->token, ...[]);
+    }
+
+    /**
+     * @test
+     *
+     * @covers \Lcobucci\JWT\Validation\Validator::assert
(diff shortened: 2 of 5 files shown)


src/Validation/Validator.php
```diff
@@ -12,6 +13,10 @@ final class Validator implements \Lcobucci\JWT\Validator
      */
     public function assert(Token $token, Constraint ...$constraints): void
     {
+        if (count($constraints) === 0) {
```
lcobucci, line 16: I know this is kind of micro optimisation, but let's avoid using `count()` here.
We can do `if ($constraints === [])`.

src/Validation/Validator.php
```diff
@@ -40,6 +45,10 @@ private function checkConstraint(
 
     public function validate(Token $token, Constraint ...$constraints): bool
     {
+        if (count($constraints) === 0) {
```
lcobucci, line 48: I know this is kind of micro optimisation, but let's avoid using `count()` here.
We can do `if ($constraints === [])`.

test/unit/Validation/ValidatorTest.php
```diff
@@ -22,6 +22,20 @@ public function createDependencies(): void
...
+     *
+     * @covers \Lcobucci\JWT\Validation\Validator::assert
+     */
+    public function assertShouldRaiseExceptionWhenNoConstraintIsGiven(): void
+    {
+        $validator = new Validator();
+
+        $this->expectException(NoContraintGiven::class);
+
+        $validator->assert($this->token, ...[]);
```
lcobucci, line 36: We can simply omit the second argument, makes things simpler here :+1:

test/unit/Validation/ValidatorTest.php
```diff
@@ -71,6 +85,18 @@ public function assertShouldNotRaiseExceptionWhenNoConstraintFails(): void
...
+    /**
+     * @test
+     *
+     * @covers \Lcobucci\JWT\Validation\Validator::validate
+     */
+    public function validateShouldReturnFalseWhenNoConstraintIsGiven(): void
+    {
+        $validator = new Validator();
+
+        self::assertFalse($validator->validate($this->token, ...[]));
```
lcobucci: We can simply omit the second argument, makes things simpler here :+1:

test/unit/Validation/ValidatorTest.php
Slamdunk: Actually I made it intentional.

It needs to be obvious to anyone that here I'm passing something, empty but something.

If someone would ever refactor `assert` method with a new syntax or type hint, we must prevent he/she to look at this test and think "Oh, this test doesn't provide the second parameter, but now it is required by code/tool/whatever, so I can delete it".

I don't know if I'm clear here: the test shouldn't be simple; the test should express the intention.

Slamdunk:
Everything done, expect the `...[]` one :smile_cat:

test/unit/Validation/ValidatorTest.php
lcobucci: Fair point :+1:

lcobucci:
@Slamdunk :ship:

PR #355, opened 2020-05-12
add toString method

p4veI:
* add a standard public toString method alongside the magic __toString methods

motivation:
* helps to avoid explicit type casts over Token/Plain class in v4

### DIFF
--- a/src/Token.php
+++ b/src/Token.php
@@ -51,5 +51,5 @@ interface Token
     /**
      * Returns an encoded representation of the token
      */
-    public function __toString(): string;
+    public function toString(): string;
 }
--- a/src/Token/DataSet.php
+++ b/src/Token/DataSet.php
@@ -49,7 +49,7 @@ final class DataSet
         return $this->data;
     }
 
-    public function __toString(): string
+    public function toString(): string
     {
         return $this->encoded;
     }
--- a/src/Token/Signature.php
+++ b/src/Token/Signature.php
@@ -34,7 +34,7 @@ final class Signature
     /**
      * Returns the encoded version of the signature
      */
-    public function __toString(): string
+    public function toString(): string
     {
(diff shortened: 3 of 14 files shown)


src/Token/DataSet.php
```diff
@@ -49,8 +49,13 @@ public function all(): array
...
-    public function __toString(): string
+    public function toString(): string
     {
         return $this->encoded;
     }
+
+    public function __toString(): string
+    {
+        return $this->toString();
+    }
```
lcobucci, line 60: It think it feels pointless to have this here if we won't use it internally.

src/Token/Signature.php
```diff
@@ -34,8 +34,13 @@ public function hash(): string
...
-    public function __toString(): string
+    public function toString(): string
     {
         return $this->encoded;
     }
+
+    public function __toString(): string
+    {
+        return $this->toString();
+    }
```
lcobucci, line 45: Same remark, we can just get rid of this method

src/Token.php
```diff
@@ -51,5 +51,10 @@ public function isExpired(DateTimeInterface $now): bool;
     /**
      * Returns an encoded representation of the token
      */
+    public function toString(): string;
+
+    /**
+     * Returns an encoded representation of the string while casting Token to string
+     */
     public function __toString(): string;
```
lcobucci, line 59: Let's really drop it.
On v3.4 we mark it as `@deprecated` and introduce `toString()` as well.

p4veI: was worried it wouldn't be approved if I drop it :) cheers, will update now.

lcobucci: Thanks, it was a bad decision back then. I just had bigger fishes to fry :sweat_smile:

PR #368, opened 2020-08-20
`Claim::getValue` can return anything

### LINKED ISSUE #367: [3.3] `Claim::getValue` can return an array
Hi @lcobucci 👋 

I'd like to report to you something that we found at @usabilla while using this amazing library 😄 

The documentation for 

https://github.com/lcobucci/jwt/blob/56f10808089e38623345e28af2f2d5e4eb579455/src/Claim.php#L27-L32

is missing the possibility of an `array`. Example: https://jwt.io/#debugger-io?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyLCJteUN1c3RvbUFycmF5Q2xhaW0iOlsidmFsdWUxIiwidmFsdWUyIl19.krhErcq8V7_W2K4dSlXtp63d8AY4LTwdU08b6DlR_rU

Can I open a PR adding it, or would it be a BC?

I found that with PHPStan, telling me that I can't use functions like `is_array` or `is_string` and they would be always `false` or `true`, respectively.

carusogabriel:
As reported and discussed via #367, we need to expand the return type of `Claim::getValue`

An example is, https://github.com/lcobucci/jwt/blob/56f10808/src/Claim/Basic.php#L36, which accepts `mixed` as the Claim's value.

Fix #367.

### DIFF
--- a/src/Claim.php
+++ b/src/Claim.php
@@ -27,7 +27,7 @@ interface Claim extends JsonSerializable
     /**
      * Returns the claim value
      *
-     * @return string
+     * @return mixed
      */
     public function getValue();
 


lcobucci (approved):

lcobucci:
I need to disable scrutinizer, so slow :facepalm:

lcobucci:
@carusogabriel thanks for the fix, will release it ASAP!

PR #375, opened 2020-08-29
Agregando header

ArmChaCib:


Ocramius:
I don't think a claim  is to be looked up in headers at all.

PR #422, opened 2020-09-27
Update `getPassphrase` PHPDoc

Spomky:
The property `passphrase` can be `null` or a string.
I propose to indicate it in the header.

If you really want to return a string, the constructor could be changed in favour of `public function __construct($content, $passphrase = '')`
Let me know if you prefer the later.

### DIFF
--- a/src/Signer/Key.php
+++ b/src/Signer/Key.php
@@ -31,7 +31,7 @@ final class Key
      * @param string $content
      * @param string $passphrase
      */
-    public function __construct($content, $passphrase = null)
+    public function __construct($content, $passphrase = '')
     {
         $this->setContent($content);
         $this->passphrase = $passphrase;


src/Signer/Key.php
```diff
@@ -83,7 +83,7 @@ public function getContent()
     }
 
     /**
-     * @return string
+     * @return null|string
```
Ocramius, line 86: This contradicts the type of the property, which also needs to be adjusted

Ocramius (dismissed):

src/Signer/Key.php
Spomky: OK so in this case I will change the default argument in the constructor.

Spomky: I think it’s preferable to set an empty string because other methods/functions expect to receive a string as argument (e.g. [OpenSSL::getPrivateKey](https://github.com/lcobucci/jwt/blob/3.4/src/Signer/OpenSSL.php#L44))

lcobucci (approved):
Thanks @Spomky!

PR #452, opened 2020-10-12
Use library specific exceptions

### LINKED ISSUE #128: Use library specific exceptions

t0mmy742:
I added a new Exception class named `InvalidArgument` in order to replace the SPL `InvalidArgumentException` class.
Tests have been updated.

Fix #128

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 namespace Lcobucci\JWT;
 
 use DateTimeImmutable;
-use InvalidArgumentException;
 use Lcobucci\JWT\Signer\Key;
 use Lcobucci\JWT\Token\Plain;
 
@@ -57,7 +56,7 @@ interface Builder
      *
      * @param mixed $value
      *
-     * @throws InvalidArgumentException When trying to set a registered claim.
+     * @throws InvalidArgument When trying to set a registered claim.
      */
     public function withClaim(string $name, $value): Builder;
 
--- a/src/InvalidArgument.php
+++ b/src/InvalidArgument.php
@@ -1,0 +1,10 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT;
+
+use InvalidArgumentException;
+
+final class InvalidArgument extends InvalidArgumentException implements Exception
+{
+}
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -3,14 +3,12 @@ declare(strict_types=1);
 
 namespace Lcobucci\JWT;
 
-use InvalidArgumentException;
-
 interface Parser
 {
     /**
      * Parses the JWT and returns a token
      *
-     * @throws InvalidArgumentException
+     * @throws InvalidArgument
      */
     public function parse(string $jwt): Token;
 }
--- a/src/Signer.php
+++ b/src/Signer.php
@@ -3,7 +3,6 @@ declare(strict_types=1);
 
 namespace Lcobucci\JWT;
 
-use InvalidArgumentException;
 use Lcobucci\JWT\Signer\Key;
 
 interface Signer
@@ -16,14 +15,14 @@ interface Signer
     /**
      * Creates a hash for the given payload
      *
-     * @throws InvalidArgumentException When given key is invalid.
+     * @throws InvalidArgument When given key is invalid.
      */
     public function sign(string $payload, Key $key): string;
 
     /**
      * Returns if the expected hash matches with the data and key
      *
-     * @throws InvalidArgumentException When given key is invalid.
+     * @throws InvalidArgument When given key is invalid.
      */
     public function verify(string $expected, string $payload, Key $key): bool;
 }
--- a/src/Validation/Constraint/ValidAt.php
+++ b/src/Validation/Constraint/ValidAt.php
@@ -5,8 +5,8 @@ namespace Lcobucci\JWT\Validation\Constraint;
 
 use DateInterval;
 use DateTimeInterface;
-use InvalidArgumentException;
 use Lcobucci\Clock\Clock;
+use Lcobucci\JWT\InvalidArgument;
 use Lcobucci\JWT\Token;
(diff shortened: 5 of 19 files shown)


src/InvalidArgument.php
```diff
@@ -0,0 +1,8 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT;
+
+final class InvalidArgument extends Exception
```
Ocramius, line 6: To reduce the BC surface, this should extend `InvalidArgumentException`

Ocramius (changes requested):
Good improvement overall: wish we were more detailed about exception types, but this is a good start nonetheless 👍

src/InvalidArgument.php
t0mmy742: The main goal here is to not depend on SPL exception class. I'm extending this class from the Exception class of this library.

We can still do BC break since we are only in alpha stage.

lcobucci: I'm sorry that my poor explanation led to this small confusion. The scope of the problem isn't the dependence on SPL but rather the fact that we directly expose it.

We're surely handling a new major release but that doesn't mean we **need** to break the BC everywhere. This kind of change is definitely one that can be compatible. and should IMHO.

lcobucci:
I agree with breaking down exceptions a bit more, we can handle that in another step.

src/InvalidArgument.php
lcobucci, line 6: As of https://github.com/lcobucci/jwt/pull/462 we should also implement the base exception `Lcobucci\JWT\Exception`

t0mmy742:
It should be better since we have #462.

t0mmy742:
Updated to resolve conflicts with #508

driesvints:
@t0mmy742 seems like the branch is out of date with the base branch (See the warning on the PR).

t0mmy742:
> @t0mmy742 seems like the branch is out of date with the base branch (See the warning on the PR).

There is no conflict with the actual master branch. Merging is still possible. It's only because of #509

src/Builder.php
```diff
@@ -57,7 +56,7 @@ public function withHeader(string $name, $value): Builder;
      *
      * @param mixed $value
      *
-     * @throws InvalidArgumentException When trying to set a registered claim.
+     * @throws InvalidArgument When trying to set a registered claim.
```
lcobucci, line 59: This should be something like `Lcobucci\JWT\Token\InvalidNameForCustomClaim`

src/Parser.php
```diff
@@ -3,14 +3,12 @@
...
 
-use InvalidArgumentException;
-
 interface Parser
 {
     /**
      * Parses the JWT and returns a token
      *
-     * @throws InvalidArgumentException
+     * @throws InvalidArgument
```
lcobucci, line 11: We have a few scenarios where `InvalidArgumentException` was being thrown:

1. Strings with incorrect basic JWT structure
1. Incorrect type for headers/claims
1. Unsupported encrypted tokens

I reckon the first two could be grouped but the third is quite a different thing. Maybe we should have 2 different exceptions here...

src/Signer.php
```diff
@@ -16,14 +15,14 @@ public function getAlgorithmId(): string;
     /**
      * Creates a hash for the given payload
      *
-     * @throws InvalidArgumentException When given key is invalid.
+     * @throws InvalidArgument When given key is invalid.
```
lcobucci, line 18: In this case, I believe we had many things under the same exception type:

1. Unable to read the file contents
1. Key isn't parseable by OpenSSL
1. Key isn't compatible (ECDSA vs RSA keys)
1. Error whilst creating the signature (OpenSSL)
1. Invalid ASN1 sequence for ECDSA signature

src/Signer.php
```diff
@@ -16,14 +15,14 @@ public function getAlgorithmId(): string;
...
      */
     public function sign(string $payload, Key $key): string;
 
     /**
      * Returns if the expected hash matches with the data and key
      *
-     * @throws InvalidArgumentException When given key is invalid.
+     * @throws InvalidArgument When given key is invalid.
```
lcobucci, line 25: In this case, I believe we had many things under the same exception type:

1. Unable to read the file contents
1. Key isn't parseable by OpenSSL
1. Key isn't compatible (ECDSA vs RSA keys)
1. Invalid signature length for ECDSA

src/Validation/Constraint/ValidAt.php
```diff
@@ -29,7 +29,7 @@ private function guardLeeway(?DateInterval $leeway): DateInterval
         }
 
         if ($leeway->invert === 1) {
-            throw new InvalidArgumentException('Leeway cannot be negative');
+            throw new InvalidArgument('Leeway cannot be negative');
```
lcobucci, line 32: This should be something like `Lcobucci\JWT\Validation\Constraint\LeewayCannotBeNegative`

lcobucci (changes requested):
@t0mmy742 thanks again for all your help.

These are some high-level thoughts about my poor exception choices on the past and what I believe we should do to consider the original issue as solved.

If you have the time and want to help let us know, otherwise I'll be playing around with this soon.

src/Parser.php
lcobucci: Don't have proper suggestions atm, though (shaaame)

src/Signer.php
lcobucci: Don't have proper suggestions atm, though (shaaame)

src/Signer.php
lcobucci: Don't have proper suggestions atm, though (shaaame)

lcobucci:
@t0mmy742 thanks a lot for pushing this one too. I've broken down all the exceptions types and decided to replace this PR, sorry about not using your work but I think that having specific exceptions is better for our API.

PR #454, opened 2020-10-13
Allow users to configure format for date registered claims

### LINKED ISSUE #231: Allow users to configure format for date registered claims
As stated in #229 we could be more flexible regarding the format use we generate the `iat`, `exp`, and `nbf` claims.

t0mmy742:
Add an boolean option to the `Configuration` class, and then sent it to the `Builder` class. Default is to keep current behaviour and, if set to `false`, date registered claims will only have seconds.
The new parameter is not the last of the constructor (potentially BC break here) as it could be a parameter changed more ofter than `$encoder` or `$decoder`.
Maybe you have another idea of how to implement this.

Fix #231

### DIFF
--- a/src/Token/Builder.php
+++ b/src/Token/Builder.php
@@ -6,16 +6,13 @@ namespace Lcobucci\JWT\Token;
 use DateTimeImmutable;
 use InvalidArgumentException;
 use Lcobucci\JWT\Builder as BuilderInterface;
+use Lcobucci\JWT\ClaimsFormatter;
 use Lcobucci\JWT\Encoder;
 use Lcobucci\JWT\Signer;
 use Lcobucci\JWT\Signer\Key;
 
 use function array_diff;
-use function array_intersect;
-use function array_keys;
 use function array_merge;
-use function count;
-use function current;
 use function in_array;
 
 final class Builder implements BuilderInterface
@@ -27,10 +24,12 @@ final class Builder implements BuilderInterface
     private array $claims = [];
 
     private Encoder $encoder;
+    private ClaimsFormatter $claimFormatter;
 
-    public function __construct(Encoder $encoder)
+    public function __construct(Encoder $encoder, ClaimsFormatter $claimFormatter)
     {
-        $this->encoder = $encoder;
+        $this->encoder        = $encoder;
+        $this->claimFormatter = $claimFormatter;
     }
 
     public function permittedFor(string ...$audiences): BuilderInterface
@@ -115,7 +114,7 @@ final class Builder implements BuilderInterface
         $headers['alg'] = $signer->getAlgorithmId();
 
         $encodedHeaders = $this->encode($headers);
-        $encodedClaims  = $this->encode($this->formatClaims($this->claims));
+        $encodedClaims  = $this->encode($this->claimFormatter->formatClaims($this->claims));
 
         $signature        = $signer->sign($encodedHeaders . '.' . $encodedClaims, $key);
         $encodedSignature = $this->encoder->base64UrlEncode($signature);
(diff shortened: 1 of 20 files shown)


t0mmy742:
I'm still not sure about my tests, as I'm using other parts of this library.

src/Token/Builder.php
```diff
@@ -28,9 +28,12 @@ final class Builder implements BuilderInterface
 
     private Parsing\Encoder $encoder;
 
-    public function __construct(Parsing\Encoder $encoder)
+    private bool $dateClaimsWithMicroseconds;
+
+    public function __construct(Parsing\Encoder $encoder, bool $dateClaimsWithMicroseconds)
```
lcobucci, line 33: I don't think using a boolean here is the way to go... it surely works but it's a smells like something that doesn't belong here.

What do you think about extracting the process of formatting claims into an interface, change `Builder#getToken()` to receive an instance that implements that interface, and change the configuration object to allow users to customise that?

We'd ship here only the formatter using microseconds IMHO but users are free to provide their own implementation and change any claim they desire prior to the encoding process.

Something like:

```php
interface ClaimFormatter
{
    /**
     * @param array<string, mixed> $claims
     *
     * @return array<string, mixed>
     */
    public function formatClaims(array $claims): array;
}

// Terrible name but I hope you can understand the intent
final class AudienceAndDatesWithMicroseconds implements ClaimFormatterintent
{
    public function formatClaims(array $claims): array
    {
        if (count($claims[RegisteredClaims::AUDIENCE]) === 1) {
            $claims[RegisteredClaims::AUDIENCE] = current($claims[RegisteredClaims::AUDIENCE]);
        }

        foreach (array_intersect(RegisteredClaims::DATE_CLAIMS, array_keys($claims)) as $claim) {
            $claims[$claim] = $this->convertDate($claims[$claim]);
        }

        return $claims;
    }

    /** @return int|string */
    private function convertDate(DateTimeImmutable $date)
    {
        $seconds      = $date->format('U');
        $microseconds = $date->format('u');

        if ((int) $microseconds === 0) {
            return (int) $seconds;
        }

        return $seconds . '.' . $microseconds;
    }
}
```

The usage would then be:

```php
$now   = new DateTimeImmutable();

$config->createBuilder()
    ->issuedBy('http://example.com')
    ->issuedAt($now)
    ->expiresAt($now->modify('+1 hour'))
    ->withClaim('uid', 1)
    ->getToken($config->getSigner(), $config->getSigningKey(), $config->getClaimFormatter());
```

What I also like about this design is that it makes it simpler for people to issue tokens with different formatters without changing their configuration.

lcobucci (commented):
@t0mmy742 first of all, thanks a lot for your contribution and sorry about my delay!

What do you think about this alternative approach to the problem? I think it would be more extensible and flexible - and would have a smaller impact on existing tests.

src/Token/Builder.php
t0mmy742: I really like this implementation, and how you extracted the process of formatting claims into an interface. I agree about giving only one implementation of `ClaimFormatter` with this library, at least for now.

Arguments of `Builder#getToken` could be member of the `Builder` class. Maybe we could think about it for the future.

Just one more thing about the `formatClaims` method : from the [RFC](https://tools.ietf.org/html/rfc7519#section-4.1.3), the `aud` claim is optionnal. We should use `isset` function before `count`.

lcobucci:
@t0mmy742 thanks for applying the changes! I'm not fully convinced about my suggestion of putting the dependency on the `Token#getToken()`... I've got a few minutes now, so I'll play around with your branch.

What I'm thinking now is to have something like this:

```php
final class Configuration
{
    // ...

    public function createBuilder(?ClaimFormatter $formatter = null): Builder
    {
        return ($this->builderFactory)($formatter ?? new DefaultFormatter());
    }

    // ...
}
```

That way we reduce the impact of the change and make things simpler for the users.

Again, thanks a lot for your help. We'll soon get this one merged!

lcobucci:
@t0mmy742 shipped and documented: https://lcobucci-jwt.readthedocs.io/en/latest/extending-the-library/#claims-formatter

It's a bit late so I may have added some typos on the docs :sweat_smile:
I'd love to hear your feedback on the final solution (sorry if I modified too much of your original code).

lcobucci:
And, sure, thank you again for helping out here!

t0mmy742:
> @t0mmy742 shipped and documented: https://lcobucci-jwt.readthedocs.io/en/latest/extending-the-library/#claims-formatter
> 
> It's a bit late so I may have added some typos on the docs 😅
> I'd love to hear your feedback on the final solution (sorry if I modified too much of your original code).

It seems good for me. Your `ChainedFormatter` is a good idea. We can now change date claims without changing everything for example.

PR #544, opened 2020-11-24
Only prefix file:// if doesn't exist already

Sephster:
This small change only prefixes key paths with _file://_ if the prefix doesn't exist already.

Apologies but I haven't added any tests for this. If you would like me to, please just shout and I will try to get to them tomorrow. Cheers Luis

### DIFF
--- a/src/Signer/Key/LocalFileReference.php
+++ b/src/Signer/Key/LocalFileReference.php
@@ -6,9 +6,13 @@ namespace Lcobucci\JWT\Signer\Key;
 use Lcobucci\JWT\Signer\Key;
 
 use function file_exists;
+use function strpos;
+use function substr;
 
 final class LocalFileReference implements Key
 {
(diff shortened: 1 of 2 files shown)


src/Signer/Key/LocalFileReference.php
```diff
@@ -30,7 +31,7 @@ public static function file(string $path, string $passphrase = ''): self
 
     public function contents(): string
     {
-        return 'file://' . $this->path;
+        return strpos($this->path, 'file://') === 0 ? $this->path : 'file://' . $this->path;
```
lcobucci, line 34: This makes sense, I'd move it to the named constructor so it's done only once.

lcobucci (changes requested):
We do need tests but I can handle it!

lcobucci (approved):

PR #546, opened 2020-11-24
Use magic method for toString

Sephster:
I suspect this might have already been suggested but wondered if it would be possible to change the `toString` methods to use the magic method instead? No worries if this idea has already been rejected, just saves a bit of code for a lazy dev like myself :D

### DIFF
--- a/src/Token.php
+++ b/src/Token.php
@@ -51,5 +51,5 @@ interface Token
     /**
      * Returns an encoded representation of the token
      */
-    public function toString(): string;
+    public function __toString(): string;
 }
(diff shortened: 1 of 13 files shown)


Ocramius (changes requested):
Overall preferable to have (and use) `toString()`: rationale can be found at https://github.com/ShittySoft/symfony-live-berlin-2018-doctrine-tutorial/pull/3#issuecomment-460441229

Sephster:
Closing after discussion with @lcobucci

lcobucci:
Just to have our discussion publicly available...

Marco's comment summarises my feelings about `__toString()` quite well.
I've used it a lot in the past but observed pretty much the same things he describes, which is why I've decided to move away from it for v4.0.

Issue #228, opened 2018-04-06
Getting claims from Token


JoeBengalen:
Hi, I noticed that the `Plain` token has a [`claims` method](https://github.com/lcobucci/jwt/blob/master/src/Token/Plain.php#L49), but it's not part of the `Token` interface.

Is this missing on purpose? If not I could create PR to add it if you like.

```php
interface Token {
    public function claims(): DataSet;
}
```

lcobucci:
@JoeBengalen it was made like that in purpose.

The general idea is that the token can either be encrypted or plain but only plain tokens have readable claims, and since the token you receive is already something that cannot really be trusted (user input) you have to be explicit regarding your expectations. Like:

```php
$token = $parser->parse($jwtFromSomewhere);

if (! $token instanceof Plain) {
    throw new \RuntimeException('NOPE!');
}
```

Sure that's a very basic validation but you could rely on our new validator to do what you want and simply use [`assert()`](https://secure.php.net/manual/en/function.assert.php) to be IDE friendly (please remember that ideally `zend.assertions` should be set to `-1` in production environment), like:

```php
$token = $parser->parse($jwtFromSomewhere);
assert($token instanceof Plain);

if (! $validator->validate($token, ...$myConstraints)) {
    throw new \RuntimeException('NOPE!');
}

// $token->claims() will be available here
```

We don't support JWE yet but this was designed as a forward compatibility layer for future implementation.

---

Please don't forget that `v4.0` is not stable yet and the public interface will still suffer some minor changes - thanks you for trying it out, though!

lcobucci:
Linking #203 and #186

lcobucci:
@JoeBengalen please let me know if that offers you enough information!

JoeBengalen:
Yes, thank you. I could've done a search on previous issues maybe :sweat_smile:

lcobucci:
@JoeBengalen sorry for my delay on getting back to you, my idea is to release a new alpha of v4 in a week or two. I'll organise the milestones ASAP to ensure that things make more sense 👍

Issue #203, opened 2017-10-13
How get Claims after encode token


lpj145:
I'm try to get claim on middleware.
Using Version 4.0@dev
![image](https://user-images.githubusercontent.com/11980109/31547582-406dfa60-affe-11e7-990a-95531931d87e.png)
But `parse` function return Plain, this is a error ?
![image](https://user-images.githubusercontent.com/11980109/31547642-87568622-affe-11e7-843d-d12d8cd92e27.png)

lcobucci:
@lpj145 that's a forward compatibility layer. As stated on #6 we plan to support encrypted tokens (which will implement that interface).

Since `parse()` deals with untrusted data, users are supposed to verify if they have the expected type of token, like:

```php
$token = $this->parser->parse($jwt);

if (! $token instanceof Plain) {
    // You should have your own exception for that, sure =)
    throw new \RuntimeException('The provided token is not valid');
}
```

lcobucci:
btw don't forget to validate the token properly using the validator 😉

lpj145:
yep i'm using:
```
/**
     * @param Plain $token
     * @return bool
     */
    private function tokenIsValid(Plain $token)
    {
        if ($token->isExpired((new SystemClock())->now())) {
            return false;
        }

        try {
            $this->validator->assert($token, new SignedWith($this->signer, $this->key));
        }catch (ConstraintViolationException $exception) {
            return false;
        }

        return true;
    }
```

lcobucci:
@lpj145 if you're not using the exception, I'd suggest to use this instead (I really need to work on the documentation 😭 ):

```php
return $this->validator->validate(
    $token,
    new ValidAt(new SystemClock()),
    new SignedWith($this->signer, $this->key)
);
```

The benefit with that approach is that the validator will stop its process when the first error comes and the constraints will be processed in the order you called the method, so you can leave the "expensive" operations to the end of your validation chain.

I'd also suggest to have the list of constraints injected, so it makes it easier for you (we're planning to add it to the `Configuration` object):

```php
return $this->validator->validate($token, ...$this->constraints);
```

lcobucci:
@lpj145 let me know if your question has been answered and we can close this issue 👍

lpj145:
I think this is a good solution, but need tests. in hour i answer to you.
```
public function process(ServerRequestInterface $request, DelegateInterface $delegate)
    {
        if (
            $this->middlewareConfig['secure'] &&
[...]

PR #552, opened 2020-11-25
Parser: return covariant type

Slamdunk:
This allows for precise static analysis


lcobucci:
@Slamdunk this has been discussed in https://github.com/lcobucci/jwt/issues/228#issuecomment-379228725 and https://github.com/lcobucci/jwt/issues/203#issuecomment-336453180.

We want to be able to implement encrypted tokens without having to break the interface. Users are expected to validate the output's type.

It's also important to mention that users can provide their own Parser now if they want to, that also opens the possibility of creating their own token classes (granted the interface is implemented).

Do you have any suggestion on how to still achieve these goals?

Slamdunk:
Gotcha

PR #555, opened 2020-11-25
Add StrictValidAt constraint, deprecate ValidAt to LooseValidAt

### LINKED ISSUE #554: [v4.0] Add stricter ValidAt constraint
Hi, it seems to me that the new `ValidAt` constraint is very dangerous:

https://github.com/lcobucci/jwt/blob/6d8665ccd924dc076a9b65d1ea8abe21d68f6958/test/unit/Validation/Constraint/ValidAtTest.php#L218-L225

It creates a false sense of security, and it's easy to mess with: I've already had a successful token misused because no time claim was set.

I propose to rename `ValidAt` to `LooseValidAt`, and introduce a new `StrictValidAt` that requires all time claims to be set

Slamdunk:
Close https://github.com/lcobucci/jwt/issues/554

### DIFF
--- a/test/unit/Validation/Constraint/ValidAtTestCase.php
+++ b/test/unit/Validation/Constraint/ValidAtTestCase.php
@@ -1,0 +1,205 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Validation\Constraint;
+
+use DateInterval;
+use DateTimeImmutable;
+use Lcobucci\Clock\Clock;
+use Lcobucci\Clock\FrozenClock;
+use Lcobucci\JWT\Token\RegisteredClaims;
+use Lcobucci\JWT\Validation\Constraint;
+use Lcobucci\JWT\Validation\ConstraintViolation;
+
+abstract class ValidAtTestCase extends ConstraintTestCase
+{
+    protected Clock $clock;
+
+    /** @before */
+    final public function createDependencies(): void
+    {
+        $this->clock = new FrozenClock(new DateTimeImmutable());
+    }
+
+    abstract protected function buildValidAtConstraint(Clock $clock, ?DateInterval $leeway = null): Constraint;
+
+    /**
+     * @test
+     *
+     * @covers ::__construct
+     * @covers ::guardLeeway
+     * @covers \Lcobucci\JWT\Validation\Constraint\LeewayCannotBeNegative
+     */
+    final public function constructShouldRaiseExceptionOnNegativeLeeway(): void
+    {
+        $leeway         = new DateInterval('PT30S');
+        $leeway->invert = 1;
+
+        $this->expectException(LeewayCannotBeNegative::class);
+        $this->expectExceptionMessage('Leeway cannot be negative');
+
+        $this->buildValidAtConstraint($this->clock, $leeway);
+    }
+
+    /**
+     * @test
+     *
+     * @covers ::__construct
+     * @covers ::guardLeeway
+     * @covers ::assert
+     * @covers ::assertExpiration
+     * @covers ::assertIssueTime
+     * @covers ::assertMinimumTime
+     *
+     * @uses \Lcobucci\JWT\Token\DataSet
+     * @uses \Lcobucci\JWT\Token\Plain
+     * @uses \Lcobucci\JWT\Token\Signature
+     */
+    final public function assertShouldRaiseExceptionWhenTokenIsExpired(): void
+    {
+        $now = $this->clock->now();
+
+        $claims = [
+            RegisteredClaims::ISSUED_AT => $now->modify('-20 seconds'),
+            RegisteredClaims::NOT_BEFORE => $now->modify('-10 seconds'),
+            RegisteredClaims::EXPIRATION_TIME => $now->modify('-10 seconds'),
+        ];
+
+        $this->expectException(ConstraintViolation::class);
+        $this->expectExceptionMessage('The token is expired');
+
+        $constraint = $this->buildValidAtConstraint($this->clock);
+        $constraint->assert($this->buildToken($claims));
+    }
+
+    /**
+     * @test
+     *
+     * @covers ::__construct
+     * @covers ::guardLeeway
+     * @covers ::assert
+     * @covers ::assertIssueTime
+     * @covers ::assertMinimumTime
+     *
+     * @uses \Lcobucci\JWT\Token\DataSet
+     * @uses \Lcobucci\JWT\Token\Plain
+     * @uses \Lcobucci\JWT\Token\Signature
+     */
+    final public function assertShouldRaiseExceptionWhenMinimumTimeIsNotMet(): void
+    {
+        $now = $this->clock->now();
+
+        $claims = [
+            RegisteredClaims::ISSUED_AT => $now->modify('-20 seconds'),
+            RegisteredClaims::NOT_BEFORE => $now->modify('+40 seconds'),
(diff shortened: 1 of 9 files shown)


Slamdunk:
I'm busy now, I'll end this this week

phpstan.neon.dist
```diff
@@ -6,3 +6,11 @@ parameters:
...
     ignoreErrors:
         - '#.*OpenSSLAsymmetricKey.*#'
+        -
+            message:
+                """
+                    #^Instantiation of deprecated class Lcobucci\\\\JWT\\\\Validation\\\\Constraint\\\\ValidAt\\:
+                    Use \\\\Lcobucci\\\\JWT\\\\Validation\\\\Constraint\\\\LooseValidAt$#
+                """
+            count: 1
+            path: test/unit/Validation/Constraint/ValidAtTest.php
```
lcobucci, line 16: Please use `@phpstan-ignore-line` or `@phpstan-ignore-next-line` to simplify the error management here.

lcobucci:
> I'm busy now, I'll end this this week

That's alright, thanks for looking into it!

Slamdunk:
@lcobucci there's a catch we need to talk about before coding forward.

Within the `v4.0.0` API it is not possible to have both `LooseValidAt` and `StrictValidAt`, take a look at `\Lcobucci\JWT\Token\Plain::isExpired`:

https://github.com/lcobucci/jwt/blob/6d8665ccd924dc076a9b65d1ea8abe21d68f6958/src/Token/Plain.php#L77-L84

This corresponds to the behavior of `LooseValidAt`. If we want to introduce the `StrictValidAt` **without** letting it know anything of `claims` (which the API doesn't expose), we must either:

1. Throw an exception when the claim is absent, handled differently by `LooseValidAt` and `StrictValidAt`
1. Add a `$strict` parameter to `isExpired` API, set differently by `LooseValidAt` and `StrictValidAt`

I don't know if the former can be tagged as a MINOR release; the latter can, with a default value, but it would make it a bit ugly.

A third option is to stick with current `ValidAt` constraint, change the behavior of `\Lcobucci\JWT\Token\Plain::isExpired` to return `true` when the claim is absent, thus making `ValidAt` the `StrictValidAt` we are looking for; this could be seen as a PATCH release, since the code had somewhat a bug. (cc @Ocramius ?)

This option makes even more sense regarding to `hasBeenIssuedBefore` and `isMinimumTimeBefore`:

https://github.com/lcobucci/jwt/blob/6d8665ccd924dc076a9b65d1ea8abe21d68f6958/src/Token/Plain.php#L67-L75

For which the absence of the claim can totally be considered a bug as per https://3v4l.org/J7OGi since I'm sure it wasn't an intended and expected behavior.

What do we do?

lcobucci:
@Slamdunk after putting some thoughts here I think I disagree with the points on this discussion.

> For which the absence of the claim can totally be considered a bug as per https://3v4l.org/J7OGi since I'm sure it wasn't an intended and expected behavior.

`exp`, `iat`, and `nbf` are **optional** according to the RFC. Therefore, my interpretation is that the token validation shouldn't fail with the absence of these claims.
Which means that this is not a bug but the expected behaviour. Although it could be made more explicit on the implementation, both `hasBeenIssuedBefore()` and `isMinimumTimeBefore` are expected to return `true` when these claims are absent.

That "loose" behaviour is consistent with other JWT libs as well.

---

With that said I don't think we actually need a `StrictValidAt`. 

I see two possibilities to solve this this:
1. Have strict constraints for each claim (e.g.: `NotExpired`, `NotIssuedInTheFuture`, and `AvailableForUsage` - these names are a bit meh but are the best I can thing of now 🤣)
1. Have a constraint that makes sure the token has a given claim (e.g.: `RequiredClaim` or `RequiredClaims`);

It's okay for us to do like `SignedWith` does and validate the Token type https://github.com/lcobucci/jwt/blob/master/src/Validation/Constraint/SignedWith.php#L24-L26 but we can create an interface for `Plain#claims()`, `Plain#signature()`, and `Plain#payload()`.

The first option would basically remove the need for using `ValidAt`: `$validator->validate($token, new NotExpired($clock), new SignedWith($signer, $key), ...);`.
The second option relies on `ValidAt` for checking the date claims: `$validator->validate($token, new RequiredClaim('exp'), new RequiredClaim('nbf'), new ValidAt($clock), new SignedWith($signer, $key), ...);`

BTW, the constraints aren't mutually exclusive, we can add them all but there's not much point in using `NotExpired` and `ValidAt` - which is something that must be documented.

@Slamdunk @Ocramius thoughts?

Slamdunk:
> Although it could be made more explicit on the implementation

Yeah, if we add an explicit `if (null === $claim) return true;` I'm ok with the default behavior.
We just need to be clear with the developer of the future.

> That "loose" behaviour is consistent with other JWT libs as well.

Side, useless note: ok, that's right, I agree, but it's far from being a good thing, JWT is known to be a f-up standard in term of fail-safe implementations.

> It's okay for us to do like `SignedWith` does and validate the Token type

Huh, didn't know that, this solves the issue right away.

Regarding your proposed solutions, I believe the end-user craves for a one that's:

1. Easy to understand
1. Easy to use
1. Safe by default
1. Unaware of implementation details

Having "strict constraints for each claim" isn't easy to use (imho) and not much safe by default, someone could forget to add one of them.

Having "a constraint that makes sure the token has a given claim" is not easy to understand nor unaware of the implementation.

So in the end I think we still need the `StrictValidAt` and the `ValidAt`-to-`LooseValidAt` edit.

The only downside to this is that all 3 claims must be set, while for example as of yet `storageless` library only uses two of them.
But to me adding a third mandatory claim is a good trade for an eady-to-use, safe-by-default validator.

I'm going to code it as soon as I can, while waiting for @Ocramius reply.

phpstan.neon.dist
Slamdunk: Done

test/unit/Validation/Constraint/ValidAtTestCase.php
```diff
@@ -0,0 +1,205 @@
...
+
+use DateInterval;
+use DateTimeImmutable;
+use Lcobucci\Clock\Clock;
+use Lcobucci\Clock\FrozenClock;
+use Lcobucci\JWT\Token\RegisteredClaims;
+use Lcobucci\JWT\Validation\Constraint;
+use Lcobucci\JWT\Validation\ConstraintViolation;
+
+abstract class ValidAtTestCase extends ConstraintTestCase
```
Slamdunk, line 14: Common tests between `LooseValidAt` and `StrictValidAt` are now written once here.
Implementations are copy-pasted though, without a common base class.

lcobucci (approved):
@Slamdunk you raise valid points (again) - especially about the criticism towards JWT.

Thanks for your contribution and for this fantastic way to debate ideas in a very constructive way. I'll rebase this branch and merge it!

Slamdunk:
@lcobucci `4.1.0` milestone has some heavy issues, like JWK and encrypted token, with no implementation yet.

Can this PR be released and those open issues moved to `4.2.0`?

lcobucci:
> @lcobucci `4.1.0` milestone has some heavy issues, like JWK and encrypted token, with no implementation yet.
> 
> Can this PR be released and those open issues moved to `4.2.0`?

I'll probably do that. I have a talk to present this weekend, so I'll focus on preparing for it. Once that's finished I'll handle this.

Slamdunk:
No hurry, I was just looking for a plan :+1:

Issue #559, opened 2020-11-25
The new behaviour of the method withClaim breaks our adapters


Spomky:
Hi,

In our application, we have adapters for several JWT libraries including this one.
The headers and claims we pass to the adapters are already well-formatted as per the RFCs.

The issue comes from the new version 3.4 (and 4.0 as well) and the [new restriction added here](https://github.com/lcobucci/jwt/blob/3.4.0/src/Builder.php#L414-L416).

As we pass registered/custom claims, the method `withClaim` was fine with it.
With the version 3.4+, an exception is thrown and breaks our application… which is something that should not append for a minor release.

The only way we’ve found is to create a mapping that checks if it is a registered claim or not and call the correct method. This situation is not great at all as it adds complexity and hardcoded method name constants…

What we suggest is to keep the previous behaviour i.e. no exception by default in 3.4:

```php
public function withClaim(string $name, $value, bool $strict = false): BuilderInterface
{
    if ($strict && in_array($name, RegisteredClaims::ALL, true)) {
        throw RegisteredClaimGiven::forClaim($name);
    }

    return $this->setClaim($name, $value);
}
```

And in v4.x, keep the current behaviour by setting `$strict` to `true`:

```php
public function withClaim(string $name, $value, bool $strict = true): BuilderInterface
{
    if ($strict && in_array($name, RegisteredClaims::ALL, true)) {
        throw RegisteredClaimGiven::forClaim($name);
    }

    return $this->setClaim($name, $value);
}
```

lcobucci:
Yeah, throwing the exception on v3.4 breaks BC and it's a mistake (sorry about that).

The guard is there because of expectations about the types used in the registered claims (arrays/DateTimeImmutable/etc).

I'd propose a different approach to this:

```php
public function withClaim(string $name, $value): BuilderInterface
{
    if (in_array($name, RegisteredClaims::ALL, true)) {
        trigger_error('A nice deprecation message', E_USER_DEPRECATED); // just to alert users
    }

    if ($name === RegisteredClaims::AUDIENCE) {
        return $this->permittedFor($value);
    }

    // and the same for the other registered claims

    return $this->setClaim($name, $value);
}
```

This means that your app will be just fine with v3.4 but will need to be updated for v4.0. Is that a reasonable approach?

Spomky:
Hi @lcobucci,

Looks good to me. We will manage that when updating the adapters for v4.0+.
I've just submitted a PR with the code you proposed.

[...]

PR #561, opened 2020-11-26
Remove RegisteredClaimGiven exception in favour of deprecation message for withClaim

Spomky:
This PR is proposed as per the discussion in #559.

### DIFF
--- a/test/unit/BuilderTest.php
+++ b/test/unit/BuilderTest.php
@@ -604,6 +604,46 @@ class BuilderTest extends TestCase
         self::assertEquals($expected, $token->claims()->get($name));
     }
 
+
+    /**
+     * @test
+     *
+     * @param string $name
+     * @param mixed $value
+     * @param mixed $expected
+     * @param null|string $otherMessage
+     *
+     * @covers ::__construct
+     * @covers ::set
+     * @covers ::canOnlyBeUsedAfter
+     * @covers ::configureClaim
+     * @covers ::convertItems
+     * @covers ::convertToDate
+     * @covers ::getToken
+     * @covers ::setRegisteredClaim
+     * @covers ::createSignature
+     * @covers ::expiresAt
+     * @covers ::issuedBy
+     * @covers ::identifiedBy
+     * @covers ::permittedFor
+     * @covers ::forwardCallToCorrectClaimMethod
+     * @covers ::issuedAt
+     *
+     * @dataProvider dataWithClaimDeprecationNotice
+     */
+    public function setShouldSendDeprecationNoticeWhenTryingToConfigureARegisteredClaim($name, $value, $expected, $otherMessage = null)
+    {
+        if ($otherMessage) {
+            $this->expectDeprecation($otherMessage);
+        }
+
+        $token = $this->createBuilder()
+            ->set($name, $value)
+            ->getToken(new None(), Key\InMemory::plainText(''));
+
+        self::assertEquals($expected, $token->claims()->get($name));
+    }
+
     public function dataWithClaimDeprecationNotice()
     {
         $now = time();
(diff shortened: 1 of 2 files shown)


test/unit/BuilderTest.php
```diff
@@ -567,12 +567,27 @@ public function withClaimMustKeepAFluentInterface()
...
+    {
+        return [
+            ['sub', 'me'],
+            ['aud', 'him'],
+            ['jti', '0123456789ABCDEF'],
+            ['iss', 'you'],
+            ['exp', time()+3600],
+            ['iat', time()],
+            ['nbf', time()],
+        ];
```
lcobucci, line 593: We should also have the expected values here (and use `self::assertEquals($expected, $token->claims()->get($name))` on the test):

```suggestion
        $now = time();
        $nowAsDate = new DateTimeImmutable('@' . $now);

        return [
            ['sub', 'me', 'me'],
            ['aud', 'him', ['him']],
            ['jti', '0123456789ABCDEF', '0123456789ABCDEF'],
            ['iss', 'you', 'you'],
            ['exp', $now+3600, $nowAsDate->modify('+1 hour')],
            ['iat', $now, $nowAsDate],
            ['nbf', $now, $nowAsDate],
        ];
```

Spomky: Thank you for the review.
The tests have been updated accordingly.

test/unit/BuilderTest.php
```diff
@@ -564,15 +564,61 @@ public function withClaimMustKeepAFluentInterface()
...
 
-        $this->createBuilder()->withClaim('sub', 'me');
+        return [
+            ['sub', 'me', 'me'],
+            ['aud', 'him', ['him']],
+            ['jti', '0123456789ABCDEF', '0123456789ABCDEF'],
+            ['iss', 'you', 'you'],
+            ['exp', $nowPlus1HourAsDate->getTimestamp(), $nowPlus1HourAsDate->getTimestamp()],
+            ['iat', $now, $nowAsDate->getTimestamp()],
+            ['nbf', $now, $nowAsDate->getTimestamp()],
```
lcobucci, line 620: This is incorrect as we rely on objects for the new API.

test/unit/BuilderTest.php
```diff
@@ -564,15 +564,61 @@ public function withClaimMustKeepAFluentInterface()
...
+
+        $this->expectDeprecation('The use of the method "withClaim" is deprecated for registered claims. Please use dedicated method instead.');
+
+        $token = $this
+            ->createBuilder()
+            ->withClaim($name, $value)
+            ->getToken($signer, $key)
+        ;
+
+        self::assertEquals($expected, $token->claims()->get($name));
```
lcobucci, line 604: Please use the public API instead of mocking here (create the token using `getToken()` as done in other tests).

test/unit/BuilderTest.php
Spomky: The problem is that the test fails when setting $nowAsDate

test/unit/BuilderTest.php
Spomky: Oh OK! I mised that method :+1:

test/unit/BuilderTest.php
lcobucci: It will fail until we add the other`if` statements in https://github.com/lcobucci/jwt/pull/561/files#diff-aaf0bc0443d3390951e65b1d6dadf6134554550b510609434969424757f1b6f9R418 (we need to forward the call to all specific methods for registered claims)

Spomky: Should be fine now

lcobucci (approved):

lcobucci:
@Spomky thanks a lot :heart:

Issue #560, opened 2020-11-26
Allow an array for audience


waltertamboer:
Hi,

I appreciate your work a lot! Thank you!

I ran into an issue where it's not allowed to set an array as audience. I believe this is not in line with the specs:

> In the general case, the "aud" value is an array of case-sensitive strings, each containing a StringOrURI value.  In the special case when the JWT has one audience, the "aud" value MAY be a single case-sensitive string containing a StringOrURI value.

https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4.1.3

https://github.com/lcobucci/jwt/blob/3.4/src/Builder.php#L111
https://github.com/lcobucci/jwt/blob/master/src/Validation/Constraint/PermittedFor.php#L14

Thanks again for your hard work.

b-water:
Hey, i got the same Issue. Audience in JWT is an Array with Element 0, but the lib tries to compare it to a string.

lcobucci:
@waltertamboer v3.x is supposed to only support a single entry in the `aud`, in v4.x you may set multiple entries by calling `Builder#permittedFor()` [multiple times or with multiple elements](https://github.com/lcobucci/jwt/blob/6d8665ccd924dc076a9b65d1ea8abe21d68f6958/src/Token/Builder.php#L35-L41#L20).

If you do need multiple entries on `aud` you should upgrade to v4.0.

> Hey, i got the same Issue. Audience in JWT is an Array with Element 0, but the lib tries to compare it to a string.

@b-water that looks like a bug during the backport 🤦 would be able to send a PR with a test reproducing the issue?

waltertamboer:
Thanks! Will upgrade and try it out.

b-water:
@lcobucci hope this helps!

lcobucci:
@b-water it does, thanks ❤️

lcobucci:
Handled by d446df1fa1b61e8d35a901779988f2574b23c150 (v4.0 only).

Closing here, folks.

PR #562, opened 2020-11-26
Fix validation of the audience claim on the new API

b-water:
As described in #560 our jwt audience contains an array with a single element which breaks the validation

### DIFF
--- a/test/unit/Validation/Constraint/PermittedForTest.php
+++ b/test/unit/Validation/Constraint/PermittedForTest.php
@@ -42,7 +42,7 @@ final class PermittedForTest extends ConstraintTestCase
         $this->expectExceptionMessage('The token is not allowed to be used by this audience');
 
         $constraint = new PermittedFor('test.com');
-        $constraint->assert($this->buildToken([RegisteredClaims::AUDIENCE => 'aa.com']));
+        $constraint->assert($this->buildToken([RegisteredClaims::AUDIENCE => ['aa.com']]));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


test/unit/Validation/Constraint/PermittedForTest.php
```diff
@@ -74,4 +74,28 @@ public function assertShouldNotRaiseExceptionWhenAudienceMatches()
...
+        // works
+        $aud        = uniqid('aud_', true);
+        $token      = $this->buildToken([RegisteredClaims::AUDIENCE => $aud]);
+        $constraint = new PermittedFor($aud);
+
+        $constraint->assert($token);
+        $this->addToAssertionCount(1);
+
+        // doesnt work
+        $token      = $this->buildToken([0 => [RegisteredClaims::AUDIENCE => $aud]]);
```
lcobucci, line 95: This should be like this right?
```suggestion
        $token      = $this->buildToken([RegisteredClaims::AUDIENCE => [$aud]]);
```

lcobucci:
I'd suggest to use the `CompatibilityLayerTest` to approach this from a functional level first. My tired brain missed that validation on https://github.com/lcobucci/jwt/blob/3.4/test/functional/CompatibilityLayerTest.php.

test/unit/Validation/Constraint/PermittedForTest.php
b-water: I tried to recreate how our jwt is parsed in the test. As you can see in the screenshots its parsed like [0 => ] [] https://imgur.com/a/Co0l1Ss

lcobucci: That's what my suggestion does :)

`[0 => [RegisteredClaims::AUDIENCE => $aud]]` would create a DataSet with the claim `0` which isn't exactly what you want 👍

lcobucci:
@b-water :ship: 

Thanks for your help!

PR #568, opened 2020-11-27
Silenced deprecation errors

GrahamCampbell:
This will fix all the issues people have been posting recently. Firing silenced deprecation errors is the convention. See, for example, Symfony. :)

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -148,7 +148,7 @@ class Builder
     private function convertToDate($value)
     {
         if (! $value instanceof DateTimeImmutable) {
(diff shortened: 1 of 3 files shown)


lcobucci:
I already explained before, that's not the approach I want to take for this package.

Deprecation aren't errors and shouldn't be treated like errors, regardless if they come from the language or from a lib.

Thanks for your time on sending this PR, though.

GrahamCampbell:
I am confused as to your approach here. Aren't you treating them like errors?

lcobucci:
@GrahamCampbell I'm treating them just like the language does. Please read carefully the explanation I gave on the issue (and on click the links).

GrahamCampbell:
Hmm, I still don't think I agree that is the best approach, from a practical perspective.

PR #569, opened 2020-11-28
support php8

dyaskur:
Please allow php 8 on 3.4 version, because kreait/firebase-php require it

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "BSD-3-Clause"
     ],
     "require": {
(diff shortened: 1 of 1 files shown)


Spomky (commented):
What about `php": ">=5.6",` ?
The test matrix should be updated as well to confirm et runs on PHP8

lcobucci:
That wasn't added on purpose, php 8 users should use v4. That's why v3.4 ships the forward compatibility layer.

Check how both oauth2 server and passport have solved the matter.

PR #592, opened 2020-12-01
php8 fixes for 3.4 branch

bradymiller:
Several fixes for PHP8 on 3.4 branch

1. Critical fix is the support for OpenSSLAsymmetricKey, which is new in PHP8 (leveraged the mechanism used in master)
2. The openssl_free_key fixes are to prevent the deprecation notices in PHP8 (didn't really follow master mechanism here :) )

### DIFF
--- a/src/Signer/OpenSSL.php
+++ b/src/Signer/OpenSSL.php
@@ -26,7 +26,10 @@ abstract class OpenSSL extends BaseSigner
 
             return $signature;
         } finally {
-            openssl_free_key($privateKey);
+            if (!version_compare(phpversion(), '8.0', '>=')) {
+                // Deprecated and no longer necessary as of PHP >= 8.0
+                openssl_free_key($privateKey);
+            }
         }
     }
 
@@ -54,7 +57,11 @@ abstract class OpenSSL extends BaseSigner
     {
         $publicKey = $this->getPublicKey($key->getContent());
         $result    = openssl_verify($payload, $expected, $publicKey, $this->getAlgorithm());
-        openssl_free_key($publicKey);
+
+        if (!version_compare(phpversion(), '8.0', '>=')) {
+            // Deprecated and no longer necessary as of PHP >= 8.0
+            openssl_free_key($publicKey);
+        }
 
         return $result === 1;
     }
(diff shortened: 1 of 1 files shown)


lcobucci:
v3.x is closed for bugs only and doesn't support PHP 8 by design.

Please follow our upgrade guide to migrate to v4.0.

bradymiller:
hi @lcobucci ,
Thanks for the quick feedback. I think I misunderstood your prior comment here (mistakenly thought goal for 3.4 branch was to port back changes to support PHP 7.3-8):
https://github.com/thephpleague/oauth2-server/pull/1146#issuecomment-729464933

Since 4.0 requires php7.4 (unless I am mistaken), it will force lots of downstream packages to stay on 3.4 and not support PHP8 for another full year (when 7.3 is no longer officially supported). The critical fix is the only thing I have noted so far that breaks in php8 (the removal of php log entries is just fluff and not a big deal and not needed) while testing our solution which uses phpleague/oauth2-server : https://github.com/openemr/openemr/pull/4069

Since our package (OpenEMR), and also guessing oauth2-server (@Sephster ), can't drop support for 7.3 for another year I'm happy to spend time on this, if possible (ie. if there are other places in 3.4 branch that need testing/fixing for php8).

lcobucci:
@bradymiller thanks for providing additional context, and sorry if I was too blunt before (I way too tired).

The back porting was related to functionality only. On v3.4 we have a forward compatibility layer, making its interface compatible with v4.0.

The strategy for you is to require both 3.4 and 4.0 on your composer.json, use configuration object as a compatibility layer (to instantiate the objects correctly), and use the new way to interact with the API.

That is the same approach taken by oauth2-server and passport.

If you have the luxury of shipping a new major release, I'd recommend letting users define the JWT configuration object - it should simplify your code.

I hope that helps. I'll check the PR on OpenEMR to see if I can help you folks there.

bradymiller:
@lcobucci , no prob on your feedback (i also experience the symptoms (fatigue, tiredness, heavy coffee input, etc) of open source at times :) ). I think the crux of the issue is that OpenEMR builds the packages for end users, so the pre-built package ideally need to support all active versions of PHP (now 7.3-8). I put more context on the OpenEMR PR response.

PR #605, opened 2020-12-10
Implement EdDSA Signer

Slamdunk:
Hope to see this in `4.1.0` release

### DIFF
--- a/src/Signer/Eddsa.php
+++ b/src/Signer/Eddsa.php
@@ -1,0 +1,36 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Signer;
+
+use Lcobucci\JWT\Signer;
+use SodiumException;
+
+use function sodium_crypto_sign_detached;
+use function sodium_crypto_sign_verify_detached;
+
+final class Eddsa implements Signer
+{
+    public function algorithmId(): string
+    {
+        return 'EdDSA';
+    }
+
+    public function sign(string $payload, Key $key): string
+    {
+        try {
+            return sodium_crypto_sign_detached($payload, $key->contents());
+        } catch (SodiumException $sodiumException) {
+            throw new InvalidKeyProvided($sodiumException->getMessage(), 0, $sodiumException);
+        }
+    }
+
+    public function verify(string $expected, string $payload, Key $key): bool
+    {
+        try {
+            return sodium_crypto_sign_verify_detached($expected, $payload, $key->contents());
+        } catch (SodiumException $sodiumException) {
+            throw new InvalidKeyProvided($sodiumException->getMessage(), 0, $sodiumException);
+        }
+    }
+}
--- a/test/functional/EddsaTokenTest.php
+++ b/test/functional/EddsaTokenTest.php
@@ -1,0 +1,141 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\FunctionalTests;
+
+use Lcobucci\JWT\Configuration;
+use Lcobucci\JWT\Keys;
+use Lcobucci\JWT\Signer\Eddsa;
+use Lcobucci\JWT\Signer\InvalidKeyProvided;
+use Lcobucci\JWT\Signer\Key\InMemory;
+use Lcobucci\JWT\Token;
+use Lcobucci\JWT\Validation\Constraint\SignedWith;
+use Lcobucci\JWT\Validation\RequiredConstraintsViolated;
+use PHPUnit\Framework\TestCase;
+
+use function assert;
+
+/**
+ * @covers \Lcobucci\JWT\Configuration
+ * @covers \Lcobucci\JWT\Encoding\JoseEncoder
+ * @covers \Lcobucci\JWT\Encoding\ChainedFormatter
+ * @covers \Lcobucci\JWT\Encoding\MicrosecondBasedDateConversion
+ * @covers \Lcobucci\JWT\Encoding\UnifyAudience
+ * @covers \Lcobucci\JWT\Token\Builder
+ * @covers \Lcobucci\JWT\Token\Parser
+ * @covers \Lcobucci\JWT\Token\Plain
+ * @covers \Lcobucci\JWT\Token\DataSet
+ * @covers \Lcobucci\JWT\Token\Signature
+ * @covers \Lcobucci\JWT\Signer\Key\LocalFileReference
+ * @covers \Lcobucci\JWT\Signer\Key\InMemory
+ * @covers \Lcobucci\JWT\Signer\Eddsa
+ * @covers \Lcobucci\JWT\Signer\InvalidKeyProvided
+ * @covers \Lcobucci\JWT\Signer\OpenSSL
+ * @covers \Lcobucci\JWT\Validation\Validator
+ * @covers \Lcobucci\JWT\Validation\Constraint\SignedWith
+ * @covers \Lcobucci\JWT\Validation\Validator
+ * @covers \Lcobucci\JWT\Validation\RequiredConstraintsViolated
+ * @covers \Lcobucci\JWT\Validation\Constraint\SignedWith
+ */
+class EddsaTokenTest extends TestCase
+{
+    use Keys;
+
+    private Configuration $config;
+
+    /** @before */
+    public function createConfiguration(): void
+    {
+        $this->config = Configuration::forAsymmetricSigner(
+            new Eddsa(),
+            static::$eddsaKeys['private'],
+            static::$eddsaKeys['public1']
+        );
+    }
+
+    /** @test */
+    public function builderShouldRaiseExceptionWhenKeyIsInvalid(): void
+    {
+        $builder = $this->config->builder();
+
+        $this->expectException(InvalidKeyProvided::class);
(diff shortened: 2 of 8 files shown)


src/Signer/Eddsa.php
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Lcobucci\JWT\Signer;
+
+use Lcobucci\JWT\Signer;
+
+use function extension_loaded;
+use function sodium_crypto_sign_detached;
+use function sodium_crypto_sign_verify_detached;
+
+class Eddsa implements Signer
```
lcobucci: Did you design this to be extended? It doesn't look like. If not, please make it `final`.

test/functional/EddsaTokenTest.php
```diff
@@ -0,0 +1,141 @@
...
+        );
+    }
+
+    /** @test */
+    public function builderShouldRaiseExceptionWhenKeyIsInvalid(): void
+    {
+        $builder = $this->config->builder();
+
+        $this->expectException(SodiumException::class);
+        $this->expectExceptionMessage('SODIUM_CRYPTO_SIGN_SECRETKEYBYTES');
```
lcobucci: We should convert this into something that implements `Lcobucci\JWT\Exception`. Perhaps `InvalidKeyProvided`?

test/unit/Signer/EddsaTest.php
```diff
@@ -0,0 +1,108 @@
...
+    public function verifyShouldRaiseAnExceptionWhenKeyIsNotParseable(): void
+    {
+        $signer = $this->getSigner();
+
+        $this->expectException(SodiumException::class);
+        $this->expectExceptionMessage('SODIUM_CRYPTO_SIGN_BYTES');
+
+        $signer->verify('testing', 'testing', InMemory::plainText('blablabla'));
+    }
+
```
lcobucci, line 100: Would you please add the examples we have on https://tools.ietf.org/html/rfc8037 here? Just to make sure we're fully compatible with the specs.

Something like (I didn't run this but it should work:tm:):

```php
/**
 * @test
 * @see https://tools.ietf.org/html/rfc8037#appendix-A.4
 */
public function signatureOfRfcExample(): void
{
    $signer  = $this->getSigner();
    $encoder = new \Lcobucci\JWT\Encoding\JoseEncoder();

    $key       = InMemory::plainText($encoder->base64UrlDecode('nWGxne_9WmC6hEr0kuwsxERJxWl7MmkZcDusAxyuf2A') . $encoder->base64UrlDecode('11qYAYKxCrfVS_7TyWQHOg7hcvPapiMlrwIaaPcHURo'));
    $payload   = $encoder->base64UrlEncode('{"alg":"EdDSA"}') . '.' . $encoder->base64UrlEncode('Example of Ed25519 signing');
    $signature = $signer->sign($payload, $key);

    self::assertSame('eyJhbGciOiJFZERTQSJ9.RXhhbXBsZSBvZiBFZDI1NTE5IHNpZ25pbmc', $payload);
    self::assertSame('hgyY0il_MGCjP0JzlnLWG1PPOt7-09PGcvMg3AIbQR6dWbhijcNR4ki4iylGjg5BhVsPt9g7sVvpAr_MuM0KAg', $encoder->base64UrlEncode($signature));
}

/**
 * @test
 * @see https://tools.ietf.org/html/rfc8037#appendix-A.5
 */
public function verificationOfRfcExample(): void
{
    $signer  = $this->getSigner();
    $encoder = new \Lcobucci\JWT\Encoding\JoseEncoder();

    $key       = InMemory::plainText($encoder->base64UrlDecode('11qYAYKxCrfVS_7TyWQHOg7hcvPapiMlrwIaaPcHURo'));
    $payload   = 'eyJhbGciOiJFZERTQSJ9.RXhhbXBsZSBvZiBFZDI1NTE5IHNpZ25pbmc';
    $signature = $encoder->base64UrlDecode('hgyY0il_MGCjP0JzlnLWG1PPOt7-09PGcvMg3AIbQR6dWbhijcNR4ki4iylGjg5BhVsPt9g7sVvpAr_MuM0KAg');

    self::assertTrue($signer->verify($signature, $payload, $key));
}
```

test/performance/EddsaBench.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace Lcobucci\JWT;
+
+use Lcobucci\JWT\Signer\Eddsa;
+use Lcobucci\JWT\Signer\Key;
+use Lcobucci\JWT\Signer\Key\InMemory;
+use PhpBench\Benchmark\Metadata\Annotations\Groups;
+
+/** @Groups({"None"}) */
```
lcobucci:
```suggestion
/** @Groups({"EdDSA"}) */
```

lcobucci (changes requested):
@Slamdunk this looks quite good, thanks :+1: 

I've got a few small requests, though :grimacing:

test/functional/EddsaTokenTest.php
Slamdunk: I've started with that approach, but realized `ext-sodium` is freaking well designed and doesn't need any help, no way you can misuse it, so specific exceptions don't make much sense to me.

If you prefer we can have something like:

```php
namespace Lcobucci\JWT {
    class SodiumException extends \SodiumException implements Exception {}
}

try {
    sodium_stuff();
} catch (\SodiumException $se) {
    throw new \Lcobucci\JWT\SodiumException($se->getMessage(), 0, $se);
}
```
But without specific exception for each failing case.

test/unit/Signer/EddsaTest.php
Slamdunk: Didn't know the RFC had real examples, awesome, I'm going to commit them!

Slamdunk: > I didn't run this but it should work

You forgot the `@covers` and `@uses` annotations... [cough cough] loser [cough cough]

test/functional/EddsaTokenTest.php
lcobucci: The problem is that Introducing a new exception type is an LSP violation. These are the exceptions expected by the clients: https://github.com/lcobucci/jwt/blob/4.1.x/src/Signer.php

Slamdunk: Mmm, that should have been pointed out by libraries like `phpstan/phpstan-strict-rules` so...

Going to push the fix

Slamdunk:
Ed25519 is awesome, among the asymmetric signing is the fastest, and its keys are so easy to store and generate:

```
+-------------+-----------+
| benchmark   | diff      |   
+-------------+-----------+
| Sha384Bench | 460.14x   |   
| Sha384Bench | 892.40x   |   
| Sha256Bench | 459.90x   |   
| Sha256Bench | 764.18x   |   
| Sha512Bench | 466.50x   |   
| Sha512Bench | 957.14x   |   
| EddsaBench  | 123.00x   |   
| EddsaBench  | 303.83x   |   
| NoneBench   | 1.00x     |   
| NoneBench   | 1.17x     |   
| Sha384Bench | 5,163.65x |
| Sha384Bench | 313.29x   |   
| Sha256Bench | 5,258.72x |
| Sha256Bench | 383.07x   |   
| Sha512Bench | 5,492.08x |
| Sha512Bench | 312.46x   |   
+-------------+-----------+
```

src/Signer/Eddsa.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        return 'EdDSA';
+    }
+
+    public function sign(string $payload, Key $key): string
+    {
+        try {
+            return sodium_crypto_sign_detached($payload, $key->contents());
+        } catch (SodiumException $sodiumException) {
+            throw new InvalidKeyProvided($sodiumException->getMessage(), 0, $sodiumException);
```
Slamdunk, line 24: I've used `InvalidKeyProvided` exception but not its factory methods because `SodiumException` has much wider scope range

lcobucci: We could introduce a named constructor there but this works too.

lcobucci:
@Slamdunk I'm just wondering about the optional dependency on sodium... Do you have anything against just requiring it? 

We added mbstring requirement in the past (which is only there for ECDSA), so the precedent is there. 

Apart from that, sodium is pretty stable and works really well - I have it enabled in most of the projects I work.

That would allow for us to remove the checks and move the "breakage" to build time instead of runtime. 

I had the idea in the past to break the signers into satellite repositories but that creates more complexity for very little benefit. 

What's your take?

Slamdunk:
> I had the idea in the past to break the signers into satellite repositories but that creates more complexity for very little benefit.

Sub-packages are a good idea for non-core features that require external libraries, you don't want to be slowed down in the mainline by a dependency.
However, extensions are always ready when a new PHP version is ready, so you're not slowed down by them, so features that require them can make into main-package.

> I'm just wondering about the optional dependency on sodium... Do you have anything against just requiring it?

No: often maintainers prefer PR that do not force anything to users, so I went with a runtime check, since we haven't had a prior discussion.

Here comes the updated push

lcobucci (approved):

lcobucci:
@Slamdunk I just rebase to add the GPG signature on commits and tidy the history a bit. Will merge once CI is done. 

Thanks a lot!

PR #606, opened 2020-12-10
php 8 fixes

kenanduman1988:


lcobucci (changes requested):
Can you please explain what this is fixing and why the build on PHP 8 doesn't fail?

lcobucci:
Closing as `invalid` since the changes don't make sense to me, the existing code runs well on PHP 8, and I received no explanation on about why this is being done.

PR #613, opened 2020-12-17
Prefer ext-sodium cache-timing-safe functions

Slamdunk:
Since from #605 `ext-sodium` is required, we can now leverage it

### DIFF
--- a/src/Encoding/JoseEncoder.php
+++ b/src/Encoding/JoseEncoder.php
@@ -55,7 +55,7 @@ final class JoseEncoder implements Encoder, Decoder
         try {
             return sodium_base642bin($data, SODIUM_BASE64_VARIANT_URLSAFE_NO_PADDING, '');
         } catch (SodiumException $sodiumException) {
-            throw CannotDecodeContent::invalidBase64String();
+            throw CannotDecodeContent::invalidBase64String($sodiumException);
         }
     }
 }
--- a/src/Encoding/JoseEncoder.php
+++ b/src/Encoding/JoseEncoder.php
@@ -6,18 +6,17 @@ namespace Lcobucci\JWT\Encoding;
 use JsonException;
 use Lcobucci\JWT\Decoder;
 use Lcobucci\JWT\Encoder;
+use SodiumException;
 
-use function base64_decode;
-use function base64_encode;
-use function is_string;
 use function json_decode;
 use function json_encode;
-use function rtrim;
-use function strtr;
+use function sodium_base642bin;
+use function sodium_bin2base64;
 
 use const JSON_THROW_ON_ERROR;
 use const JSON_UNESCAPED_SLASHES;
 use const JSON_UNESCAPED_UNICODE;
+use const SODIUM_BASE64_VARIANT_URLSAFE_NO_PADDING;
 
 /**
  * A utilitarian class that encodes and decodes data according with JOSE specifications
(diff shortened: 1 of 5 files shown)


test/performance/Hmac/HmacBench.php
```diff
@@ -11,7 +11,7 @@
 /** @Groups({"Hmac"}) */
 abstract class HmacBench extends SignerBench
 {
-    private const ENCODED_KEY = 'hJtXIZ2uSN5kbQfbtTNWbpdmhkV8FJG+Onbc6mxCcYg';
+    private const ENCODED_KEY = 'hJtXIZ2uSN5kbQfbtTNWbpdmhkV8FJG+Onbc6mxCcYg=';
```
Slamdunk, line 14: Previously `\Lcobucci\JWT\Signer\Key\InMemory::base64Encoded` accepted both paddad and non-padded base64 strings due to `base64_decode` looseness.

For the purpose of `InMemory` class, I chose `SODIUM_BASE64_VARIANT_ORIGINAL` and I chose not to accept (alias fallback) non-padded keys.

I consider this sort of a bug, so no BC break for me here.

test/unit/Encoding/JoseEncoderTest.php
```diff
@@ -148,7 +148,7 @@ public function base64UrlDecodeMustRaiseExceptionWhenInvalidBase64CharsAreUsed()
         $this->expectExceptionCode(0);
         $this->expectExceptionMessage('Error while decoding from Base64Url, invalid base64 characters detected');
 
-        $decoder->base64UrlDecode('áááááá');
+        $decoder->base64UrlDecode('ááá');
```
Slamdunk, line 151: So, `áááááá` is a valid `SODIUM_BASE64_VARIANT_URLSAFE_NO_PADDING` string, where `ááá` is not.

Currently no idea why, but I trust `ext-sodium` much more than `base64_decode`.

Slamdunk (commented):
Also `\Lcobucci\JWT\Signer\Ecdsa\MultibyteStringConverter` could benefit from sodium functions, but I don't have the required knowledge to do the right things in the right way.

src/Encoding/JoseEncoder.php
```diff
@@ -48,18 +47,15 @@ public function jsonDecode(string $json)
...
     public function base64UrlDecode(string $data): string
     {
-        // Padding isn't added back because it isn't strictly necessary for decoding with PHP
-        $decodedContent = base64_decode(strtr($data, '-_', '+/'), true);
-
-        if (! is_string($decodedContent)) {
+        try {
+            return sodium_base642bin($data, SODIUM_BASE64_VARIANT_URLSAFE_NO_PADDING, '');
+        } catch (SodiumException $sodiumException) {
             throw CannotDecodeContent::invalidBase64String();
```
lcobucci: We should forward the previous exception then, that would be of great help for folks.

src/Signer/Key/InMemory.php
```diff
@@ -35,9 +38,9 @@ public static function plainText(string $contents, string $passphrase = ''): sel
 
     public static function base64Encoded(string $contents, string $passphrase = ''): self
     {
-        $decoded = base64_decode($contents, true);
-
-        if ($decoded === false) {
+        try {
+            $decoded = sodium_base642bin($contents, SODIUM_BASE64_VARIANT_ORIGINAL, '');
+        } catch (SodiumException $sodiumException) {
             throw CannotDecodeContent::invalidBase64String();
```
lcobucci: Same here, forward the previous exception.

lcobucci (changes requested):
@Slamdunk that's quite great! Thanks 👍 

Just a teeny tiny thing that can be improved.

src/Encoding/JoseEncoder.php
Slamdunk: I thought about it: is it ok for you to accept for `CannotDecodeContent::invalidBase64String` an optional exception argument, so we can keep the API while forwarding the sodium exception?

lcobucci: Sure thing

Slamdunk: Turned out `CannotDecodeContent::invalidBase64String` is only used with sodium now, so I've enforced the API with the sodium exception

lcobucci (approved):

src/Encoding/JoseEncoder.php
lcobucci: Yeap, it's a minor BC-break but it's quite acceptable. Folks will only catch the exception.

Slamdunk:
Mmm, shouldn't `\Lcobucci\JWT\Encoding\CannotDecodeContent::invalidBase64String` be internal? I mean, the factory method, not the class as a whole

lcobucci:
@Slamdunk we could mark it as internal. I just didn't bother to do that before.

Slamdunk:
That's sound careless for a [perfectionist](https://speakerdeck.com/lcobucci/living-with-this-perfectionist-mind-v1) :stuck_out_tongue:

Issue #611, opened 2020-12-13
Should the TokenInterface contain a method for claims() ?


Tjeerd:
After upgrading to v4 of the library my static code analysis is failing because claims() is not defined Token interface. I guess this method should be declared, also because the code of the library itself is depending on it but others as well. Do you agree? If so I will create a PR for this change.

nanawel:
I was about to ask the same question :)

t0mmy742:
It was already discussed on #186 and #203 

> Since `parse()` deals with untrusted data, users are supposed to verify if they have the expected type of token, like:
> 
> ```
> $token = $this->parser->parse($jwt);
> 
> if (! $token instanceof Plain) {
>     // You should have your own exception for that, sure =)
>     throw new \RuntimeException('The provided token is not valid');
> }
> ```

lcobucci:
Thanks @t0mmy742, that's indeed the explanation.

What I'm thinking about (for v4.1) is to have an interface for the unencrypted token, so that users can do the check using it instead.

That was suggested in https://github.com/lcobucci/jwt/pull/186#discussion_r138533421 and I've essentially changed my mind (especially because that allows for better extensibility).

Tjeerd:
Right, sorry to have missed those other issues. Closing this one as I understand and agree with the answers given there.

PR #615, opened 2020-12-20
Add missing claims method definition to Token interface

sasa-b:
I've noticed that the Token interface is missing the definition of the claims method.


lcobucci:
That has been discussed in many issues, the last was https://github.com/lcobucci/jwt/issues/611.

Would you like to modify this PR according to the discussion?

lcobucci:
Replaced by #616

PR #616, opened 2020-12-23
Expose interface that represents unencrypted tokens

### LINKED ISSUE #203: How get Claims after encode token

### LINKED ISSUE #228: Getting claims from Token

### LINKED ISSUE #611: Should the TokenInterface contain a method for claims() ?

Slamdunk:
Resolve https://github.com/lcobucci/jwt/pull/186
Resolve https://github.com/lcobucci/jwt/issues/228
Resolve https://github.com/lcobucci/jwt/issues/203
Resolve https://github.com/lcobucci/jwt/pull/552
Resolve https://github.com/lcobucci/jwt/issues/611
Resolve https://github.com/lcobucci/jwt/pull/615

### DIFF
--- a/docs/extending-the-library.md
+++ b/docs/extending-the-library.md
@@ -172,6 +172,7 @@ To create your own implementation of constraint you must implement the `Lcobucci
 
 ```php
 use Lcobucci\JWT\Token;
+use Lcobucci\JWT\UnencryptedToken;
 use Lcobucci\JWT\Validation\Constraint;
 use Lcobucci\JWT\Validation\ConstraintViolation;
 
@@ -179,7 +180,7 @@ final class SubjectMustBeAValidUser implements Constraint
 {
     public function assert(Token $token): void
     {
-        if (! $token instanceof Token\Plain) {
+        if (! $token instanceof UnencryptedToken) {
             throw new ConstraintViolation('You should pass a plain token');
         }
 
--- a/docs/parsing-tokens.md
+++ b/docs/parsing-tokens.md
@@ -8,7 +8,7 @@ To parse a token you must create a new parser (easier when using the [configurat
 
 ```php
 use Lcobucci\JWT\Configuration;
-use Lcobucci\JWT\Token\Plain;
+use Lcobucci\JWT\UnencryptedToken;
 
 $config = $container->get(Configuration::class);
 assert($config instanceof Configuration);
@@ -19,7 +19,7 @@ $token = $config->parser()->parse(
     . '2gSBz9EOsQRN9I-3iSxJoFt7NtgV6Rm0IL6a8CAwl3Q'
 );
 
-assert($token instanceof Plain);
+assert($token instanceof UnencryptedToken);
 
 $token->headers(); // Retrieves the token headers
 $token->claims(); // Retrieves the token claims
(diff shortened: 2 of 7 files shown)


src/PlainToken.php
```diff
@@ -0,0 +1,14 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT;
+
+use Lcobucci\JWT\Token\DataSet;
+
+interface PlainToken extends Token
```
lcobucci: I'd call this `UnencryptedToken` in this context.

lcobucci (changes requested):
We also need to change the SignedWith constraint: https://github.com/lcobucci/jwt/blob/4.1.x/src/Validation/Constraint/SignedWith.php#L24

(and possibly need to add the payload method to the this interface as well).

Slamdunk:
> (and possibly need to add the payload method to the this interface as well).

I didn't change `SignedWith` because indeed it also requires `payload` and `signature`, two methods only useful within the `SignedWith` context.
Just one last confirmation: do you want me to add the two methods mentioned in the new `UnencryptedToken` interface?

lcobucci:
> > (and possibly need to add the payload method to the this interface as well).
> 
> I didn't change `SignedWith` because indeed it also requires `payload` and `signature`, two methods only useful within the `SignedWith` context.
> Just one last confirmation: do you want me to add the two methods mentioned in the new `UnencryptedToken` interface?

Yes, please. These are relevant to unencrypted tokens.

The main benefit we gain here is to allow users to provider their own builders/parsers/tokens implementation and everything will just work.

lcobucci (approved):
The only thing we're missing to allow people to make full use of this is to change the builder to use the interface as well... That's a minor BC-break, though. I'll think about it for the next few days. Let's get this merged, though.

@Slamdunk thanks!

lcobucci:
@Slamdunk would please sign your commits? Otherwise I will rebase your branch on the weekend.

Slamdunk:
The users don't **need** `\Lcobucci\JWT\Builder::getToken` to get changed, even though of course it would be good in the next major. It already return a `Plain` token.

Users will still need `assert($token instanceof UnencryptedToken);` because the place where they need this is after `\Lcobucci\JWT\Parser::parse` call, which should always return a `\Lcobucci\JWT\Token` in case encrypted token will be implemented, afaict.

> would please sign your commits?

Yeah, sorry, It's weird to me that commits appear unsigned just because the key is expired. I'm waiting my Yubikeys, in the meantime I'm using short-lived keys.

I'll re-sign soon

lcobucci:
@Slamdunk thanks!

PR #617, opened 2020-12-23
Provide the UnixTimestampDates formatter

Slamdunk:


src/Encoding/ChainedFormatter.php
```diff
@@ -20,6 +20,11 @@ public static function default(): self
         return new self(new UnifyAudience(), new MicrosecondBasedDateConversion());
     }
 
+    public static function defaultWithUnixTimestampDates(): self
```
lcobucci: Perhaps we could drop the "default" prefix? Using it wasn't my brightest moment (sleep deprivation and all, you know).

lcobucci:
@Slamdunk thanks!

PR #619, opened 2020-12-30
Improve CI configuration

t0mmy742:
Multiple small things here:

- Change the way we are using `action/cache` from GitHub Actions. We were caching `vendor` directory, which is not really a good practice as mentioned [here](https://github.com/shivammathur/setup-php#cache-composer-dependencies). Moreover, cache is now detecting changes of `composer.lock`, since it used an hash of this file for generating the key.
- I reorganized GitHub Actions workflows. I added PHPUnit tests to PHP 8, in addition of PHP 7.4.
- PHPBench updated to 1.0.0-alpha4. Previous versions does not support PHP 8. It does not seem to change anything for us. We should be ready when a stable version get released.
- vfsstream minimal version is now 1.6.7. Released last year, this version corrected a deprecated feature in PHP 7.4, removed in PHP 8.0 ([see here](https://github.com/bovigo/vfsStream/pull/189)). Last version is 1.6.8.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -31,14 +31,16 @@ jobs:
           ini-values: memory_limit=-1
           tools: composer:v2, cs2pr
 
+      - name: Get composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
       - name: "Cache dependencies"
         uses: "actions/cache@v2"
         with:
-          path: |
-            ~/.composer/cache
-            vendor
-          key: "php-${{ matrix.php-version }}-${{ matrix.dependencies }}"
-          restore-keys: "php-${{ matrix.php-version }}-${{ matrix.dependencies }}"
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: "php-${{ matrix.php-version }}-composer-${{ matrix.dependencies }}-${{ hashFiles('**/composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-${{ matrix.dependencies }}-"
 
       - name: "Install lowest dependencies"
         if: ${{ matrix.dependencies == 'lowest' }}
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -16,6 +16,7 @@ jobs:
           - "locked"
         php-version:
           - "7.4"
+          - "8.0"
         operating-system:
           - "ubuntu-latest"
 
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -19,6 +19,7 @@ jobs:
           - "development"
         php-version:
           - "7.4"
+          - "8.0"
         operating-system:
           - "ubuntu-latest"
 
@@ -29,6 +30,7 @@ jobs:
       - name: "Install PHP"
         uses: "shivammathur/setup-php@v2"
         with:
+          coverage: "none"
           php-version: "${{ matrix.php-version }}"
           ini-values: memory_limit=-1
           tools: composer:v2, cs2pr
@@ -64,7 +66,7 @@ jobs:
         run: "make phpunit"
 
   phpunit-rc:
-    name: "PHPUnit tests on PHP 8"
+    name: "PHPUnit tests (nightly)"
 
     runs-on: ${{ matrix.operating-system }}
 
@@ -73,7 +75,7 @@ jobs:
         dependencies:
           - "locked"
         php-version:
-          - "8.0"
+          - "8.1"
         operating-system:
           - "ubuntu-latest"
 
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -33,14 +33,16 @@ jobs:
           ini-values: memory_limit=-1
           tools: composer:v2, cs2pr
 
+      - name: Get composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
       - name: "Cache dependencies"
         uses: "actions/cache@v2"
         with:
-          path: |
-            ~/.composer/cache
-            vendor
-          key: "php-${{ matrix.php-version }}-${{ matrix.dependencies }}"
-          restore-keys: "php-${{ matrix.php-version }}-${{ matrix.dependencies }}"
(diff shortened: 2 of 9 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -8,16 +8,12 @@ jobs:
...
+    runs-on: "ubuntu-latest"
 
     strategy:
       matrix:
-        dependencies:
-          - "locked"
         php-version:
           - "7.4"
-        operating-system:
-          - "ubuntu-latest"
```
Slamdunk, line 20: I think this single-variable style was chosen to ease the management between `*.yml` action files.
So it would be better to revert these (set of) changes.

Slamdunk (changes requested):
PR looks ok, just one revert IMHO :+1:

.github/workflows/benchmarks.yml
t0mmy742: I surely miss something, but I don't see how it can ease some kind of management since all action files are using the same `runs-on`.

Slamdunk: Considering the OS as a variable helps adding perhaps Windows in the future, as well as keep in sync with template repositories. I can't tell you if @lcobucci is doing this, but I do and it's easier to just diff the `strategy.matrix` section and leave everything else copy-pasted across repositories.

t0mmy742: At the beginning, I remove this variable since it was always the same and used only for `runs-on`. But your story about templates and keeping things in sync convinces me. I'm going to push modifications.
Thanks for the advice 👍

Ocramius (approved):
LGTM 👍

Ocramius:
Ah, I think the required checks need to be re-configured for this to be merged 😬

.github/workflows/benchmarks.yml
```diff
@@ -12,8 +12,6 @@ jobs:
 
     strategy:
       matrix:
-        dependencies:
-          - "locked"
```
Slamdunk, line 16: The same reasoning for `os` should go for `dependencies` too: the goal is to copy-paste https://github.com/lcobucci/jwt/pull/619/files#diff-ea12f60188cdd90bc99e5d0af2eb91647bbe4a9199176aa1ec5240f65efed510R45 everywhere

Moreover, explicit matrix is easier to debug than implicit ones

Slamdunk (changes requested):

.github/workflows/benchmarks.yml
Ocramius: Note: most of the matrix is copied from my repos, where I really like this explicitness too 👍

t0mmy742: Well, if you think it's better, let's do it 😊 
Should we keep unused steps (which depends on `dependencies` like [this](https://github.com/lcobucci/jwt/pull/619/files#diff-dfd15ae70247bb45f121522ec2e50e5da9aed17e23116d94ec887a4769c44294L43)) ?

Slamdunk: IMHO, yes

Slamdunk (approved):

Ocramius:
These pipelines no longer exist:

 > * PHPUnit tests on PHP 8 (locked, 8.0, ubuntu-latest) 
 > * Run benchmarks (locked, 7.4, ubuntu-latest) 

These changes are OK, but require @lcobucci to change repo settings before merge

lcobucci:
Thanks folks! I'm taking a break from github and my laptop for a couple of days. Will take a look at it around the weekend.

.github/workflows/phpunit.yml
```diff
@@ -60,45 +61,3 @@ jobs:
...
-    name: "PHPUnit tests on PHP 8"
-
-    runs-on: ${{ matrix.operating-system }}
-
-    strategy:
-      matrix:
-        dependencies:
-          - "locked"
-        php-version:
-          - "8.0"
```
lcobucci, line 74: We should use 8.1 here to keep an eye on dev version.

.github/workflows/benchmarks.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         dependencies:
           - "locked"
         php-version:
-          - "7.4"
```
lcobucci, line 18: Does v1 of phpbench work on PHP 7.4?

composer.json
```diff
@@ -28,8 +28,8 @@
     "require-dev": {
         "infection/infection": "^0.20",
         "lcobucci/coding-standard": "^6.0",
-        "mikey179/vfsstream": "^1.6",
-        "phpbench/phpbench": "^0.17",
+        "mikey179/vfsstream": "^1.6.7",
+        "phpbench/phpbench": "^1.0.0-alpha4",
```
lcobucci: Let's use `^1.0@alpha` here. Then we'll get the newer releases too.

lcobucci (changes requested):
Just some tiny things, thanks for the help @t0mmy742!

.github/workflows/benchmarks.yml
t0mmy742: Yes, it supports PHP ^7.2 || ^8.0.
Should we keep benchmarks on PHP 7.4 ?
I started to update to PHP 8.0 only this one because we should have problems with static analysis while we keep PHP 7.4 support.

lcobucci:
@t0mmy742 did you revert the cache optimisation on purpose? I thought the feedback was mostly about the templating.

t0mmy742:
@lcobucci I missed something with `git rebase` :cry:
I'm going to correct this

lcobucci:
> @lcobucci I missed something with `git rebase` 
> I'm going to correct this

It's alright, happens to everyone :laughing: That's why the reflog is there for

.github/workflows/benchmarks.yml
lcobucci: @t0mmy742 what problems do you mean? I think the only one is the object/resource isn't it? That's being ignored atm.

t0mmy742: Yes, it's about this

lcobucci: So we can have both, shouldn't cause any issue on benchmarks. That's mostly out of curiosity and I'm planning to use their comparison tooling (that comes with v1.0).

I'll rebase your PR and get it merged!

lcobucci (approved):

lcobucci:
@t0mmy742 thanks a lot!

Issue #629, opened 2021-01-13
JWT 4.0.0: Unable to set an array as "aud" claim


SvenRtbg:
I am in the process of migrating from 3.3 via 3.4 (used right now) to 4.0.

I stumbled upon this situation: I have test cases for an OpenID-Connect library that set an array as audience claim. If I call `withClaim('aud', ['one', 'two']` I get a deprecation message. So I change the call to `permittedFor(['one','two'])` and get "Array to string conversion".

Based on the JWT RFC https://tools.ietf.org/html/rfc7519#section-4.1.3 I would say that arrays with multiple aud claim entries are allowed and possible, and I would also think they are quite common, because these test cases are there because there are some very annoying validation rules for OIDC going on depending on the "aud" claim being an array with multiple entries and an "azp" claim being present. But this is only the background info.

Bottom line: I would very much like to be able to set an "aud" claim with an array entry. Currently I cannot. Calling `permittedFor()` multiple times will overwrite the previous values.

I have also seen that the rendering of the token makes some attempts to normalize the representation of the "aud" claim. I can understand why one would do this, but unless every software consuming JWT is normalizing the difference between "single entry array" or "string", I fear this may affect situations where it is important to send either way.

SvenRtbg:
My test method for this scenario:

    public function testSettingAudClaimArray(): void
    {
        $builder = new Builder();
        $builder->permittedFor('someone');
        $builder->permittedFor('someone2');
        $token = $builder->getToken(new None(), InMemory::plainText(''));
        $json = $token->toString();

        $parser = new Parser();
        $newToken = $parser->parse($json);
        $audience = $newToken->claims()->get('aud');
        $this->assertIsArray($audience);
        $this->assertCount(2, $audience);
    }

lcobucci:
@SvenRtbg as you can see on the interface, the method is variadic https://github.com/lcobucci/jwt/blob/4.0.0/src/Builder.php#L20, so you can pass multiple audiences at once.

Alternatively, you can call the method more than once, as shown in your sample test scenario and covered by this test on our suite: https://github.com/lcobucci/jwt/blob/4.0.0/test/unit/Token/BuilderTest.php#L95-L118

With that said, have you tried these approaches? Is the library misbehaving in this regard or is it a documentation matter?

SvenRtbg:
[...]

PR #630, opened 2021-01-14
Improve deprecation for aud claims

SvenRtbg:
Addresses some concerns in #629.

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -10,6 +10,33 @@ We're building on the version `v3.4.0` a forward compatibility layer to help use
 To help on the migration process, all deprecated components are being marked with `@deprecated` and deprecated behaviour will trigger a `E_USER_DEPRECATED` error.
 However, you can also find here the instructions on how to make your code compatible with both versions.
 
+### General migration strategy
+
+Update your existing software to the latest 3.4.x version using composer:
+
+```sh
+composer require lcobucci/jwt ^3.4
+```
+
+Then run your tests and fix all deprecation notices.
+Also change all calls to deprecated methods, even if they are not triggering any notices.
+Tools like [`phpstan/phpstan-deprecation-rules`](https://github.com/phpstan/phpstan-deprecation-rules) can help finding them.
+
+Note that PHPUnit tests will only fail if you have the `E_USER_DEPRECATED` error level activated - it is a good practice to run tests using `E_ALL`.
+Data providers that trigger deprecation messages will not fail tests at all, only print the message to the console.
+Make sure you do not see any of them before you continue.
+
+Now you can upgrade to the latest 4.x version:
+
+```sh
+composer require lcobucci/jwt ^4.0
+```
+
+Remember that some deprecation messages from the 3.4 version may have notified you that things still are different in 4.0, so you may find you need to adapt your own code once more at this stage.
+
+While you are at it, consider using the new configuration object.
+The details are listed below.
+
 ### Inject the configuration object instead of builder/parser/key
 
 This object serves as a small service locator that centralises the JWT-related dependencies.
@@ -153,10 +180,13 @@ Here's the migration:
 
 #### Support for multiple audiences
 
-Even though we didn't officially support multiple audiences, it was technically possible to achieve that by manually setting the `aud` claim.
+Even though we didn't officially support multiple audiences, it was technically possible to achieve that by manually setting the `aud` claim to an array with multiple strings.
 
-For this case, there are different approaches to migrate your code.
-Feel free to choose your preferred one.
+If you parse a token with 3.4, and read it's contents with `\Lcobucci\JWT\Token#getClaim()` or`\Lcobucci\JWT\Token#getClaims()`, you will only get the first element of such an array back.
+If the audience claim does only contain a string, or only contains one string in the array, nothing changes.
+Please [upgrade to the new Token API](#use-the-new-token-api) for accessing claims in order to get the full audience array again (e.g. call `Token#claims()->get('aud')`).
+
+When creating a token, use the new method `Builder#permittedFor()` as detailed below.
 
 ##### Multiple calls to `Builder#permittedFor()`
 
--- a/src/Token.php
+++ b/src/Token.php
@@ -232,6 +232,9 @@ class Token
         }
 
         if ($name === RegisteredClaims::AUDIENCE && is_array($value)) {
+            if (count($value) > 1) {
+                trigger_error('You will only get the first array entry as a string. Use Token::claims()->get() instead.', E_USER_DEPRECATED);
+            }
             return current($value);
         }
 
--- a/test/unit/TokenTest.php
+++ b/test/unit/TokenTest.php
@@ -284,6 +284,25 @@ class TokenTest extends \PHPUnit\Framework\TestCase
      *
      * @uses \Lcobucci\JWT\Token::__construct
      * @uses \Lcobucci\JWT\Token::convertToDataSet
+     * @uses \Lcobucci\JWT\Token::hasClaim
+     * @uses Lcobucci\JWT\Claim\Basic
+     *
+     * @covers ::getClaim
+     */
+    public function getClaimShouldEmitDeprecationMessageWhenAccessingAudClaimArray()
+    {
+        $this->expectDeprecation('You will only get the first array entry as a string. Use Token::claims()->get() instead.');
+        $token = new Token([], ['aud' => ['test1', 'test2']]);
+
+        $this->assertEquals('test1', $token->getClaim('aud'));
+        $this->assertEquals(['test1', 'test2'], $token->claims()->get('aud'));
+    }
+
+    /**
+     * @test
+     *
+     * @uses \Lcobucci\JWT\Token::__construct
+     * @uses \Lcobucci\JWT\Token::convertToDataSet
      *
      * @covers ::verify
      * @covers ::getPayload


SvenRtbg:
Seems like I have to enable signatures for my commits... will do ASAP, but it will take a while. Please review and give feedback.

src/Token.php
```diff
@@ -232,6 +232,9 @@ public function getClaim($name, $default = null)
         }
 
         if ($name === RegisteredClaims::AUDIENCE && is_array($value)) {
+            if (count($value) > 1) {
```
Ocramius, line 235: Should have a test that expects a deprecation

src/Builder.php
```diff
@@ -108,6 +108,10 @@ public function canOnlyBeUsedBy($audience, $replicateAsHeader = false)
      */
     public function permittedFor($audience, $replicateAsHeader = false)
     {
+        if (is_array($audience)) {
+            trigger_error('Using arrays for audience claims is not really supported in 3.4 and will behave different in 4.0.', E_USER_DEPRECATED);
```
Ocramius: Should have a test that expects a deprecation

Ocramius (changes requested):
Note: I don't think you actually need to add GPG signatures to commits.

test/unit/BuilderTest.php
```diff
@@ -110,6 +110,31 @@ public function permittedForCanReplicateItemOnHeader()
...
+    {
+        $this->expectDeprecation('Using arrays for audience claims is not really supported in 3.4 and will behave different in 4.0.');
+
+        $builder = $this->createBuilder();
+        $builder->permittedFor(['test1', 'test2']);
+
+        $token = $builder->getToken(new None(), Key\InMemory::plainText(''));
+
+        self::assertEquals(['typ' => 'JWT', 'alg' => 'none'], $token->getHeaders());
+        self::assertEquals(['aud' => new EqualsTo('aud', 'Array')], $token->getClaims());
```
Slamdunk: If I recall correctly, as soon as the deprecation occurs, the test halts, so later code can be deleted.
Moreover I suggest to expect the deprecation right before the call that raises it, so previous code cannot fall into a false positive case:
```suggestion
        $builder = $this->createBuilder();

        $this->expectDeprecation('Using arrays for audience claims is not really supported in 3.4 and will behave different in 4.0.');
        $builder->permittedFor(['test1', 'test2']);
```

test/unit/TokenTest.php
```diff
@@ -578,6 +578,25 @@ public function isPermittedForShouldReturnTrueWhenAudienceMatches()
...
+     */
+    public function isPermittedForWillTriggerDeprecationWhenAudienceIsArray()
+    {
+        $this->expectDeprecation('Behaviour for audience claims has changed in 3.4, you will only get the first array entry as a string. In 4.0 this will change again with more flexibility, please consult the documentation.');
+        $token = new Token(
+            [],
+            [RegisteredClaims::AUDIENCE => ['testing1', 'testing2']]
+        );
+
+        self::assertSame('testing1', $token->getClaim('aud'));
```
Slamdunk: Same here:
```suggestion
        $token = new Token(
            [],
            [RegisteredClaims::AUDIENCE => ['testing1', 'testing2']]
        );
        
        $this->expectDeprecation('Behaviour for audience claims has changed in 3.4, you will only get the first array entry as a string. In 4.0 this will change again with more flexibility, please consult the documentation.');

        $token->getClaim('aud');
```

Slamdunk (dismissed):

test/unit/BuilderTest.php
SvenRtbg: Basically I copied an existing test case right from above, which has the exact same structure that you see in the new test case.

In fact the test execution continues, so tests further down the code are still being used. This is not the same as expecting exceptions, the deprecation code will actually collect more than one and compare if they are all correctly expected.

One could argue that testing the headers has nothing to do with the test case, but this will only remove one line.

test/unit/TokenTest.php
SvenRtbg: See above

test/unit/BuilderTest.php
Slamdunk: Ok then

Ocramius (approved):

SvenRtbg:
Do we still have an issue here because of the commit signature requirement?

SvenRtbg:
I force-pushed again with a GPG key attached.

src/Builder.php
```diff
@@ -108,6 +108,10 @@ public function canOnlyBeUsedBy($audience, $replicateAsHeader = false)
...
+        }
```
lcobucci: Even though possible, the doc block states that the method expects a `string`.

Deviating from that is using the library different from what it was designed for. So, the user (in this case) is shooting at their own feet.

The limitation caused by ancient PHP version support is something I really prefer not to handle.

src/Token.php
```diff
@@ -232,6 +232,9 @@ public function getClaim($name, $default = null)
...
+                trigger_error('Behaviour for audience claims has changed in 3.4, you will only get the first array entry as a string. In 4.0 this will change again with more flexibility, please consult the documentation.', E_USER_DEPRECATED);
```
lcobucci: The behaviour didn't change. The library was never designed to have an array as value for the audience - you can see that on the doc block of all methods (even on the old interface), this `if` just made that explicit.

Using `withClaim()` or `setClaim()` to configure `aud` worked back then because things were overlooked and I've assumed people would use the specific methods to define registered claims.

Given we already have a deprecation message to prevent that, I'm not convinced we need yet another deprecation message here.

docs/upgrading.md
```diff
@@ -239,3 +262,26 @@ Your code should be adapted to manipulate tokens like this:
...
+token. Reading from the token via `\Lcobucci\JWT\Token::getClaim()` or`\Lcobucci\JWT\Token::getClaims()` will revert 
+this array wrapping and only ever return the first entry in the "aud" claims array, if it is one, or return the string 
+otherwise.
+
+If you pass an array to the builder or if you read an array from the token, a deprecation message will be
+triggered to inform you of this behaviour change. There is no workaround for version 3.4 if you have to
+access or create "aud" claims that are arrays with multiple entries.
+
+In 4.0, the `permittedFor()` method accepts a variadic parameter, so you can either call 
+`permittedFor("audience1", "audience2")` or `permittedFor(...["audience1", "audience2"])`
```
lcobucci: This assumes that the library had official support for multiple audiences on v3, which is not the case.

I'm happy to improve the docs to mention that things accidentally used to work because of lack of type declaration and loose validation on custom claim configuration, though.

docs/upgrading.md
```diff
@@ -10,6 +10,29 @@ We're building on the version `v3.4.0` a forward compatibility layer to help use
...
+
+Now you can upgrade to the latest 4.x version:
+
+    composer require lcobucci/jwt ^4.0
+
+At this stage, you'll probably find some more rough edges. Not every method call in the 3.4 version is
+100% identical to the 4.0 counterpart. So feel free to optimize your code at this point to leverage the
+core improvements like the configuration object, validation being separated from the `Token` value object and such. 
+The details are listed below.
+
```
lcobucci: We (try) to use [one sentence per line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line) on the docs to make things a bit easier to maintain (especially regarding diffs in the future).

docs/upgrading.md
```diff
@@ -10,6 +10,29 @@ We're building on the version `v3.4.0` a forward compatibility layer to help use
 To help on the migration process, all deprecated components are being marked with `@deprecated` and deprecated behaviour will trigger a `E_USER_DEPRECATED` error.
 However, you can also find here the instructions on how to make your code compatible with both versions.
 
+### General migration strategy
+
+It is recommended to update your existing software from any previous 3.x version to the latest 3.4.x version using 
+composer:
+
+    composer require lcobucci/jwt ^3.4
```
lcobucci: Please use the triple backslashes with `sh` as syntax highlight here.

docs/upgrading.md
```diff
@@ -10,6 +10,29 @@ We're building on the version `v3.4.0` a forward compatibility layer to help use
...
+you do not see any of them before you continue.
+
+Now you can upgrade to the latest 4.x version:
+
+    composer require lcobucci/jwt ^4.0
+
+At this stage, you'll probably find some more rough edges. Not every method call in the 3.4 version is
+100% identical to the 4.0 counterpart. So feel free to optimize your code at this point to leverage the
+core improvements like the configuration object, validation being separated from the `Token` value object and such. 
+The details are listed below.
```
lcobucci: Is there any rough edge that you've found apart from the cases documented in this file? If not, why this paragraph is relevant?

The idea of this doc section is to be a step-by-step kind of thing, so people would go through the things below and modify their code. I'm confused with the need for summarising the changes.

docs/upgrading.md
```diff
@@ -10,6 +10,29 @@ We're building on the version `v3.4.0` a forward compatibility layer to help use
 To help on the migration process, all deprecated components are being marked with `@deprecated` and deprecated behaviour will trigger a `E_USER_DEPRECATED` error.
 However, you can also find here the instructions on how to make your code compatible with both versions.
 
+### General migration strategy
+
+It is recommended to update your existing software from any previous 3.x version to the latest 3.4.x version using
```
lcobucci: Isn't this implied on the mention of the compatibility layer above? Perhaps we don't need a sub-section for the "General migration strategy" but add a paragraph above encouraging people to go through v3.4?

The remarks about PHPUnit are important but maybe using a note/warning might get more attention?

lcobucci (changes requested):
@SvenRtbg thanks for sending your suggestions.

I'm hesitant to add the deprecation messages because v3 wasn't designed to use arrays on the audience claim. This worked by "accident" for your application (because of several issues on the lib).

Apart from that argument, adding deprecation messages in a patch release is a no-go for me and v3 series is in bug-fix mode only.

SvenRtbg:
I can relate, however I think the 3.4 version is aimed at supporting the transition to 4.0.

Before 3.4, I could easily use `withClaim('aud',[...])` without complaints. Deprecation messages from that call lead to using `permittedFor()`, and this doesn't work when using an array as the parameter, which has been fine before.

And I cannot really skip migrating this call in 3.4 because in 4.0 the previous method is deleted. And I also cannot really avoid calling `permittedFor()` because the `withClaim()` in 3.4 calls `forwardCallToCorrectClaimMethod()` internally, which calls `permittedFor()` for `'aud'` claims. So it will always cast the parameter to a string, changing any (previously valid) array to the string "Array", and messing with the application - or in my case, the test cases.

So I would consider this change a backwards incompatible behaviour regarding `withClaim()`, and it should be explained to the developer somehow. The least intrusive way is the deprecation message.

lcobucci:
@SvenRtbg I completely understand your point and frustration. My main concern is that you were relying on something that wasn't officially supported so, I struggle to label it as a BC-break.

However, let's summon other maintainers here. @Ocramius @Slamdunk what are your thoughts on the matter?

Slamdunk:
`3.3` adhered to RFC better than `3.4`, so I agree with @SvenRtbg 

What have been done in `3.4` should be considered to me only an oversight, affecting a minority of users.
I like the message of `trigger_error` written in the `getClaim` method, and I'd adopt a similar one into `permittedFor` as well:
```diff
-Using arrays for audience claims is not really supported in 3.4 and will behave different in 4.0.
+Support to arrays for audience was mistakenly removed from 3.3 in 3.4. In 4.0 this will change again with more flexibility, please consult the documentation.
```
And then I'd tag a new patch release for `3.4`.

lcobucci:
The discussion isn't about adherence but the fact that it has never been officially supported.

It used to work by mistake, which was made clearer on v3.4. The official support only came in v4.

So how can we label the fact that something used to work by mistake and now doesn't work anymore as a BC-break?

Slamdunk:
> So how can we label the fact that something used to work by mistake and now doesn't work anymore as a BC-break?

I see you are referring to the `string` doc-block [already mentioned here](https://github.com/lcobucci/jwt/pull/630#discussion_r565695175).

There is no right answer: it was a mistake to not dynamically type-check the variables in the first place, as well as it is a user mistake to not use static analysis.

`3.3` is the past, `4.0` is the future, `3.4` is a bridge between the two: as is, I don't value the correct label for the issue, but rather the clear message to the users on what happened and how to remedy. This PR, maybe with better messages in the `trigger_error`, is good at aiming this goal, imho.

SvenRtbg:
I do like @Slamdunk 's suggestion for the deprecation message better. Will change that. The comments regarding the upgrade file will also be considered.

As has been said before, arrays used to work in 3.3, they stopped working in 3.4, and work again in 4.0. I had to figure out why my test cases failed after updating to the compatible 3.4 version, and I'd very much like to have something reassuring message that it isn't my code that is still broken during upgrades, but a use case that will be supported again.

Thanks for your feedback.

docs/upgrading.md
SvenRtbg: Rephrased and put emphasis on deprecation messages, and using the new features detailed later in this document.

docs/upgrading.md
SvenRtbg: Will reformat.

docs/upgrading.md
SvenRtbg: I think I found the hidden reason, and illustrated it in the new version of this part.

src/Token.php
SvenRtbg: This deprecation message is targeting the readback from a token, and if some external system is adding more than one audiences as an array, this will trigger on accessing the claim.

src/Builder.php
SvenRtbg: I used the suggested message from @Slamdunk below.

lcobucci:
@SvenRtbg @Slamdunk thank you very much for your patience and making me think.

> 3.3 is the past, 4.0 is the future, 3.4 is a bridge between the two

This is a simple and yet very interesting sentence - and it made me reconsider the whole thing.

Considering v3.4 as the bridge between the two versions, I've created a bridge with a missing plank.

I took the deliberate decision of not backporting the support for multiple audiences and that was a **wrong decision**. It was wrong simply because the interface/behaviour isn't the same as v4.

Making the old API behave as the new one requires a minor BC-break (claim replication, broken functionality which wasn't removed when introducing the new builder methods in v3.3 or so). 

I've created a patch to fix this issue (https://github.com/lcobucci/jwt/pull/656) and would like for us to move to that direction instead - even if we have to do a minor BC-break.

Would you please review it and share your thoughts there?

If we all agree with that solution, we can repurpose this PR to keep some of the improvements @SvenRtbg done on the docs and release it together with v3.4.4.

Thank you again for your patience throughout the discussion and sorry for being attached to such petty details - sometimes I get stuck in useless peculiarities.

SvenRtbg:
I like deprecation messages, so I added them to the `Token` class for every deprecated method. I'm not sure if this is going too far, however one has to deal with deprecated methods anyway, and the easier they are to spot the easier they are to remove.

Regarding "too far": I have the feeling that I'd probably have to deal with internal calls to methods triggering deprecation notices in order to get rid of them, otherwise one would not be able to upgrade a code base to be free of them.

I intentionally left the upgrade document in a separate commit, to be able to cherrypick it. Let me know what you think.

SvenRtbg:
The library still exposes an asymmetric behavior when dealing with audience arrays and using the deprecated methods, as described in my PR now. So without an update to the new Token API, there will be errors appearing in edge cases. But I think the world can live with that regardless.

I'm not sure I added this fact into the `ParserTest` sufficiently, I just duplicated the test cases that seem to accidentally use the audience claim as example data.

lcobucci:
> I like deprecation messages, so I added them to the Token class for every deprecated method. I'm not sure if this is going too far, however one has to deal with deprecated methods anyway, and the easier they are to spot the easier they are to remove.
>
> Regarding "too far": I have the feeling that I'd probably have to deal with internal calls to methods triggering deprecation notices in order to get rid of them, otherwise one would not be able to upgrade a code base to be free of them.

My (implicit) policy is to only add the deprecation messages for things that can't be converted with `@deprecated` (like deprecating arguments or deprecation of logic blocks). IDEs and tools like `phpstan/phpstan-deprecation-rules` help people to move away from components annotated with `@deprecated`, which reduces the noise generated by the library.

With that said, let's please only add the messages when using arrays on `aud` claim via the old API :+1: 

> I intentionally left the upgrade document in a separate commit, to be able to cherrypick it. Let me know what you think.

I reaaaaaaaaally appreciate your commit organisation :heart:

docs/upgrading.md
```diff
@@ -10,6 +10,31 @@ We're building on the version `v3.4.0` a forward compatibility layer to help use
 To help on the migration process, all deprecated components are being marked with `@deprecated` and deprecated behaviour will trigger a `E_USER_DEPRECATED` error.
 However, you can also find here the instructions on how to make your code compatible with both versions.
 
+### General migration strategy
+
+Update your existing software to the latest 3.4.x version using composer:
+
+```sh
+    composer require lcobucci/jwt ^3.4
```
lcobucci:
```suggestion
composer require lcobucci/jwt ^3.4
```

docs/upgrading.md
```diff
@@ -10,6 +10,31 @@ We're building on the version `v3.4.0` a forward compatibility layer to help use
...
+
+Then run your tests and fix all deprecation notices.
+
+Note that PHPUnit tests will only fail if you have the `E_USER_DEPRECATED` error level activated (it is a good idea to run tests using `E_ALL`, btw.), and deprecation messages during the execution of data providers will not affect the test result at all, only output text to the console, because they are called before test execution starts.
+Make sure you do not see any of them before you continue.
+
+Now you can upgrade to the latest 4.x version:
+
+```sh
+    composer require lcobucci/jwt ^4.0
```
lcobucci:
```suggestion
composer require lcobucci/jwt ^4.0
```

docs/upgrading.md
```diff
@@ -10,6 +10,31 @@ We're building on the version `v3.4.0` a forward compatibility layer to help use
...
 
+### General migration strategy
+
+Update your existing software to the latest 3.4.x version using composer:
+
+```sh
+    composer require lcobucci/jwt ^3.4
+```
+
+Then run your tests and fix all deprecation notices.
```
lcobucci, line 21:
```suggestion
Then run your tests and tools like [`phpstan/phpstan-deprecation-rules`](https://github.com/phpstan/phpstan-deprecation-rules) to fix all deprecation notices.
```

docs/upgrading.md
```diff
@@ -10,6 +10,31 @@ We're building on the version `v3.4.0` a forward compatibility layer to help use
...
+
+Note that PHPUnit tests will only fail if you have the `E_USER_DEPRECATED` error level activated (it is a good idea to run tests using `E_ALL`, btw.), and deprecation messages during the execution of data providers will not affect the test result at all, only output text to the console, because they are called before test execution starts.
```
lcobucci: This sentence is a bit long... what do you think about this:

```suggestion
Note that PHPUnit tests will only fail if you have the `E_USER_DEPRECATED` error level activated - it is a good practice to run tests using `E_ALL`.
Deprecation messages during the execution of data providers will not affect the test result at all, only output text to the console, because they are called before test execution starts.
```

docs/upgrading.md
```diff
@@ -10,6 +10,31 @@ We're building on the version `v3.4.0` a forward compatibility layer to help use
...
+Make sure you do not see any of them before you continue.
+
+Now you can upgrade to the latest 4.x version:
+
+```sh
+    composer require lcobucci/jwt ^4.0
+```
+
+Remember that some deprecation messages from the 3.4 version may have notified you that things still are different in 4.0, so you may find you need to adapt your own code once more at this stage.
+Not every method call in the 3.4 version is 100% identical to the 4.0 counterpart.
```
lcobucci: @SvenRtbg wasn't that mitigated with the backport of multiple audiences? Do you have other examples that we need to address?

I do expect that code that uses the forward compatibility layer should simply work on v4.x.

docs/upgrading.md
```diff
@@ -153,10 +178,13 @@ Here's the migration:
...
-Even though we didn't officially support multiple audiences, it was technically possible to achieve that by manually setting the `aud` claim.
+Even though we didn't officially support multiple audiences, it was technically possible to achieve that by manually setting the `aud` claim to an array with multiple strings.
+
+If you parse a token with 3.4, and read it's contents with `\Lcobucci\JWT\Token#getClaim()` or`\Lcobucci\JWT\Token#getClaims()`, you will only get the first element of such an array back.
+If the audience claim does only contain a string, or only contains one string in the array, nothing changes.
+Please [upgrade to the new Token API](#use-the-new-token-api) for accessing claims in order to get the full audience array again (e.g. call `Token#claims()->get('aud')`).
 
-For this case, there are different approaches to migrate your code.
-Feel free to choose your preferred one.
+When creating a token, use the new method `Builder#permittedFor()` as detailed below:
```
lcobucci:
```suggestion
When creating a token, use the new method `Builder#permittedFor()` as detailed below.
```

lcobucci (commented):
Just some feedback on the docs :)

docs/upgrading.md
SvenRtbg: I'm struggling a bit with this part of the text because your suggestion is different from my experience, but I will add the tool link.

docs/upgrading.md
SvenRtbg: I think I was referring to the fact that deprecated methods, or more precisely `getClaim('aud')`, changed the return value compared to 3.3. And that difference does still exist. For a deprecated method that should not be used... it's probably unwarranted to talk about an undetermined amount of methods that "may" be different.

SvenRtbg: I opted for removing that sentence. I do not remember enough of my migration process to have proof for this statement, and I don't want to redo that work again. :)

SvenRtbg: Is there a reason the 4.0 thru 4.2 `Token` interface has no `claims()` method defined? I'm comparing 3.4 `Token` and 4.0 `Token` side-by-side right now to see if the method signatures match (reasonably).

SvenRtbg: `payload()` and `signature()` also seem to be missing.

SvenRtbg: I found them in the 4.2 variant...

SvenRtbg:
I think we are good now: Improved upgrade guide, and the offending issue has a dedicated deprecation message with test case.

lcobucci (approved):
That's perfect. Thanks a lot for your help @SvenRtbg!

lcobucci:
@SvenRtbg all released! I'll do some manual merge-ups since the release automation failed due to branch naming and all.

Thanks again!

PR #632, opened 2021-01-15
Use the correct class name for in-memory keys in documentation

lots0logs:
Update configuration docs to match current code implementation.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -19,7 +19,7 @@ In order to use it, you must:
 
 ### Configuration initialisation
 
-The `Lcobucci\JWT\Signer\Key` object is used for symmetric/asymmetric signature.
+The `Lcobucci\JWT\Signer\Key\InMemory` object is used for symmetric/asymmetric signature.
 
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -81,9 +81,11 @@ use Lcobucci\JWT\Signer;
 use Lcobucci\JWT\Signer\Key\LocalFileReference;
 use Lcobucci\JWT\Signer\Key\InMemory;
 
+// You may use RSA OR ECDSA and all their variations (256, 384, and 512) and EdDSA over Curve25519
+$signer        = new Signer\Rsa\Sha512();
+$signer        = Signer\Ecdsa\Sha512::create();
```
lcobucci: I think we should use an alert to convey this difference (instead of re-assigning the variable).

lcobucci (changes requested):
Thanks @lots0logs!

Just a small detail, we should also change this in v4.0.x branch (I'll apply the change and rebase things)

lcobucci (approved):

PR #637, opened 2021-01-21
Changed string return type to float according to JWT specs

enrico-atzeni:
NumericDate of `exp` and `iat` claims do not respect JWT specifications as of here
`exp`: https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4.1.4
`iat`: https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4.1.6

both says
> [...] Its value MUST be a number containing a NumericDate value.  [...]

but `MicrosecondBasedDateConversion`'s `convertDate` method returns a string instead of a number.

`return $seconds . '.' . $microseconds;`

Fixed it casting the string to a **float**:

`return (float) ($seconds . '.' . $microseconds);`

### DIFF
--- a/src/Encoding/MicrosecondBasedDateConversion.php
+++ b/src/Encoding/MicrosecondBasedDateConversion.php
@@ -25,7 +25,7 @@ final class MicrosecondBasedDateConversion implements ClaimsFormatter
         return $claims;
     }
 
-    /** @return int|string */
+    /** @return int|float */
     private function convertDate(DateTimeImmutable $date)
     {
         $seconds      = $date->format('U');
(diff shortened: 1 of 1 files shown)


Slamdunk:
Duplicate of https://github.com/lcobucci/jwt/pull/618
See also https://github.com/lcobucci/jwt/pull/617 https://github.com/lcobucci/jwt/discussions/623

enrico-atzeni:
thanks @Slamdunk, didn't find it before, I will use the fix mentioned in #618

lcobucci:
Hello folks, thanks to @yassinrais we managed to [fix this compatibility issue whilst still preserving the microseconds precision](https://github.com/lcobucci/jwt/pull/706). We already released new versions to solve the problem.

PR #656, opened 2021-02-06
Fix forward compatibility layer gap for audience claim 

lcobucci:
When backporting the v4's API we wrongly decided to preserve the lack of support to multiple audiences.

The problem with that decision is that it makes the FC-layer not totally compatible with the future, making it harder for users to properly migrate their code.

Even though it wasn't officially supported, users could configure multiple audiences using the method `Builder#set()` or
`Builder#withClaim()`.

This backports the support for multiple audiences whilst maintaining the deprecated functionality of replicating headers.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -15,10 +15,16 @@ use Lcobucci\JWT\Token\DataSet;
 use Lcobucci\JWT\Token\RegisteredClaimGiven;
 use Lcobucci\JWT\Token\RegisteredClaims;
 
+use function array_diff;
+use function array_filter;
 use function array_key_exists;
+use function array_merge;
+use function array_shift;
+use function count;
 use function current;
 use function in_array;
 use function is_array;
+use function is_bool;
 use function trigger_error;
 use const E_USER_DEPRECATED;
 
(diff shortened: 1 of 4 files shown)


src/Builder.php
```diff
@@ -95,20 +98,24 @@ public function __construct(
...
      * Configures the audience
      *
-     * @param string $audience
-     * @param bool $replicateAsHeader
+     * @param list<string> $audiences
      *
      * @return Builder
      */
-    public function permittedFor($audience, $replicateAsHeader = false)
+    public function permittedFor(...$audiences)
```
Slamdunk, line 114: I think we should check whether the last `$audiences` element is a `boolean`, and inform the user that, for this method, BC broke to favor FC

lcobucci: Yeap, just had the same thought before going to bed. I'll update it 👍

Slamdunk (approved):

src/Builder.php
lcobucci: @Slamdunk just applied that with a filter for string audiences only - the filter might cause minor issues but it's more compatible with new the behaviour (we can change to a string cast btw).

SvenRtbg:
Wow, you are quick. However, what about reading back an arbitrary JWT with this library, containing multiple audience values in an array? That code hasn't changed, or has it? If not, then it is still filtering out the first entry and not giving back the rest.

lcobucci:
@SvenRtbg I believe that's covered by https://github.com/lcobucci/jwt/pull/656/files#diff-6afad99ba73827781eb1ec0ddfa0dc1dbc2db1e04f56ca4a51a65748f82f8b65R225 would you like to perform additional tests with the branch before we release?

SvenRtbg:
I'm referring to this code here: https://github.com/lcobucci/jwt/blob/3.4/src/Token.php#L234-L236

SvenRtbg:
And another instance is here: https://github.com/lcobucci/jwt/blob/3.4/src/Claim/Factory.php#L69-L71

lcobucci:
That's expected not to work and won't be changed because `$token->getClaim()` should be replaced by `$token->claims()->get()`. In a nutshell, only the new API will be affected by this backport.

I know that used to work on the v3.3 with the manual claim configuration but applying the support to that too is a bit too far IMHO.

SvenRtbg:
What about adding a deprecation message for this case then? Yes, the `Token::getClaim()` and `Token::getClaims()` methods are marked deprecated and are somewhat pointing to the new method.

Anyways, I'd try to get back to this on monday.

lcobucci:
@SvenRtbg a deprecation message would be alright IMHO :+1:

SvenRtbg:
Ok, I'll have a look again right now, rebasing my own PR and trying to find a spot for this depraction message. I'll have to rephrase the update guide anyways.

PR #657, opened 2021-02-06
Migrate to native dependabot

lcobucci:


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+  directory: "/"
+  schedule:
+    interval: daily
+    time: "06:00"
+  reviewers:
+    - lcobucci
+  versioning-strategy: increase
+  open-pull-requests-limit: 20
+  allow:
+    - dependency-type: all
```
Slamdunk, line 13: May I ask you if this is really necessary?
I fear that so many notifications just end up being trashed by everyone, even the ones that may have value been ~reviewed~ supervisioned.
What about `direct` only?

DOC: https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#allow

lcobucci: I want for all dependencies to be as up-to-date as possible. There's tooling to auto-merge everything that passes CI so there's very little tracking to be done.

I do understand the concern around the volume of notifications there's also rate limiting happening sometimes... the trade-offs in play here is to ensure that development of the library works in the same way for everybody and that nobody has to update dependencies manually.

I'd love if dependabot would run things with `composer require x -W` to update the dependencies of our dependencies together with the changes being introduced but that's not the case now 😞

lcobucci: (I do hate when symfony update their packages without releasing anything, though)

PR #667, opened 2021-02-11
Support composer classmap-authoritative option

### LINKED ISSUE #666: [3.4.*] composer aliases missing when classmap-authoritative is applied
**Affected version:** 3.4
**PHP Version:** 7.3.22

Encountered the issue in https://github.com/scheb/2fa/issues/57.

It appears the `lcobucci/jwt/compat/class-aliases.php` is not properly configured to support the `classmap-authoritative`, expecting `class_exists()` to call the autoloader which is not run when `classmap-authoritative` is applied.

Thoughts on changing `class-aliases.php` to?:

```php
class_exists(\Lcobucci\JWT\Token\Plain::class) || class_alias(\Lcobucci\JWT\Token::class, \Lcobucci\JWT\Token\Plain::class);
class_exists(\Lcobucci\JWT\Token\Signature::class) || class_alias(\Lcobucci\JWT\Token::class, \Lcobucci\JWT\Token\Signature::class);
```

fyrye:
Fix #666 
Forcibly calls `class_alias()` when `class_exists()` autoload feature fails due to composer `classmap-authoritative` option being enabled.

### DIFF
--- a/compat/class-aliases.php
+++ b/compat/class-aliases.php
@@ -1,5 +1,4 @@
 <?php
 
-class_exists(\Lcobucci\JWT\Token\Plain::class);
-class_exists(\Lcobucci\JWT\Token\Signature::class);
-
+class_exists(\Lcobucci\JWT\Token\Plain::class, false) || class_alias(\Lcobucci\JWT\Token::class, \Lcobucci\JWT\Token\Plain::class);
+class_exists(\Lcobucci\JWT\Token\Signature::class, false) || class_alias(\Lcobucci\JWT\Signature::class, \Lcobucci\JWT\Token\Signature::class);
--- a/src/Token/Plain.php
+++ b/src/Token/Plain.php
@@ -5,4 +5,4 @@ namespace Lcobucci\JWT\Token;
 use Lcobucci\JWT\Token;
 use function class_alias;
 
-class_alias(Token::class, Plain::class);
+class_exists(Plain::class, false) || class_alias(Token::class, Plain::class);
--- a/src/Token/Signature.php
+++ b/src/Token/Signature.php
@@ -5,4 +5,4 @@ namespace Lcobucci\JWT\Token;
 use Lcobucci\JWT\Signature as SignatureImpl;
 use function class_alias;
 
-class_alias(SignatureImpl::class, Signature::class);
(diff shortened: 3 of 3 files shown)


lcobucci:
@fyrye nice catch, I hate the fact we had to introduce class aliases 😭

Just one question: can we have the alias only on this file instead? It was nice to have the "class" files being on the expected folder and all but having the beast contained sounds like a better idea to me.

fyrye:
@lcobucci yes it is possible to use only `class_alias()` instead. 
However, most of the libraries I have seen use something like `class_exists(Bar::class, false) || class_alias(Foo::class, Bar::class);`.
That way there is a safe-guard to prevent reassigning the alias, which will emit a warning and uses the `false` flag to prevent the autoloader from attempting to load the file.
If you have no objections, I can apply the `false` flag to the pull request for the same reasons.

fyrye:
The coverage error is being caused by the `class_alias()` call on the `Token\Signature.php` and `Token\Plain.php` files, since composer is including the `compat/class-aliases.php` and the coverage scans the files.

To resolve; the appropriate `class_exists() || class_alias()` line would need to be added to the individual token files, alternatively fixed by removing the `false` flag to have the autoloader load the token files.

lcobucci:
What I meant with my question is if it would be possible to completely get rid of the files that declare the aliases inside of `src`. Leaving that job exclusively to `class-aliases.php`.

Does that make sense to you?

fyrye:
@lcobucci yes, they can be safely removed in favor of relying solely on `compat/class-aliases.php` to define the aliases, since the individual `Token\Plan.php` and `Token\Signature.php` files do not actually declare a class.

However, if an autoloader other than composer is used by others, it may cause issues with their usage.
I do know of many developers who choose to manually download libraries and use a custom autoloader instead of using composer. 
With that said, leaving the `src/` files shouldn't cause any issues and serves as a placeholder should people use `include` manually on those files.

fyrye:
Apologies with all the forced pushes, I was resolving the "Unverified" commit issue and squashed my changes.

lcobucci:
That's alright, thanks for you changes.

I don't like the duplication but, honestly, v3 needs to die. So, let's get this released.

PR #670, opened 2021-02-12
Fix compatibility issues with older libsodium version (< 1.0.14)

### LINKED ISSUE #665: Call to undefined function sodium_bin2base64
After last update 4.1.0 i got error  *Call to undefined function sodium_bin2base64() {"exception":"[object] (Error(code: 0): Call to undefined function sodium_bin2base64() at vendor/lcobucci/jwt/src/Encoding/JoseEncoder.php:50)* my php version is 7.4.13  and sodium version 1.0.13. If i change sodium version to 1.0.18 everything is fine.

Slamdunk:
Fixes #665 

The solution proposed here aims at all the following goals together:

1. Do not let user choose: use the safest implementation available
2. Keep the code simple: the polyfill is not really a polyfill since it's not cache-time-safe resistant, but its simple
3. Test the most without altering CI build
4. Achieve the most code-coverage with the fewest `@codeCoverageIgnore` annotations
5. Keep the 100% mutation score with the fewest `ignore` additions to `infection.json.dist`
6. Do not pollute the global namespace, be compatible with https://github.com/paragonie/sodium_compat
7. Remove dependency from `SodiumException` to avoid polyfilling it

### DIFF
--- a/src/Encoding/CannotDecodeContent.php
+++ b/src/Encoding/CannotDecodeContent.php
@@ -6,7 +6,6 @@ namespace Lcobucci\JWT\Encoding;
 use JsonException;
 use Lcobucci\JWT\Exception;
 use RuntimeException;
-use SodiumException;
 
 final class CannotDecodeContent extends RuntimeException implements Exception
 {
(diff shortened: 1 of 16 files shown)


src/Encoding/CannotDecodeContent.php
```diff
@@ -15,8 +14,8 @@ public static function jsonIssues(JsonException $previous): self
         return new self('Error while decoding from JSON', 0, $previous);
     }
 
-    public static function invalidBase64String(SodiumException $sodiumException): self
+    public static function invalidBase64String(): self
```
Slamdunk, line 17: Not a BC-Break according to https://github.com/lcobucci/jwt/pull/613#discussion_r544938427

lcobucci (approved):
A bit sad that we can't really force the requirements for the extension buuut ces't la vie.

Thanks @Slamdunk!

Issue #140, opened 2016-11-07
Provide a high level API to simplify usage


lcobucci:
@Ocramius gave a nice idea to help users: based on the `Configuration` have a really simple way to get a valid token or null from a raw token (string), something like:

```php
function (string $tokenData) use ($configuration): ?Token {
    $parser = $configuration->getParser();
    $validator = $configuration->getValidator();
    $constraints = $configuration->getConstraints();

    try {
        $token = $parser->parse($tokenData);
    } catch (\Lcobucci\JWT\Exception $e) {
        return null;
    }

    if (!$validator->validate($token, ...$constraints) {
        return null;
    }

    return $token;
};
```

In order to achieve this we would need to implement #139 but also add the validation constraints in the configuration object.

lcobucci:
Depends on #129 (sure)

PR #680, opened 2021-02-16
Add misuse resistant SecureParser

### LINKED ISSUE #663: Misuse resistant API
Hi, one thing I like very much about [PASETO](https://paseto.io/) is it's impossible to use it in insecure ways.
I'm not talking about JWT RFC flaws, but rather about API that we can and should implement to force a safe usage.

Here are two examples of `lcobucci/jwt:v4` misuse:

```php
/**
 * 1) Validation missing
 */
$jwtToken  = (new JwtParser(new JoseEncoder()))->parse($_POST['jwt']);
loginUser($jwtToken->claims()->get(RegisteredClaims::AUDIENCE));

/**
 * 2) Validation typo
 */
$jwtToken  = (new JwtParser(new JoseEncoder()))->parse($_POST['jwt']);
// It should be assert()
(new Validator())->validate(
    $jwtToken,
    new SignedWith(new Eddsa(), InMemory::base64Encoded('foo'))
);
loginUser($jwtToken->claims()->get(RegisteredClaims::AUDIENCE));
```

For `v5` I propose these main BC-breaks:

1. API **MUST** require and consume a `SignedWith` validator in order to `parse` a JWT Token
2. API **MUST NOT** allow `claims` part to be read nor consumed before the `parse` call
3. API **MUST** allow `headers` part to be read and consumed before the `parse` call
4. API **SHOULD** accept additional validators to be consumed on `parse` call
5. (*bonus*) API **MUST** require and consume either a `LooseValidAt` or `StrictValidAt` validator in order to `parse` a JWT Token

Slamdunk:
Close #663 

- [x] Expose new `SecureParser` API
- [x] Add `SecureParser` implementations into `Configuration`
- [x] ~Deprecate `Parser`~ no deprecation for now
- [ ] Document `SecureParser` and `Parser-to-SecureParser` upgrade path

Documentation will be updated here after implementation review.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -110,13 +110,11 @@ final class Configuration
         return ($this->builderFactory)($claimFormatter ?? ChainedFormatter::default());
     }
 
-    /** @deprecated Use SecureParser instead. */
     public function parser(): Parser
     {
         return $this->parser;
     }
 
-    /** @deprecated Use SecureParser instead. */
     public function setParser(Parser $parser): void
     {
         $this->parser = $parser;
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -110,11 +110,13 @@ final class Configuration
         return ($this->builderFactory)($claimFormatter ?? ChainedFormatter::default());
     }
 
+    /** @deprecated Use SecureParser instead. */
     public function parser(): Parser
     {
         return $this->parser;
     }
 
+    /** @deprecated Use SecureParser instead. */
     public function setParser(Parser $parser): void
     {
         $this->parser = $parser;
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -19,11 +19,13 @@ use Lcobucci\JWT\Validation\Constraint;
  */
 final class Configuration
 {
+    private Decoder $decoder;
     private Parser $parser;
     private Signer $signer;
     private Key $signingKey;
     private Key $verificationKey;
     private Validator $validator;
+    private ?SecureParser $secureParser = null;
 
     /** @var Closure(ClaimsFormatter $claimFormatter): Builder */
     private Closure $builderFactory;
@@ -38,10 +40,12 @@ final class Configuration
         ?Encoder $encoder = null,
         ?Decoder $decoder = null
     ) {
+        $this->decoder = $decoder ?? new JoseEncoder();
+
         $this->signer          = $signer;
         $this->signingKey      = $signingKey;
         $this->verificationKey = $verificationKey;
-        $this->parser          = new Token\Parser($decoder ?? new JoseEncoder());
+        $this->parser          = new Token\Parser($this->decoder);
         $this->validator       = new Validation\Validator();
 
         $this->builderFactory = static function (ClaimsFormatter $claimFormatter) use ($encoder): Builder {
@@ -114,6 +118,20 @@ final class Configuration
     public function setParser(Parser $parser): void
     {
         $this->parser = $parser;
+    }
+
+    public function secureParser(): SecureParser
+    {
+        if ($this->secureParser === null) {
+            $this->secureParser = new Token\SecureParser($this->decoder, $this->validator);
+        }
+
+        return $this->secureParser;
+    }
+
+    public function setSecureParser(SecureParser $secureParser): void
+    {
+        $this->secureParser = $secureParser;
     }
 
     public function signer(): Signer
--- a/src/Token/SecureParser.php
+++ b/src/Token/SecureParser.php
@@ -1,0 +1,29 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Token;
+
+use Lcobucci\JWT\Decoder;
+use Lcobucci\JWT\SecureParser as SecureParserInterface;
+use Lcobucci\JWT\Token as TokenInterface;
+use Lcobucci\JWT\Validation\Constraint;
+use Lcobucci\JWT\Validation\SignedWith;
+use Lcobucci\JWT\Validation\ValidAt;
+use Lcobucci\JWT\Validator;
+
+final class SecureParser implements SecureParserInterface
+{
+    private Parser $parser;
+    private Validator $validator;
+
+    public function __construct(Decoder $decoder, Validator $validator)
+    {
+        $this->parser    = new Parser($decoder);
+        $this->validator = $validator;
+    }
+
+    public function parse(string $jwt, SignedWith $signedWith, ValidAt $validAt, Constraint ...$constraints): TokenInterface
+    {
+        return $this->parser->parse($jwt);
+    }
+}
--- a/src/Token/SecureParser.php
+++ b/src/Token/SecureParser.php
@@ -22,7 +22,7 @@ final class SecureParser implements SecureParserInterface
         $this->validator = $validator;
     }
 
-    public function parse(
+    public function parseJwt(
         string $jwt,
         SignedWith $signedWith,
         ValidAt $validAt,
--- a/src/Token/SecureParser.php
+++ b/src/Token/SecureParser.php
@@ -22,8 +22,21 @@ final class SecureParser implements SecureParserInterface
         $this->validator = $validator;
     }
 
-    public function parse(string $jwt, SignedWith $signedWith, ValidAt $validAt, Constraint ...$constraints): TokenInterface
-    {
-        return $this->parser->parse($jwt);
+    public function parse(
+        string $jwt,
+        SignedWith $signedWith,
+        ValidAt $validAt,
+        Constraint ...$constraints
+    ): TokenInterface {
+        $token = $this->parser->parse($jwt);
+
+        $this->validator->assert(
+            $token,
+            $signedWith,
+            $validAt,
+            ...$constraints
+        );
+
+        return $token;
     }
 }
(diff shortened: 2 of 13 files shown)


src/Configuration.php
```diff
@@ -106,16 +110,32 @@ public function builder(?ClaimsFormatter $claimFormatter = null): Builder
...
+    /** @deprecated Use SecureParser instead. */
     public function setParser(Parser $parser): void
     {
         $this->parser = $parser;
     }
 
+    public function secureParser(): SecureParser
+    {
+        if ($this->secureParser === null) {
+            $this->secureParser = new Token\SecureParser($this->decoder, $this->validator);
```
Slamdunk, line 126: Lazy loading is compelled here by BC constructor

src/Token/SecureParser.php
```diff
@@ -0,0 +1,42 @@
...
+use Lcobucci\JWT\Validator;
+
+final class SecureParser implements SecureParserInterface
+{
+    private Parser $parser;
+    private Validator $validator;
+
+    public function __construct(Decoder $decoder, Validator $validator)
+    {
+        $this->parser    = new Parser($decoder);
```
Slamdunk, line 21: `Parser` dependency has been intentionally internalized: if and when it will be removed, `SecureParser` API should not be affected and we can just copy-paste its code here.

test/unit/Token/SecureParserTest.php
```diff
@@ -0,0 +1,627 @@
...
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use RuntimeException;
+
+use function uniqid;
+
+/**
+ * @coversDefaultClass \Lcobucci\JWT\Token\SecureParser
+ *
+ * @uses  \Lcobucci\JWT\Token\Parser
```
Slamdunk, line 24: Since `Parser` dependency is hardcoded internally, this test was actually copy-pasted as the two classes, from outside, are not related.

lcobucci:
@Slamdunk what was the reason for going this way instead of the proposed solution? This changes a lot of things that should not be affect IMHO.

lcobucci:
Just to clarify, my main question is that I believe there's no need for a different interface (nor deprecating the existing one).

It makes things much more complicated.

Slamdunk:
> what was the reason for going this way instead of the proposed solution?

I read [your suggestion](https://github.com/lcobucci/jwt/issues/663#issuecomment-777290151) before coding, but it presents some issues:

1. `SecureParser::parse` must `@throws RequiredConstraintsViolated` making it not BC with `Parser::parse`
2. Injecting constraints in the constructor would make `SecureParser` token-specific, in contrast to `Parser` which is token-agnostic
3. It would make user code extremely dangerous: the final user diff would look like
```diff
 $token = $config->parser()->parse('...');

-$config->validator()->assert(
-    $token,
-    $config->validationConstraints()
-);

 loginUser($token->claims()->get('sub'));
```
What if the `Parser` instance is then changed again, but the user forgets to re-write the assertion?

Instead the proposed `SecureParser` here has no room for doubt.

> It makes things much more complicated.

May I ask you how?

lcobucci:
> 1. SecureParser::parse must @throws RequiredConstraintsViolated making it not BC with Parser::parse

I don't think we should throw that exception. Using `Validator#validate()` would be more appropriate for this kind of thing since we'd stop on the first error.

> 2. Injecting constraints in the constructor would make SecureParser token-specific, in contrast to Parser which is token-agnostic

The proposed design is nothing but a decorator, which is the most appropriated approach here since we want to decorate the parsing behaviour with validation.

> 3. It would make user code extremely dangerous: the final user diff would look like

That depends on how we implement (check the link below).

> May I ask you how?

I think using code is simpler than explaining in words, check this: https://github.com/lcobucci/jwt/pull/682

My main wishes here are:

1. Have things working side-by-side: there's no need (yet) to deprecate things as we're using this as an experiment to check user adoption
2. Reduce as much as possible the number of changes in the code: the API is already extensive enough and we should use the provided extension points

Slamdunk:
First things first.

I am aware of the 100k daily average downloads of this excellent library: APIs should be concise, well thought, extensively documented and change as little as possible to provide smooth, clear and reliable upgrades when modifications happen.

I do strongly believe that, in its current form, API has security flaws that cannot be overcome with both:

1. A `parse(string $jwt, SignedWith $signedWith): Plain` signature (as minimum)
2. A compelled transition to it (with deprecations as first step)

The solution proposed in https://github.com/lcobucci/jwt/pull/682 doesn't solve the 3rd point of https://github.com/lcobucci/jwt/pull/680#issuecomment-780344470 as far as I can tell, and my belief is that security shouldn't be optional.
I would push my ideas even though against the RFC, but it occurs that they are going right in the RFC direction, that this library initially missed.

You've said "*My main wishes here are*": beside the important goals I wrote here about API, I don't understand why you are reluctant to change them; my best guess is that you don't agree with me, as you haven't provided better arguments than personal preferences.

I'm totally fine with them, I respect them and the amazing work you've done with this little yet powerful library, as well as its property up to you.
But if you prefer not to follow my advice for both a new signature and deprecations, I prefer to dismiss the argument (and PR) in its entirety as I believe #682 will do much more harm than good.

Slamdunk:
@lcobucci I'm still ready to fix every code and stylistic issue in this PR, as well as headstrong about the correctness and the need of the presented API.

But after 2 months, I need to free my mind of this, either by updating the code or closing as rejected, so I'd like to ask you a definitive decision about the topic.

lcobucci:
@Slamdunk sorry about the delay here... my mind is going crazy about multiple things atm.

My issue with replacing the current parser with one that verifies the signature right after deserialising the token is that it limits the design of the API, disallowing current users to use different keys based on the issuer and/or the token (relying on the `kid` header).

That also limits the possibility of parsing encrypted tokens since the signature can only be verified after the decryption, which is something that I still want to support.

My suggestion of creating something on the side was to make the "generic scenario" more secure and easy to handle but not being limited to it.

I hope that gives you enough clarity on where I'm coming from. I'd leave it up to you to decide to turn this PR into something else or simply close it as "won't fix". :+1:

Slamdunk:
> My issue with replacing the current parser with one that verifies the signature right after deserialising the token is that it limits the design of the API, disallowing current users to use different keys based on the issuer and/or the token (relying on the `kid` header).

This is exactly what I expect from a secure implementation

> That also limits the possibility of parsing encrypted tokens since the signature can only be verified after the decryption, which is something that I still want to support.

I have no knowledge over encrypted token, but AFAIK they have no signatures, and the parsing should be way different from regular tokens; but I'm talking about something I don't know so I trust you here.
I guess that I should rename `SecureParser` to `SecureJwtParser` or `SecureParser::parse` to `SecureParser::parseJwt` then, or something more restricted to its scope, tell me

> I hope that gives you enough clarity on where I'm coming from. I'd leave it up to you to decide to turn this PR into something else or simply close it as "won't fix". +1

Democracy is the art of compromise, and since February I have reassessed the priority and importance of the proposed implementation. I think the secure API is vital, and having it here instead of a fork can have greater impact over the overall userland code.
So now I think it's good to pursue a smoother path and do not `@deprecate` the current parser API: would you accept this PR with this change?

Slamdunk:
Code updated as "discussed"

lcobucci:
@Slamdunk I took a couple of days to think about the topic and I'm on-board with the idea.

I think we have naming issues (maybe caused by the fact that the current parser is actually a deserialiser).

I also believe that the scenarios I've described are important to consider - they might not be the most common path, though. We can address them by looking at JWK concepts - e.g.: passing a keyset that resolves the right key to be used based on the key identifier present on the token (throwing an exception when the key isn't listed).

That's obviously an idea that needs to be better explored.

With that said, I'd like for us to put some thought about these possibilities before merging this PR (mainly because I think we'd find better alternatives that accommodates all the needs).

I know it's a bit frustrating to postpone things even further but would you be up for having a synchronous whiteboard session to design things?

Slamdunk:
I would be very glad to, and I hope also @Ocramius can join us.
I'm on GMT+2

Ocramius:
We discussed this on a call.

Pending review by @lcobucci and @Slamdunk:

 * we want to create a simplified misuse-resistant API similar to what is #140
 * the misuse-resistant API must be in form of a factory: `makeJwtThing(SymmetricKey | KeyPair $key): GimmeToken`
 * `GimmeToken` API will be providing a `function (string $rawToken) => Token | ErrorParsing | ErrorValidating` (or an exception instead of return union type)
 * discussed key rotation and key rings / key chains - those are interesting, but not part of this simplified API
 * signing/reading should be simple enough to do with this object:

```php
interface Key {}

interface SymmetricKey implements Key {}
interface KeyPair implements Key {}

function makeMeJwtThingy(KeySet $key): JwtThingy {}

$token = makeMeJwtThingy(RsaKeypair::fromFiles($secret, $public))
     ->read($_POST['token'], ...$moreValidators);

$token = makeMeJwtThingy(RsaKeypair::fromFiles($secret, $public))
     ->write(['foo' => 'bar'], ExpirationTime::seconds(3600));
```

 * we also discussed adding the current algorithm to the `Key` interface (perhaps sub-interface to avoid BC break), so that we can construct the appropriate signer / validator with the right algos upfront
 * we discussed `Key` ergonomics: to have constructors that throw exceptions in case of malformed keys, or keys that don't match their respective algos (future, not relevant now)

lcobucci:
We also agreed to close this one and keep the discussion on #140.

I'll be providing a draft PR with the unimplemented API and @Slamdunk will tackle the implementation side.

PR #682, opened 2021-02-17
A very wip kind of thing

lcobucci:


src/Configuration.php
```diff
@@ -111,6 +116,37 @@ public function parser(): Parser
         return $this->parser;
     }
 
+    public function secureParser(Constraint ...$extraConstraints): ValidateDuringParsing
+    {
+        $constraints           = array_merge($this->validationConstraints, $extraConstraints);
+        $timeVerification      = null;
+        $signatureVerification = null;
+
+        foreach ($constraints as $i => $constraint) {
```
lcobucci, line 125: This is horrible, honestly, but I don't want to run duplicated signature/time verifications

src/Token/ValidateDuringParsing.php
```diff
@@ -0,0 +1,56 @@
...
+
+    public function parse(string $jwt): Token
+    {
+        $token = $this->decorated->parse($jwt);
+
+        $constraints = [
+            $this->timeVerification,
+            ...$this->extraConstraints,
+            $this->signatureVerification,
+        ];
```
lcobucci, line 48: Ensures the "heavy" validation is done only at the end, saving time if the token is expired or if other constraints fail

PR #684, opened 2021-02-19
Use right fallback method.

pkarppi:
Use right fallback method.


Slamdunk:
Gosh, can't believe it: there are only TWO lines of code untested and I've made a bug in there, so sorry.

The target branch should be `4.1.x`: I've changed it here but you haven't created a separate branch, so rebase is not feasable.

lcobucci:
@pkarppi @Slamdunk thanks for your contribution! New version was just released 👍

PR #699, opened 2021-03-03
Add SignedWithOneOf constraint

fabiante:
This constraint allows to validate a token against multiple public keys

### DIFF
--- a/src/Validation/Constraint/SignedWithOneOf.php
+++ b/src/Validation/Constraint/SignedWithOneOf.php
@@ -1,0 +1,55 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Validation\Constraint;
+
+use Lcobucci\JWT\Signer;
+use Lcobucci\JWT\Token;
+use Lcobucci\JWT\UnencryptedToken;
+use Lcobucci\JWT\Validation\Constraint;
+use Lcobucci\JWT\Validation\ConstraintViolation;
+
+final class SignedWithOneOf implements Constraint
+{
+    private Signer $signer;
+
+    /**
+     * @var Signer\Key[]
+     */
+    private array $keys;
+
+    /**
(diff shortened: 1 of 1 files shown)


fabiante:
@lcobucci Hi! I am currently integration OAuth into a PHP backend and noticed that in order to support key rotation, I would need to test a token against multiple public keys. For that I created this constraint that may or may not be usefull in the base library.

I admit there is also a way to implement without a new constraint, instead checking the `kid` of a token and then validating against the appropriate key. But maybe this constraint is still usefull?

Edit: I will of course fix test failures and add documentation in case this is merge-worthy 😉

Ocramius:
> For that I created this constraint that may or may not be usefull in the base library.

Hey @FabianTe - the important question is whether this is part of the spec, or whether it is a detail that exists only within OAuth?

fabiante:
Thank you @Ocramius for your critical comment. While answering your comment I had a look at the OpenID Connect specification. There I read the following about [rotating signing keys](https://openid.net/specs/openid-connect-core-1_0.html#RotateSigKeys):

> Rotation of signing keys can be accomplished with the following approach. The signer publishes its keys in a JWK Set at its jwks_uri location and includes the kid of the signing key in the JOSE Header of each message to indicate to the verifier which key is to be used to validate the signature. Keys can be rolled over by periodically adding new keys to the JWK Set at the jwks_uri location. The signer can begin using a new key at its discretion and signals the change to the verifier using the kid value. The verifier knows to go back to the jwks_uri location to re-retrieve the keys when it sees an unfamiliar kid value. The JWK Set document at the jwks_uri SHOULD retain recently decommissioned signing keys for a reasonable period of time to facilitate a smooth transition.

I have to admit, that I unconciously implemented this constraint to solve a problem, that would be solved by using the `kid` property in token headers.

Therefore this should not be merged I think 😉

Issue #702, opened 2021-03-14
[BUG] Datetime should be a float instead of string


yassinrais:
Generated by the method ->toString()

```
eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJodHRwOi8vMTI3LjAuMC4xOjgwMDAiLCJhdWQiOiJodHRwOi8vMTI3LjAuMC4xOjgwMDAiLCJpYXQiOiIxNjE1NzUwODQzLjk2Mzg4MCIsImp0aSI6ImYyMjYwYjZjNTM5MTk5YjFmZGM1YmFjODFlNWIxNDNhIiwiZXhwIjoiMTYxNzA0MzI0My45NjM4ODAiLCJ1aWQiOiI1MDA5OTFjMS1lMWNhLTQ5ODUtYjBiMy1mNWNiYmRhNmU5MGIifQ._FgRzQFb4wtXk7zDoQofZH30KyhYx2jKOtyWh-cQtjw
```
[IMAGE] decoded by official JWT website
```json
{
  "iss": "http://127.0.0.1:8000",
  "aud": "http://127.0.0.1:8000",
  "iat": "1615750843.963880",
  "jti": "f2260b6c539199b1fdc5bac81e5b143a",
  "exp": "1617043243.963880",
  "uid": "500991c1-e1ca-4985-b0b3-f5cbbda6e90b"
}
```
We should expect
```json
{
  "iss": "http://127.0.0.1:8000",
  "aud": "http://127.0.0.1:8000",
  "iat": 1615750843.963880,
  "jti": "f2260b6c539199b1fdc5bac81e5b143a",
  "exp": 1617043243.963880,
  "uid": "500991c1-e1ca-4985-b0b3-f5cbbda6e90b"
}
```
Instead of the previus one

yassinrais:
duplicate of #618

zerkms:
This should be reopened: the bug is there and the referred PR is closed.

yassinrais:
we are working in a possible solution #706

lcobucci:
Hello folks, thanks to @yassinrais we managed to [fix this compatibility issue whilst still preserving the microseconds precision](https://github.com/lcobucci/jwt/pull/706). We already released new versions to solve the problem.

PR #706, opened 2021-03-19
Fix usage of non JSON numeric values for time fractions (without having precision issues)

yassinrais:
🖋 The problem of precession of timestamp floats is solved by using a json_encode instead of (string).

This solves the **float round** problem when we encode/parse floats .

### DIFF
--- a/src/Encoding/MicrosecondBasedDateConversion.php
+++ b/src/Encoding/MicrosecondBasedDateConversion.php
@@ -25,16 +25,13 @@ final class MicrosecondBasedDateConversion implements ClaimsFormatter
         return $claims;
     }
 
-    /** @return int|string */
+    /** @return int|float */
     private function convertDate(DateTimeImmutable $date)
     {
-        $seconds      = $date->format('U');
-        $microseconds = $date->format('u');
-
-        if ((int) $microseconds === 0) {
-            return (int) $seconds;
+        if ($date->format('u') === '000000') {
+            return (int) $date->format('U');
         }
 
-        return $seconds . '.' . $microseconds;
+        return (float) $date->format('U.u');
     }
 }
--- a/src/Token/Parser.php
+++ b/src/Token/Parser.php
@@ -12,7 +12,11 @@ use function array_key_exists;
 use function count;
 use function explode;
 use function is_array;
+use function is_string;
+use function json_encode;
 use function strpos;
+
+use const JSON_THROW_ON_ERROR;
 
 final class Parser implements ParserInterface
 {
@@ -105,7 +109,9 @@ final class Parser implements ParserInterface
                 continue;
             }
 
-            $claims[$claim] = $this->convertDate((string) $claims[$claim]);
+            $date = $claims[$claim];
+
+            $claims[$claim] = $this->convertDate(is_string($date) ? $date : json_encode($date, JSON_THROW_ON_ERROR));
         }
 
         return $claims;
--- a/test/unit/Encoding/MicrosecondBasedDateConversionTest.php
+++ b/test/unit/Encoding/MicrosecondBasedDateConversionTest.php
@@ -36,8 +36,8 @@ final class MicrosecondBasedDateConversionTest extends TestCase
         $formatted = $formatter->formatClaims($claims);
 
         self::assertSame(1487285080, $formatted[RegisteredClaims::ISSUED_AT]);
-        self::assertSame('1487285080.000123', $formatted[RegisteredClaims::NOT_BEFORE]);
-        self::assertSame('1487285080.123456', $formatted[RegisteredClaims::EXPIRATION_TIME]);
+        self::assertSame(1487285080.000123, $formatted[RegisteredClaims::NOT_BEFORE]);
+        self::assertSame(1487285080.123456, $formatted[RegisteredClaims::EXPIRATION_TIME]);
         self::assertSame('test', $formatted['testing']); // this should remain untouched
     }
 
@@ -62,7 +62,7 @@ final class MicrosecondBasedDateConversionTest extends TestCase
         $formatted = $formatter->formatClaims($claims);
 
         self::assertSame(1487285080, $formatted[RegisteredClaims::ISSUED_AT]);
-        self::assertSame('1487285080.123456', $formatted[RegisteredClaims::EXPIRATION_TIME]);
+        self::assertSame(1487285080.123456, $formatted[RegisteredClaims::EXPIRATION_TIME]);
         self::assertSame('test', $formatted['testing']); // this should remain untouched
     }
 }
--- a/test/functional/TimeFractionPrecisionTest.php
+++ b/test/functional/TimeFractionPrecisionTest.php
@@ -1,0 +1,61 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\FunctionalTests;
+
+use DateTimeImmutable;
+use Lcobucci\JWT\Configuration;
+use Lcobucci\JWT\Token\Plain;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers \Lcobucci\JWT\Configuration
+ * @covers \Lcobucci\JWT\Encoding\JoseEncoder
+ * @covers \Lcobucci\JWT\Encoding\ChainedFormatter
+ * @covers \Lcobucci\JWT\Encoding\MicrosecondBasedDateConversion
+ * @covers \Lcobucci\JWT\Encoding\UnifyAudience
+ * @covers \Lcobucci\JWT\Token\Builder
+ * @covers \Lcobucci\JWT\Token\Parser
+ * @covers \Lcobucci\JWT\Token\Plain
+ * @covers \Lcobucci\JWT\Token\DataSet
+ * @covers \Lcobucci\JWT\Token\Signature
+ * @covers \Lcobucci\JWT\Signer\Key\InMemory
+ * @covers \Lcobucci\JWT\Signer\None
+ * @covers \Lcobucci\JWT\Validation\Validator
+ * @covers \Lcobucci\JWT\Validation\RequiredConstraintsViolated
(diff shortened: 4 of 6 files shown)


yassinrais (commented):
Only int and float are accepted as returns from a converted date.

[RFC7519 Page 6 : NumericDate](https://tools.ietf.org/html/rfc7519#page-6)

src/Encoding/MicrosecondBasedDateConversion.php
```diff
@@ -25,16 +25,16 @@ public function formatClaims(array $claims): array
...
+        $timestamp    = $date->format('U.u');
         $microseconds = $date->format('u');
 
         if ((int) $microseconds === 0) {
-            return (int) $seconds;
+            return (int) $timestamp;
         }
 
-        return $seconds . '.' . $microseconds;
+        return (float) $timestamp;
```
zerkms: I believe this whole function can be replaced with just one statement: `return (float)$date->format('U.u');`

yassinrais: Yes true, but i was thinking that they will be a cases when we compare int with float , this can cause an issue , im not totaly sur yet , but i will check

test/unit/Encoding/MicrosecondBasedDateConversionTest.php
```diff
@@ -36,8 +36,8 @@ public function dateClaimsHaveMicrosecondsOrSeconds(): void
         $formatted = $formatter->formatClaims($claims);
 
         self::assertSame(1487285080, $formatted[RegisteredClaims::ISSUED_AT]);
-        self::assertSame('1487285080.000123', $formatted[RegisteredClaims::NOT_BEFORE]);
-        self::assertSame('1487285080.123456', $formatted[RegisteredClaims::EXPIRATION_TIME]);
+        self::assertSame(1487285080.000123, $formatted[RegisteredClaims::NOT_BEFORE]);
```
zerkms, line 39: All these tests are wrong: `1487285080.000123` this literal cannot be represented precisely in memory as a floating point number, so you're comparing 2 imprecise values.

You cannot "fix" the problem in the userland, because that's how the language and IEEE754 works.

zerkms:
After reading all other discussions (and code for this PR) I have an impression there is no understanding on what's happening:

php's float64 (and literally **every other's** programming language implementation of IEEE754) fails to represent precisely more than 53 bits of significand precision.

This leads to "rounding" problem (which in fact is just lack of precision in the float64 ieee754 type, nothing really "rounds").

You cannot solve it in userland using built-in numeric types only.

What **MUST** be done though: date fields **MUST** be encoded as numbers, it should not even be an option, it's **REQUIRED** by the spec. Ideally, all changes that led to making the generated jwt non-complaint should be reverted, because there is no way or need to "fix" it, since it only broke everything even "worse".

What can be done if one wants precision over 53 bits of significand precision they need:

1. Bring a custom type that allows it (no built in php types do)
2. Implement json de-/serialisation algorithm that would employ that type (`json_decode`/`json_encode` cannot do that)
3. A custom type for datetime should be implemented (the built-in php data types suffer from the same "problem" as every other numeric php type)

And once again: just shuffling `(float)` into random places will not change anything.

> This solves the float round problem when we encode/parse floats .

The problem with that "problem" is that it never was _formalised_: some arbitrary values were chosen as a "proof" of having the "issue" and then the "solution" was optimised to fit only that limited problem. If the library wants to solve the problem with "rounding" floats - then it should be first defined to what point. At this point it's not obvious why losing 18th significant digit is worse than losing 15th one.

yassinrais:
> After reading all other discussions (and code for this PR) I have an impression there is no understanding on what's happening

If you read my old PR #702, the goal was to keep the 6 decimal digits same from a timestamp after encoding/decoding Json. If you think my solution has a problem or won't work, can you show me a case or example and explain why?

> This leads to "rounding" problem (which in fact is just lack of precision in the float64 ieee754 type, nothing really "rounds").

Im talking only about the 6 digits of timestamp and not talking about *floats* 

>  At this point it's not obvious why losing 18th significant digit is worse than losing 15th one.

Can you please explain what you mean by this?

Also what do you belive that can be changed to fix this "**the bug is there and the referred PR is closed**"

src/Encoding/MicrosecondBasedDateConversion.php
lcobucci: That's correct, it simplifies everything.

src/Token/Parser.php
```diff
@@ -105,7 +107,19 @@ private function parseClaims(string $data): array
...
+                $encoded = json_encode($claimvalue);
+
+                if ($encoded !== false) {
+                    $claims[$claim] = $this->convertDate($encoded);
+
+                    continue;
+                }
+            }
+
+            $claims[$claim] = $this->convertDate((string) $claimvalue);
```
lcobucci: We shouldn't have issues encoding this again to json, so this can be simplified to:

```suggestion
            $date = $claims[$claim];

            $claims[$claim] = $this->convertDate(is_string($date) ? $date : json_encode($date, JSON_THROW_ON_ERROR));
```

lcobucci (changes requested):
We're missing the test that guarantees that precision of time fractions is kept after the parsing - using that `json_encode()` logic.

Without that we're just asserting that the claims are now using floats - which doesn't address the concern we had before.

I'll add the test and rebase your branch. Either we prove that your approach works or that @zerkms is correct.

yassinrais:
>I'll add the test and rebase your branch. Either we prove that your approach works or that @zerkms is correct.

Looking forward to hearing from you 👌.

lcobucci:
@zerkms do you think I'm missing something on https://github.com/lcobucci/jwt/pull/706/commits/484430f808db5dd907537e9f671ef1c3bd890ab3? Just to make it 100% transparent the scope of this is to make sure the precision up to milliseconds is respected.

lcobucci:
@Ocramius would you also have a look at this, please?

src/Token/Parser.php
```diff
@@ -105,7 +109,9 @@ private function parseClaims(string $data): array
                 continue;
             }
 
-            $claims[$claim] = $this->convertDate((string) $claims[$claim]);
+            $date = $claims[$claim];
+
+            $claims[$claim] = $this->convertDate(is_string($date) ? $date : json_encode($date, JSON_THROW_ON_ERROR));
```
lcobucci, line 114: To make this explicit as well, the check is only needed for compatibility with previously issued tokens...

test/unit/Encoding/MicrosecondBasedDateConversionTest.php
```diff
@@ -35,17 +34,16 @@ public function dateClaimsHaveMicrosecondsOrSeconds(): void
         $formatter = new MicrosecondBasedDateConversion();
         $formatted = $formatter->formatClaims($claims);
 
-        self::assertSame(1487285080, $formatted[RegisteredClaims::ISSUED_AT]);
-        self::assertSame('1487285080.000123', $formatted[RegisteredClaims::NOT_BEFORE]);
-        self::assertSame('1487285080.123456', $formatted[RegisteredClaims::EXPIRATION_TIME]);
+        self::assertSame(1487285080.0, $formatted[RegisteredClaims::ISSUED_AT]);
```
Slamdunk: This is a BC break for those who are relying right now to strict check to integers, isn't it?

Slamdunk (commented):
Awesome solution, but I think we should keep integer type for integer timestamps

yassinrais (commented):
> Awesome solution, but I think we should keep integer type for integer timestamps

Yes, i remember now , That's why I left the integers and floats as returns. , because if you count, many use this library will got some problems about existing tokens that contains only integers. 👏🏻

test/unit/Encoding/MicrosecondBasedDateConversionTest.php
lcobucci: The interface uses DateTimeImmutable. The BC-break is with the encoded token (which, as reported many times, shouldn't never be a string).

lcobucci: We can keep the int, not an issue

lcobucci:
I'll handle that int stuff. Thanks for the review @Slamdunk

zerkms:
> If you read my old PR #702, the goal was to keep the 6 decimal digits same from a timestamp after encoding/decoding Json. If you think my solution has a problem or won't work, can you show me a case or example and explain why?

Your implementation would work, I just thought this: https://github.com/lcobucci/jwt/pull/618#issuecomment-782922028 was a problem.

> Im talking only about the 6 digits of timestamp and not talking about floats

Timestamp is encoded as a float.

Ocramius:
@lcobucci I was kinda glancing over this earlier today, and I don't see
massive problems with the patch itself.

It seems fine code-wise, but I don't understand the rationale behind it:
mostly lacking reference to issues connected to this.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

lcobucci:
@Ocramius I've tried to explain the issue on the commit message: https://github.com/lcobucci/jwt/pull/706/commits/9a961f4541ba13ac27dcebaaa222ad8e342ce14b does it give you more info?

> Your implementation would work, I just thought this: #618 (comment) was a problem.

It is only a problem when casting to strings, that was @yassinrais' finding.

lcobucci (approved):

lcobucci:
I've applied the changes and extracted the test to a separate place, would you all give a final check? Even though this is a minor BC-break, it's solving a compatibility issue that has been introduced since v4.0 and we should ship it as a patch.

Slamdunk (approved):

lcobucci:
@yassinrais once again, thank you very much for your help in finding the fix :+1:

yassinrais:
@lcobucci  you're welcome ! 
I use it so I'm supposed to help 😁👌
It's great to see the progress made  🚀.

PR #710, opened 2021-03-23
Prevent type conversions issues when parsing time-fractions

### LINKED ISSUE #709: Usage of `json_encode()` to convert time fractions to microseconds cause precision issues
message: "Value is not in the allowed date format: 1616481863.528781890869140625"

lcobucci:
We received reports of users experiencing exceptions when parsing JWTs with timestamps with precision beyond microseconds on time-fractions.

This guards against having any kind of edge-case when dealing with floats as input to the creation of DateTimeImmutable objects.

### DIFF
--- a/src/Token/Parser.php
+++ b/src/Token/Parser.php
@@ -12,14 +12,13 @@ use function array_key_exists;
 use function count;
 use function explode;
 use function is_array;
-use function is_string;
-use function json_encode;
-use function strpos;
-
-use const JSON_THROW_ON_ERROR;
+use function is_numeric;
+use function number_format;
 
 final class Parser implements ParserInterface
 {
+    private const MICROSECOND_PRECISION = 6;
+
     private Decoder $decoder;
 
     public function __construct(Decoder $decoder)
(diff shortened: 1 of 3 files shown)


yassinrais (commented):
src/Token/Parser.php:118
```php
     * @param int|float|string $timestamp
```

Are you sur that its okey to put even string into param type ? 

@param => accepted type in _'documentation'_

lcobucci:
> src/Token/Parser.php:118
> 
> ```
>      * @param int|float|string $timestamp
> ```
> 
> Are you sur that its okey to put even string into param type ?
> 
> @param => accepted type in _'documentation'_

Not accepting it is a BC-break with previously created tokens in v4. So, we have to support until v5 - that's also why I added a test to cover that.

We can mark that as deprecated but it's not really a hill I'd die on.

yassinrais (commented):
~some bad news : https://3v4l.org/pTqmM 🤔 wait a minute~ nvm, i forgot the was u.u 😅 and number_format add .u format to integer

yassinrais:
I dont have anything to add 💪 gj

PR #725, opened 2021-04-27
[docs] Clarify change in date-formats

jaylinski:
I recently upgraded a project to `lcobucci/jwt:4.1`. My issue was that tokens generated by v4 were rejected by the v3 version of this library, because the `iat`-claim validation always failed.

I could solve this issue by using the method described in this PR.

I'm not sure what caused this issue, but I guess it was leeway-related.

Related to #710

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -131,7 +131,7 @@ There are 4 main differences on the new API:
 
 1. Token configuration methods were renamed
 1. Signature is created via `Builder#getToken()` (instead of `Builder#sign()`)
-1. `DateTimeImmutable` objects are now for the registered claims with dates
+1. `DateTimeImmutable` objects are now used for the registered claims with dates, which are by default encoded as floats with microseconds as precision
 1. Headers should be replicated manually - whenever necessary
 
 Here's the migration:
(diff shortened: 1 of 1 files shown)


docs/upgrading.md
```diff
@@ -131,7 +131,7 @@ There are 4 main differences on the new API:
 
 1. Token configuration methods were renamed
 1. Signature is created via `Builder#getToken()` (instead of `Builder#sign()`)
-1. `DateTimeImmutable` objects are now for the registered claims with dates
+1. `DateTimeImmutable` objects are now used for the registered claims with dates and will use microtime instead of unix timestamps
```
lcobucci: I like the explicitness here, just think that we can improve the construction. What do you think about this:

```suggestion
1. `DateTimeImmutable` objects are now used for the registered claims with dates, which are by default encoded as floats with microseconds as precision
```

lcobucci (changes requested):
Just a teeny tiny thing.

Thanks for your contribution @jaylinski!

docs/upgrading.md
jaylinski: I think it is much better. 👍 I pushed the changes.

lcobucci (approved):

PR #758, opened 2021-06-10
Token\Parser: specific covariant return type

Slamdunk:
This is not a BC-Break because:

* Return type covariance was introduced in PHP 7.4, which is the minimum version supported for `v4`
* The class is final

### DIFF
--- a/src/Token/Parser.php
+++ b/src/Token/Parser.php
@@ -6,7 +6,7 @@ namespace Lcobucci\JWT\Token;
 use DateTimeImmutable;
 use Lcobucci\JWT\Decoder;
 use Lcobucci\JWT\Parser as ParserInterface;
-use Lcobucci\JWT\Token as TokenInterface;
+use Lcobucci\JWT\UnencryptedToken;
 
 use function array_key_exists;
 use function count;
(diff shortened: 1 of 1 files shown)


lcobucci:
What's your motivation here? There was an intention behind the design, which is being disregarded with this change.

Slamdunk:
To ease IDEs and static analysis tools.

You would argue that `Token\Parser` is intended to return a `TokenInterface` (just like its interface) so in the future we can return `EncryptedToken` too: this PR is fine with that intention as this can be reverted without being a BC Break.
But as long as `return new Plain()` is hardcoded, I see no point in tiring users to add `assert($token instanceof UnencryptedToken);` everywhere.

lcobucci:
Users are likely to depend on the interface rather than the implementation, so that will either get unnoticed or will force people to have assertions to check which parser implementation is being used.

I know the docs show a call to `assert()` but, considering that the unparsed token can be categorised as user input, I believe people should do a validation that runs on production environments.

We also can't go back because that would violate the expectations of people didn't do the check to see if the token was plain or encrypted.

What do you think about leaving that concern to the high level API we discussed a few weeks back?

Slamdunk:
> Users are likely to depend on the interface rather than the implementation, so that will either get unnoticed or will force people to have assertions to check which parser implementation is being used.

Indeed this PR was intended only for those (few) who rely on the implementation.

> We also can't go back because that would violate the expectations of people didn't do the check to see if the token was plain or encrypted.

To me expectations are not API: the only contract `Token\Parser` has is to implement `Parser` interface, as long as it does it, it's backward compatible as far as I can tell.

> What do you think about leaving that concern to the high level API we discussed a few weeks back?

As you wish

PR #759, opened 2021-06-10
Add simplified API

### LINKED ISSUE #140: Provide a high level API to simplify usage

### LINKED ISSUE #663: Misuse resistant API

Slamdunk:
Close #140 
Close #663
See https://github.com/lcobucci/jwt/pull/680#issuecomment-847297373

The goal is to provide a simple API that addresses 99% of use cases securely.

* No interface by design, it should never be mocked
* Only hardcoded dependencies by design
* Tests must reflect API purpose: one-liner call to fulfill this library scope 

Documentation will be updated here after implementation review.


src/SimpleParser.php
```diff
@@ -0,0 +1,33 @@
...
+
+use Lcobucci\JWT\Encoding\JoseEncoder;
+use Lcobucci\JWT\Token as TokenInterface;
+use Lcobucci\JWT\Token\Parser;
+use Lcobucci\JWT\Validation\Constraint;
+use Lcobucci\JWT\Validation\SignedWith;
+use Lcobucci\JWT\Validation\ValidAt;
+use Lcobucci\JWT\Validation\Validator;
+
+final class SimpleParser
```
lcobucci: I don't have a good naming suggestion now but it would be nice to avoid using `Simple` here... I think we also talked about having a method to build tokens, with a signature like `function (array $claims): string` (where claims is `array<string, mixed>`.

src/SimpleParser.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    public function parseJwt(
+        string $jwt,
+        SignedWith $signedWith,
+        ValidAt $validAt,
+        Constraint ...$constraints
+    ): TokenInterface {
```
lcobucci: This is a good opportunity to introduce the `UnencryptedToken`, granted we also validate that. We can start with an `assert()` call and evolve it into a `Constraint` in the future (as anonymous class would work fine, even).

test/unit/SimpleParserTest.php
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * @test
+     *
+     * @covers ::parseJwt
+     */
+    public function badSigner(): void
+    {
+        $this->expectException(RequiredConstraintsViolated::class);
```
lcobucci: By only checking the exception type we aren't able to determine which constraint fails (which causes the reported escaped mutation). We should also validate the message to remove the mutation... it kind of falls into the scope since the component is forwarding the args.

src/SimpleParser.php
```diff
@@ -0,0 +1,33 @@
...
+        $token = (new Parser(new JoseEncoder()))->parse($jwt);
+
+        (new Validator())->assert(
```
lcobucci: Here's an interesting thing to discuss: how much of "internals" do we want to expose in this high-level API? This implementation will throw for all possible scenarios... do we want to let the user know of that or should we return `null` for the "errors"?

If `null` is fine, then we need to put a try/catch around the parser and should probably use `validate()` instead of `assert()`.

lcobucci (commented):
Thanks for kicking-off this ❤️ 

Some notes here 👍

src/SimpleParser.php
Slamdunk: I've updated it to `JwtFacade` so everytime a dev is facing the LoC:

* it's always clear we are dealing with JWT without aliasing the import statement
* it's known it's just a Facade: regular users would be fine, advanced users know they can dig deep into code for further features

Regarding the token creation, see https://github.com/lcobucci/jwt/pull/759#discussion_r649700273

src/JwtFacade.php
```diff
@@ -0,0 +1,45 @@
...
+use Lcobucci\JWT\Validation\Constraint;
+use Lcobucci\JWT\Validation\SignedWith;
+use Lcobucci\JWT\Validation\ValidAt;
+use Lcobucci\JWT\Validation\Validator;
+
+use function assert;
+
+final class JwtFacade
+{
+    public function getBuilder(): Builder
```
Slamdunk: Instead of https://github.com/lcobucci/jwt/pull/759#discussion_r649256312 I've chosen to only have this as token creation help because:

1. If we accept `array<string, mixed> $claims` than we have to map them back to `Builder` interface, which would just be a nightmare
2. `Builder` is already a rock solid interface, easy to deal with within IDEs and cannot be misused. I'll implement soon a `ExpirationRequiredBuilder` just to be sure users calls `expiresAt`
3. It would also result in a one-liner code, just like the test shows

Makefile
```diff
@@ -12,12 +12,12 @@ vendor/composer/installed.json: composer.json composer.lock
...
 phpunit:
-	@vendor/bin/phpunit
+	@php -d zend.assertions=1 vendor/bin/phpunit
 
 .PHONY: infection
 infection:
-	@vendor/bin/phpunit --testsuite=unit --coverage-xml=build/coverage-xml --log-junit=build/junit.xml $(PHPUNIT_FLAGS)
-	@vendor/bin/infection -s --threads=$(PARALLELISM) --coverage=build $(INFECTION_FLAGS)
+	@php -d zend.assertions=1 vendor/bin/phpunit --testsuite=unit --coverage-xml=build/coverage-xml --log-junit=build/junit.xml $(PHPUNIT_FLAGS)
+	@php -d zend.assertions=1 vendor/bin/infection -v -s --threads=$(PARALLELISM) --coverage=build --skip-initial-tests $(INFECTION_FLAGS)
```
Slamdunk, line 20: 1. `zend.assertions=1` mandatory for everything to work correctly now
2. `-v` to have the actual diff for failing mutations
3. `--skip-initial-tests` to speed things up, no need to repeat the exact same tests already run in the previous line

src/JwtFacade.php
Slamdunk: For point [2] have a look at the new commit https://github.com/lcobucci/jwt/pull/759/commits/f539fcc2f63a25fee42b9e059eda0070be34504d

src/SimpleParser.php
Slamdunk: Let's step back for a moment: what are (99% of) the real scenarios this code raises an exception for?

1. The developer is messing things up during development
2. Someone wants to hack into production
3. Production server's clock is messed up

For all the above, I'd like to have the full exception to be thrown.

In regular runs, I have never dealt bad path validation, do you?

lcobucci: My message was not very good, honestly... what I meant is that I don't think that we need to expose details on what were the failures here. Using `Validator#validate()` interrupts the validation on the first issue, which is what I'd recommend for production environments (I'd also swap the `$validAt` and `$signedWith` so that date validation happens before).

src/JwtFacade.php
```diff
@@ -4,23 +4,26 @@
...
     public function parse(
         string $jwt,
         SignedWith $signedWith,
         ValidAt $validAt,
         Constraint ...$constraints
-    ): TokenInterface {
+    ): UnencryptedToken {
         $token = (new Parser(new JoseEncoder()))->parse($jwt);
 
+        assert($token instanceof UnencryptedToken);
```
lcobucci, line 45: This can't be an assertion, it must be a runtime validation since we're dealing with user input.

src/JwtFacade.php
lcobucci: I'd prefer us containing the impact to only this component.
Instead of using an array of claims, we could use a closure:

```php
/**
 * @param Closure(Builder):Builder $customiseBuilder
 */
public function issueToken(Signer $signer, Key $signingKey, Closure $customiseBuilder): UnencryptedToken
{
    $builder = new Token\Builder(new JoseEncoder(), ChainedFormatter::withUnixTimestampDates());
    $now = new DateTimeImmutable();

    return $customiseBuilder($builder)
        ->issuedAt($now)
        ->canOnlyBeUsedAfter($now)
        ->expiresAt($now->modify('+5 minutes'))
        ->getToken($signer, $signingKey);
}
```

We can surely modify the position of the date claims configuration but I think this combines the best of both worlds. We make sure the time constraints are set and users can rely on the `Builder` interface to perform the customisation (without having to maintain other components).

lcobucci (commented):
@Slamdunk sorry about my delay here... my job is demanding quite of energy lately.

I like how the parsing is done but I do think that we shouldn't need other components than the façade for the issuing.

src/JwtFacade.php
Slamdunk: I've set timestamps before calling the closure, so the wise user can overwrite them - but never unset them :wink:

src/JwtFacade.php
Slamdunk: Ok, but I can't test the runtime exception since the returned value of `(new Parser)->parse()` is fixed.
Should I add the runtime check into infection ignore list too, so?

src/SimpleParser.php
Slamdunk: I have no idea what you are asking for :sweat:

lcobucci: The fundamental difference between `Validator#assert()` and `Validator#validate()` is that the former goes through all the constraints and aggregates the violations (and their messages), while the latter returns `false` at the very failed constraint.

What I'm trying to discuss here is if, for this high-level API, we should go through all constraints when the first one has failed. In my projects, I don't feel that need, ergo my question.

src/JwtFacade.php
lcobucci: What I would like to avoid is having to also modify this component when implementing nested tokens.
However, the alternative solution - relying on optional constructor injection of the interfaces instead of instantiating things on the method - would make things a bit more complicated now.

Let's just accept this trade-off for now...

src/SimpleParser.php
Slamdunk: My use cases tell me the opposite: most of (my) failures come from misconfiguration and server misalignment; in all scenarios, even for the few malicious targeted attempts I underwent, the more information I had the sooner I fixed the production.

Slamdunk:
@lcobucci from our conversation here emerged that such high level API may be missing something we both forget. In order to not need a new major in case we did miss something, what about labeling this class as `@experimental` for now?

lcobucci:
> @lcobucci from our conversation here emerged that such high level API may be missing something we both forget. In order to not need a new major in case we did miss something, what about labeling this class as `@experimental` for now?

That makes total sense :)
Is `@experimental` something standardised, though?

src/SimpleParser.php
lcobucci: Let's leave it like this, then, my use-cases are generally not the default for people 😛 

This does highlight a small oversight on the validation API... we only have the error msg from the constraint but no reference to which constraint caused the violation. I'll have a look at this in another PR.

lcobucci:
We're missing a test case that makes sure multiple violation messages are aggregated, I'll provide that in a separate PR so we can unblock this.

lcobucci:
Thanks @Slamdunk for all the help and sorry (again) about the delays.

Slamdunk:
> That makes total sense :)
> Is `@experimental` something standardized, though?

Not in PHP FIG:
https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md
But Symfony uses it:
https://symfony.com/doc/current/contributing/code/experimental.html

I would use this line:
```php
/**
 * @experimental This class is not covered by the backward compatibility promise, yet
 */
```

lcobucci:
> > That makes total sense :)
> > Is `@experimental` something standardized, though?
> 
> Not in PHP FIG: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md But Symfony uses it: https://symfony.com/doc/current/contributing/code/experimental.html
> 
> I would use this line:
> 
> ```
> /**
>  * @experimental This class is not covered by the backward compatibility promise, yet
>  */
> ```

There's nothing on PHP Documentor as well. Let's avoid a non-standard annotation.

PR #771, opened 2021-07-28
Fix Memory overflow caused by circular reference

canren:
When I use swoole as the server, memory overflow caused by circular reference


lcobucci:
@canren thanks for taking the time to work on this. However, this is not a security issue and v3.x series isn't under active maintenance.

Please upgrade to v4, which solves this and other design issues as well.

lcobucci:
Closing as won't fix as per my comment.

Issue #786, opened 2021-09-22
`claims()` missing in Token interface


Patrick-Remy:
The Token interface of 4.x does not include the `claims()` method, as suggested to use in migration guide https://lcobucci-jwt.readthedocs.io/en/latest/upgrading/#use-the-new-token-api

It only includes the `headers()` method.

https://github.com/lcobucci/jwt/blob/52c43b2ba8aa30efcc3a38b7aa7d0c4f560dfe19/src/Token.php#L11-L14

SvenRtbg:
`claims()` is present in the `UnencryptedToken` interface, and I think it may make sense because you'd probably be unable to access them if the token is encrypted.

But what is your case here? Do you have any code that may illustrate what you want to do?

Patrick-Remy:
My static analysis engine pointed out
```
$token = $jwtConfig->parser()->parse($jwtString);
$foo = $token->claims()->get('bar'); // Undefined method 'claims'
```

As `parse()` method does return just a `Token`.

lcobucci:
@Patrick-Remy your static analysis is reminding you to validate user input.

Parse will potentially return encrypted tokens in the future, so you need to make sure that what you received as literal string is what your application expects.

lcobucci:
Closing as invalid as per my comment

Issue #775, opened 2021-08-06
Unencrypted Token API Not Visible In Parsed Result


XedinUnknown:
## The Problem
[`Parser#parse()`][] [returns][1] an [`UnencryptedToken`][], but only guarantees a [`Token`][]. This makes it impossible to retrieve essential information, such as claims, from the parsed result, because it is not declared in the guaranteed result's interface.

[`Parser#parse()`]: https://github.com/lcobucci/jwt/blob/4.2.x/src/Token/Parser.php#L29
[`Token`]: https://github.com/lcobucci/jwt/blob/4.2.x/src/Token.php
[`UnencryptedToken`]: https://github.com/lcobucci/jwt/blob/4.2.x/src/UnencryptedToken.php
[1]: https://github.com/lcobucci/jwt/blob/4.2.x/src/Token/Parser.php#L35

t0mmy742:
It was already discussed on #186 and #203

> Since parse() deals with untrusted data, users are supposed to verify if they have the expected type of token

You can try something like this:

```PHP
$token = $this->parser->parse($jwt);

if (!$token instanceof UnencryptedToken) {
    throw new \RuntimeException('Invalid token');
}
```

You can use your own exception of course

lcobucci:
@t0mmy742 thanks for the help here :+1:

@XedinUnknown as said above, this is a forward compatibility layer.
It was added to avoid having to break the API when implementing encryption.

#759 is adding a high-level API that will always return an unencrypted token, and we'll soon be shipping it.

XedinUnknown:
@t0mmy742, that's exactly what I did. It's fine, but sub-optimal of course.

@lcobucci, thanks, understood 👍

PR #814, opened 2022-01-20
Add `Parser::parseUnencrypted()`

spawnia:
The following code currently works fine, but triggers a static analysis error due to the loose return type of `Parser::parse()`:

```php
$token = (new Parser(new JoseEncoder()))->parse($jwtString);
$token->claims();

// PHPStan says: Call to an undefined method Lcobucci\JWT\Token::claims().
```

I understand from previous discussion that `Parser::parse()` may also be used to handle encrypted tokens in the future.

Instead of handling other possible parse results in the application - and duplicating proper error handling - I think it is valuable for this library to offer a method that only deals with unencrypted tokens. This way, application code is simplified and proper error handling is ensured. Thus, I propose the following new method:

```php
$token = (new Parser(new JoseEncoder()))->parseUnencrypted($unencryptedJwtString);
assert($token instanceof UnencryptedToken);
```

### DIFF
--- a/src/Token/Parser.php
+++ b/src/Token/Parser.php
@@ -7,10 +7,12 @@ use DateTimeImmutable;
 use Lcobucci\JWT\Decoder;
 use Lcobucci\JWT\Parser as ParserInterface;
 use Lcobucci\JWT\Token as TokenInterface;
+use Lcobucci\JWT\UnencryptedToken;
 
 use function array_key_exists;
 use function count;
 use function explode;
+use function get_class;
 use function is_array;
 use function is_numeric;
 use function number_format;
@@ -26,7 +28,12 @@ final class Parser implements ParserInterface
         $this->decoder = $decoder;
     }
 
-    public function parse(string $jwt): Plain
+    /**
+     * Parse the given JWT string into a structured representation.
+     *
+     * @return TokenInterface an encrypted or unencrypted token
+     */
+    public function parse(string $jwt): TokenInterface
     {
         [$encodedHeaders, $encodedClaims, $encodedSignature] = $this->splitJwt($jwt);
 
@@ -37,6 +44,22 @@ final class Parser implements ParserInterface
             new DataSet($this->parseClaims($encodedClaims), $encodedClaims),
             $this->parseSignature($header, $encodedSignature)
         );
+    }
+
+    /**
+     * Parse the given JWT string into a plain token.
+     *
+     * @throws InvalidTokenStructure if the input unexpectedly decodes to an encrypted token.
+     */
+    public function parseUnencrypted(string $unencrypted): UnencryptedToken
+    {
+        $token = $this->parse($unencrypted);
+
+        if (! ($token instanceof UnencryptedToken)) {
+            throw InvalidTokenStructure::unencryptedTokenExpected(get_class($token));
+        }
+
+        return $token;
     }
 
     /**
@@ -53,6 +76,9 @@ final class Parser implements ParserInterface
         if (count($data) !== 3) {
             throw InvalidTokenStructure::missingOrNotEnoughSeparators();
         }
+
+        // @codingStandardsIgnoreLine Missing variable $data before or after the documentation comment.
+        /** @var array{string, string, string} $data proven by count() */
 
         return $data;
     }
--- a/src/Token/Parser.php
+++ b/src/Token/Parser.php
@@ -26,7 +26,7 @@ final class Parser implements ParserInterface
         $this->decoder = $decoder;
     }
 
-    public function parse(string $jwt): TokenInterface
+    public function parse(string $jwt): Plain
     {
         [$encodedHeaders, $encodedClaims, $encodedSignature] = $this->splitJwt($jwt);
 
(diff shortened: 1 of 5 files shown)


Ocramius (approved):
IMO OK - note that relying on the `Parser` interface still wouldn't give you a `Plain` token

SvenRtbg:
I have the feeling that `parse()` only giving you a `Token` interface is intentional for some future feature, so relying on the one implementation to always returning a `Plain` token, or even enforcing that in the interface may counteract that intention.

Ocramius:
Nah, perfectly valid for an implementation to return a more specific type.

The fact that the parser implementation may return new token types in future would become a BC break, which is also fine to have as an explicit transition.

src/Token/Parser.php
```diff
@@ -26,7 +26,7 @@ public function __construct(Decoder $decoder)
         $this->decoder = $decoder;
     }
 
-    public function parse(string $jwt): TokenInterface
+    public function parse(string $jwt): Plain
```
Slamdunk: This has already been discussed:

1. https://github.com/lcobucci/jwt/pull/758
2. https://github.com/lcobucci/jwt/issues/775
3. https://github.com/lcobucci/jwt/issues/786

spawnia: Thanks for pointing me towards this previous work. I took inspiration from your approach in `JwtFacade::parse()` and extended `Parser` itself with a method that expects to only handle unencrypted tokens.

src/Token/Parser.php
```diff
@@ -39,10 +46,26 @@ public function parse(string $jwt): TokenInterface
         );
     }
 
+    /**
+     * Parse the given JWT string into a plain token.
+     *
+     * @throws InvalidTokenStructure if the input unexpectedly decodes to an encrypted token.
+     */
+    public function parseUnencrypted(string $unencrypted): UnencryptedToken
```
spawnia, line 54: I don't know how to test this, other than repeating the tests for `parse()` - perhaps they should be parameterized with a data provider to test both `parse()` and `parseUnencrypted()`?

src/Token/Parser.php
```diff
@@ -39,10 +46,26 @@ public function parse(string $jwt): TokenInterface
...
+    {
+        $token = $this->parse($unencrypted);
+
+        if (! ($token instanceof UnencryptedToken)) {
```
spawnia, line 58: This will lead to escaped mutants, can this be ignored for infection testing?

lcobucci:
I left the extension point on purpose and would prefer not to remove it or provide different methods to handle this.

JWTs are usually user input and I believe it's responsibility of the users of this library to rely on the interface and verify they have the expected type of token.

The facade is there to offer a simplified API, it's only pending docs and a few more tests (that I couldn't do yet). Doesn't it satisfy your needs?

spawnia:
> JWTs are usually user input and I believe it's responsibility of the users of this library to rely on the interface and verify they have the expected type of token.

I can follow your reasoning, but I want to point out that it is somewhat arbitrary where the line is drawn. The following are all examples of a client not parsing a proper JWT:
- JWT is a string and not null: client must check or get a type error
- JWT has three dots: parser checks and throws if not
- JWT has some required claims, standard claims follow structure: parser checks and throws if not
- JWT is signed with a given key: client may check by invoking `SignedWith`
- JWT is unencrypted: ?

Calling `Parser::parse()` can be thought of as asking: "I have a string that I believe to be a JWT, please decode it and confirm that assumption to me.". The added `Parser::parseUnencrypted()` is asking a similar thing, with the added stipulation that the string is believed to be an unencrypted JWT string. I don't see a fundamental difference between those, and believe they are equally valid. Given that the library only supports unencrypted JWT anyway, it would be great if that use case would be made as simple as possible and require little extra code.

> The facade is there to offer a simplified API, it's only pending docs and a few more tests (that I couldn't do yet).

The facade is doing something quite similar to this proposed API: it combines multiple assumptions about the passed in JWT. However, it does so at a more coarse level - requiring two additional constraints. I think this shows that there is a spectrum of how much constraints the library methods validate, and I believe this pull request is closing a gap between the currently existing two extremes.

> Doesn't it satisfy your needs?

Due to reasons outside of my immediate control, we are currently forced to forgo time based validation of the tokens we handle.

Would you be more comfortable if I were to add this method as `JwtFacade::parseUnencrypted()`?

lcobucci:
@spawnia sorry about my lack of response here (`E_TOO_MANY_THINGS_HAPPENING_AT_THE_SAME` :rofl:).

> Due to reasons outside of my immediate control, we are currently forced to forgo time based validation of the tokens we handle.

Then you could still rely on it but provide your own implementation of a `ValidAt` constraint that doesn't validate anything, right? IMHO that makes your specific use case explicit, whilst allowing for people to challenge that in the future.

I've posted on #822 some more thoughts on this encrypted/unencrypted thing. It would be nice to have your PoV there =)

lcobucci:
Closing here for now as that's not the direction I'd like us to take.

The main "conflict point" is that in my understanding of the [RFC](https://www.rfc-editor.org/rfc/rfc7519#appendix-A) an encrypted/nested JWT is still a JWT, therefore parsing should not provide guarantees but support both types.

PR #827, opened 2022-03-30
Add constraint for private claim validation

### LINKED ISSUE #826: Add withClaim validation for custom claim validation
OWASP has a [list of things to improve security in JWT](https://cheatsheetseries.owasp.org/cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.html#token-sidejacking). One of those things regarding token sidejacking is validating custom claims against a random fingerprint. It would be great if this library provided a way to compare custom claims and values and validate the claim exists and the value is correct.

james-bw:
Re: https://github.com/lcobucci/jwt/issues/826

### DIFF
--- a/src/Validation/Constraint/HasClaimWithValue.php
+++ b/src/Validation/Constraint/HasClaimWithValue.php
@@ -1,0 +1,47 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Validation\Constraint;
+
+use Lcobucci\JWT\Token;
+use Lcobucci\JWT\UnencryptedToken;
+use Lcobucci\JWT\Validation\Constraint;
+use Lcobucci\JWT\Validation\ConstraintViolation;
+
+use function in_array;
+
+final class HasClaimWithValue implements Constraint
+{
+    private string $claim;
+
+    /** @var mixed */
+    private $expectedValue;
+
+    /** @param mixed $expectedValue */
+    public function __construct(string $claim, $expectedValue)
+    {
+        if (in_array($claim, Token\RegisteredClaims::ALL, true)) {
+            throw CannotValidateARegisteredClaim::create($claim);
+        }
+
+        $this->claim         = $claim;
+        $this->expectedValue = $expectedValue;
+    }
+
+    public function assert(Token $token): void
+    {
+        if (! $token instanceof UnencryptedToken) {
+            throw new ConstraintViolation('You should pass a plain token');
+        }
+
+        $claims = $token->claims();
+
+        if (! $claims->has($this->claim)) {
+            throw new ConstraintViolation('The token does not have the claim "' . $this->claim . '"');
+        }
+
+        if ($claims->get($this->claim) !== $this->expectedValue) {
+            throw new ConstraintViolation('The claim "' . $this->claim . '" does not have the expected value');
+        }
+    }
+}
--- a/docs/validating-tokens.md
+++ b/docs/validating-tokens.md
@@ -75,6 +75,7 @@ This library provides the following constraints:
 * `Lcobucci\JWT\Validation\Constraint\SignedWith`: verifies if the token was signed with the expected signer and key
 * `Lcobucci\JWT\Validation\Constraint\StrictValidAt`: verifies presence and validity of the claims `iat`, `nbf`, and `exp` (supports leeway configuration)
 * `Lcobucci\JWT\Validation\Constraint\LooseValidAt`: verifies the claims `iat`, `nbf`, and `exp`, when present (supports leeway configuration)
+* `Lcobucci\JWT\Validation\Constraint\HasClaimWithValue`: verifies that a custom claim has the expected value (not recommended when comparing cryptographic hashes)
 
 Example code for adding a constraint to the configuration object:
 
--- a/src/Validation/Constraint/CannotValidateARegisteredClaim.php
+++ b/src/Validation/Constraint/CannotValidateARegisteredClaim.php
@@ -1,0 +1,17 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Validation\Constraint;
+
+use InvalidArgumentException;
+use Lcobucci\JWT\Exception;
+
+final class CannotValidateARegisteredClaim extends InvalidArgumentException implements Exception
+{
+    public static function create(string $claim): self
+    {
+        return new self(
+            'The claim "' . $claim . '" is a registered claim, another constraint must be used to validate its value'
+        );
+    }
+}
--- a/test/unit/Validation/Constraint/HasClaimWithValueTest.php
+++ b/test/unit/Validation/Constraint/HasClaimWithValueTest.php
@@ -1,0 +1,112 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Validation\Constraint;
+
+use Lcobucci\JWT\Token;
+use Lcobucci\JWT\Validation\ConstraintViolation;
+
+/** @coversDefaultClass \Lcobucci\JWT\Validation\Constraint\HasClaimWithValue */
+final class HasClaimWithValueTest extends ConstraintTestCase
+{
+    /**
+     * @test
+     * @dataProvider registeredClaims
+     *
+     * @covers ::__construct
+     * @covers \Lcobucci\JWT\Validation\Constraint\CannotValidateARegisteredClaim
+     */
+    public function registeredClaimsCannotBeValidatedUsingThisConstraint(string $claim): void
+    {
+        $this->expectException(CannotValidateARegisteredClaim::class);
+        $this->expectExceptionMessage(
+            'The claim "' . $claim . '" is a registered claim, another constraint must be used to validate its value'
+        );
+
+        new HasClaimWithValue($claim, 'testing');
+    }
+
+    /** @return iterable<string, array{string}> */
+    public function registeredClaims(): iterable
+    {
+        foreach (Token\RegisteredClaims::ALL as $claim) {
+            yield $claim => [$claim];
+        }
+    }
+
+    /**
+     * @test
+     *
+     * @covers ::__construct
+     * @covers ::assert
+     *
+     * @uses \Lcobucci\JWT\Token\DataSet
+     * @uses \Lcobucci\JWT\Token\Plain
+     * @uses \Lcobucci\JWT\Token\Signature
+     */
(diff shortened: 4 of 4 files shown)


src/Token.php
```diff
@@ -48,6 +48,11 @@ public function isMinimumTimeBefore(DateTimeInterface $now): bool;
      */
     public function isExpired(DateTimeInterface $now): bool;
 
+    /**
+     * Returns if the token has the claim and the value matches
+     */
+    public function withClaim(string $claimId, string $claimValue): bool;
```
Slamdunk:
```suggestion
    public function hasClaimWithValue(string $claimId, string $claimValue): bool;
```
Not exactly sure what's the best name for this method, but `with` has been used by PSR to convey a factory method: https://www.php-fig.org/psr/psr-7/

Slamdunk (dismissed):

src/Token.php
james-bw: Of course. I'll change both token and constraint.

Slamdunk:
@Ocramius https://github.com/lcobucci/jwt/runs/5749877948?check_suite_focus=true#step:4:250
```
[BC] ADDED: Method hasClaimWithValue() was added to interface Lcobucci\JWT\UnencryptedToken
[BC] ADDED: Method hasClaimWithValue() was added to interface Lcobucci\JWT\Token
2 backwards-incompatible changes detected
```
Why **adding** a method to an interface should be considered a BC Break?

SvenRtbg:
Adding new methods to an interface requires implementation of it in every class implementing the interface. This is incompatible with "I just update and everything works". :)

src/Validation/Constraint/HasClaimWithValue.php
```diff
@@ -0,0 +1,30 @@
...
+
+    public function __construct(string $claimId, string $claimValue)
+    {
+        $this->claimId    = $claimId;
+        $this->claimValue = $claimValue;
+    }
+
+    public function assert(Token $token): void
+    {
+        if (! $token->hasClaimWithValue($this->claimId, $this->claimValue)) {
```
lcobucci: I really like the idea. Just believe that the check could be contained here - which would reduce the changes quite drastically.

We should also verify that we receive a unencrypted token (similar to what's done on the SignedWith constraint). That's important because we don't have access to claims on encrypted tokens.

lcobucci (changes requested):

Slamdunk:
Ok, now I see the need for https://wiki.php.net/rfc/sealed_classes as in this context `Lcobucci\JWT\UnencryptedToken` and `Lcobucci\JWT\Token` are not meant to be implemented outside this library

src/Validation/Constraint/HasClaimWithValue.php
```diff
@@ -0,0 +1,37 @@
...
+use Lcobucci\JWT\Validation\Constraint;
+use Lcobucci\JWT\Validation\ConstraintViolation;
+
+final class HasClaimWithValue implements Constraint
+{
+    private string $claimId;
+
+    private string $claimValue;
+
+    public function __construct(string $claimId, string $claimValue)
```
lcobucci: The value can be anything, not only a string.

```suggestion
    /** @var mixed */
    private mixed $claimValue;
    
    /** @param mixed $claimValue */
    public function __construct(string $claimId, $claimValue)
```

src/Token/Plain.php
```diff
@@ -83,6 +83,11 @@ public function isExpired(DateTimeInterface $now): bool
         return $now >= $this->claims->get(RegisteredClaims::EXPIRATION_TIME);
     }
 
+    public function hasClaimWithValue(string $claimId, string $claimValue): bool
+    {
+        return ($this->claims->has($claimId) && $this->claims->get($claimId) === $claimValue);
+    }
+
```
lcobucci: Not needed anymore

```suggestion
```

src/Validation/Constraint/HasClaimWithValue.php
```diff
@@ -0,0 +1,37 @@
...
+            throw new ConstraintViolation('You should pass a plain token');
+        }
+
+        $claimIsValid = $token->claims()->has($this->claimId) &&
+            $token->claims()->get($this->claimId) === $this->claimValue;
+        if (!$claimIsValid) {
+            throw new ConstraintViolation(
+                'The token did not have the expected claim'
+            );
+        }
```
lcobucci: I'd suggest splitting the checks to have more specific messages

```suggestion
        $claims = $token->claims();
        
        if (! $claims->has($this->claimId)) {
            throw new ConstraintViolation('The token did not have the expected claim');
        }
        
        if ($claims->get($this->claimId) !== $this->claimValue) {
            throw new ConstraintViolation('The claim "' . $this->claimId . '" does not have the expected value');
        }
```

lcobucci (changes requested):

src/Validation/Constraint/HasClaimWithValue.php
SvenRtbg: Only thing is there is no `mixed` type in PHP below 8.0

lcobucci: 😭  true that, something to be done soon

james-bw: Hmmm, should I change this to HasClaimWithStringValue.php, with the expectation to move to a more generic mixed when the library requires >php8.0 ?

lcobucci: What would be the value of having multiple constraints with only different type declarations?

james-bw: To handle the type difference in types you commented on? I can go with whatever, but atm this validation won't be able to do a claim with a different type like a boolean:

`isAdmin: true`

What do you want me to do here?

Ocramius:
@lcobucci before adding this, please check the premise in #826 - the entire idea (so far) smells, IMO

src/Validation/Constraint/HasClaimWithValue.php
lcobucci: > atm this validation won't be able to do a claim with a different type like a boolean:

Using `mixed` allows that and I believe that's the value of having a single implementation.

> What do you want me to do here?

I don't see value in having a string only constraint.
Not declaring a type already defines it as `mixed` and soon we'll drop v7.x support (according to [packagist](https://packagist.org/packages/lcobucci/jwt/php-stats#4) we have about 60% of adoption) we put the type declaration.

lcobucci:
> @lcobucci before adding this, please check the premise in https://github.com/lcobucci/jwt/issues/826 - the entire idea (so far) smells, IMO

@Ocramius I saw it... I'm honestly abstracting that page because I think it's useful to have a generic constraint for custom claims in the lib.

src/Validation/Constraint/HasClaimWithValue.php
lcobucci: I'm thinking now... maybe getting a closure would make things more useful for a generic constraint (any things @Ocramius)?

Something like this (using PHP 8.x syntax just to illustrate):

```php
final class ClaimIsDeclared implements Constraint
{
    /** @var Closure(mixed):bool|null $validateValue
    public function __construct(
        private readonly string $claim,
        private readonly ?Closure $validateValue = null
    ) {}

    public function assert(Token $token): void
    {
        if (! $token instanceof UnencryptedToken) {
            throw new ConstraintViolation('You should pass a plain token');
        }

         if (! $claims->has($this->claimId)) {
            throw new ConstraintViolation('The token did not have the expected claim');
        }

       if ($this->validateValue === null) { return }

        if (! ($this->validateValue)($claims->get($this->claimId))) {
            throw new ConstraintViolation('The claim "' . $this->claimId . '" does not have the expected value');
        }
    }
}

// usage

 // no value validation
new ClaimIsDeclared('testing');

 // inlined closure
new ClaimIsDeclared('testing', static fn (mixed $value): bool => $value === 'something');

 // with fancy 8.1+ things
new ClaimIsDeclared('testing', VerifyExpectedValueForTestingClaim::__invoke(...));
```

SvenRtbg: The value of a claim constraint is that it does specific stuff. Being that generic easily opens the argument why there are for example explict date constraints or subject constraints if they can easily be replaced by just stuffing the correct closure into the generic constraint.

I see that validating if claims exist and possibly if claims have a specifiable value is useful - anything else sound highly user specific, and the difference between writing a closure and implementing `Constraint` is tiny - with slight advantages towards the latter IMO.

lcobucci: That's a fair point

lcobucci: Just to expand on the reason for a closure... the initial motivation to introduce this is related to hash comparison and doing that operation with `===` opens a security issue (timing attacks) - that's why we use `hash_equals()`. 

I'd really love to avoid having two different constraints in the lib for that purpose...

lcobucci:
> Just to expand on the reason for a closure... the initial motivation to introduce this is related to hash comparison and doing that operation with === opens a security issue (timing attacks) - that's why we use hash_equals().

@james-bw have you thought about that? This PR does NOT add a safe way to perform hash comparison.

lcobucci:
Also, you can run `make` to execute most of the checks and solve the problems locally.

james-bw:
> > Just to expand on the reason for a closure... the initial motivation to introduce this is related to hash comparison and doing that operation with === opens a security issue (timing attacks) - that's why we use hash_equals().
> 
> @james-bw have you thought about that? This PR does NOT add a safe way to perform hash comparison.

I did not. I am not aware of how === leads to a timing attack. I'll have to investigate further. 

...

OK I understand the timing attack now, and understand why === is not to be used for cryptographic stuff. However, custom claim validation is no different to existing code: 

```
    public function isIdentifiedBy(string $id): bool
    {
        return $this->claims->get(RegisteredClaims::ID) === $id;
    }

    public function isRelatedTo(string $subject): bool
    {
        return $this->claims->get(RegisteredClaims::SUBJECT) === $subject;
    }
```

I think this validation should be consistent.

james-bw:
> Also, you can run `make` to execute most of the checks and solve the problems locally.

I'm developing on Windows/git bash, so don't have make natively. I've been trying to get my IDE working with the equivalent tests. At this stage I'm about to walk away and leave you to it, I was just trying to help with a quick win and this is turning into ben hur.

lcobucci (approved):

PR #832, opened 2022-04-05
Add Blake2b signature algorithm

Slamdunk:
Hi, JWT has never formalized Blake2b as a signing algorithm, hence the `BLAKE2B` algorithm ID I chose in the code was made up by me and I don't know if this should be merged of not, maybe we could use the `@experimental` tag on the class, I don't know.

But if this is nowhere near a standard (in JWT I mean, Blake2b *is* a standard), why proposing it? Because of this:
```
+--------------+-------------------+-----+------+-----+----------+-----------+---------+
| benchmark    | subject           | set | revs | its | mem_peak | mode      | rstdev  |
+--------------+-------------------+-----+------+-----+----------+-----------+---------+
| EddsaBench   | benchSignature    |     | 100  | 5   | 4.359mb  | 20.221μs  | ±2.59%  |
| EddsaBench   | benchVerification |     | 100  | 5   | 4.359mb  | 45.970μs  | ±0.62%  |
| Sha256Bench  | benchSignature    |     | 100  | 5   | 4.359mb  | 60.589μs  | ±6.34%  |
| Sha256Bench  | benchVerification |     | 100  | 5   | 4.359mb  | 95.866μs  | ±4.34%  |
| Sha384Bench  | benchSignature    |     | 100  | 5   | 4.359mb  | 61.166μs  | ±1.13%  |
| Sha384Bench  | benchVerification |     | 100  | 5   | 4.359mb  | 104.584μs | ±0.76%  |
| Sha512Bench  | benchSignature    |     | 100  | 5   | 4.359mb  | 60.701μs  | ±5.33%  |
| Sha512Bench  | benchVerification |     | 100  | 5   | 4.359mb  | 114.601μs | ±1.13%  |
| Sha256Bench  | benchSignature    |     | 100  | 5   | 4.359mb  | 837.150μs | ±3.02%  |
| Sha256Bench  | benchVerification |     | 100  | 5   | 4.359mb  | 36.188μs  | ±3.82%  |
| Sha384Bench  | benchSignature    |     | 100  | 5   | 4.359mb  | 803.909μs | ±3.31%  |
| Sha384Bench  | benchVerification |     | 100  | 5   | 4.359mb  | 38.787μs  | ±3.91%  |
| Sha512Bench  | benchSignature    |     | 100  | 5   | 4.359mb  | 791.626μs | ±1.59%  |
| Sha512Bench  | benchVerification |     | 100  | 5   | 4.359mb  | 36.873μs  | ±3.04%  |
| NoneBench    | benchSignature    |     | 100  | 5   | 4.359mb  | 0.123μs   | ±38.90% |
| NoneBench    | benchVerification |     | 100  | 5   | 4.359mb  | 0.170μs   | ±2.33%  |
| Sha256Bench  | benchSignature    |     | 100  | 5   | 4.359mb  | 2.218μs   | ±8.02%  |
| Sha256Bench  | benchVerification |     | 100  | 5   | 4.359mb  | 2.410μs   | ±0.20%  |
| Sha384Bench  | benchSignature    |     | 100  | 5   | 4.359mb  | 2.262μs   | ±13.38% |
| Sha384Bench  | benchVerification |     | 100  | 5   | 4.359mb  | 2.451μs   | ±0.33%  |
| Sha512Bench  | benchSignature    |     | 100  | 5   | 4.359mb  | 2.285μs   | ±6.95%  |
| Sha512Bench  | benchVerification |     | 100  | 5   | 4.359mb  | 2.491μs   | ±10.81% |
| Blake2bBench | benchSignature    |     | 100  | 5   | 4.359mb  | 0.562μs   | ±2.72%  |
| Blake2bBench | benchVerification |     | 100  | 5   | 4.359mb  | 0.760μs   | ±0.83%  |
+--------------+-------------------+-----+------+-----+----------+-----------+---------+
```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -70,11 +70,12 @@ $configuration = Configuration::forSymmetricSigner(
 
 Currently supported symmetric algorithms:
 
-| Name      | Description        | Class                                    | Key length req. |
-| --------- | ------------------ | ---------------------------------------- |-----------------|
-| `HS256`   | HMAC using SHA-256 | `\Lcobucci\JWT\Signer\Hmac\Sha256`       | 256 bits        |
-| `HS384`   | HMAC using SHA-384 | `\Lcobucci\JWT\Signer\Hmac\Sha384`       | 384 bits        |
-| `HS512`   | HMAC using SHA-512 | `\Lcobucci\JWT\Signer\Hmac\Sha512`       | 512 bits        |
+| Name      | Description        | Class                               | Key length req. | Notes                                                                                                      |
+|-----------|--------------------|-------------------------------------|-----------------|------------------------------------------------------------------------------------------------------------|
+| `HS256`   | HMAC using SHA-256 | `\Lcobucci\JWT\Signer\Hmac\Sha256`  | 256 bits        |                                                                                                            |
+| `HS384`   | HMAC using SHA-384 | `\Lcobucci\JWT\Signer\Hmac\Sha384`  | 384 bits        |                                                                                                            |
+| `HS512`   | HMAC using SHA-512 | `\Lcobucci\JWT\Signer\Hmac\Sha512`  | 512 bits        |                                                                                                            |
+| `BLAKE2B` | Blake2b keyed Hash | `\Lcobucci\JWT\Signer\Hmac\Blake2b` | 256 bits        | Not a [JWT standard](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms) |
 
 Deprecated symmetric algorithms in `v4`:
 
--- a/src/Signer/Blake2b.php
+++ b/src/Signer/Blake2b.php
@@ -1,0 +1,36 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Signer;
+
+use Lcobucci\JWT\Signer;
+
+use function hash_equals;
+use function sodium_crypto_generichash;
+use function strlen;
+
+final class Blake2b implements Signer
+{
+    private const MINIMUM_KEY_LENGTH_IN_BITS = 256;
+
+    public function algorithmId(): string
+    {
+        return 'BLAKE2B';
+    }
+
+    public function sign(string $payload, Key $key): string
+    {
+        $actualKeyLength = 8 * strlen($key->contents());
+
+        if ($actualKeyLength < self::MINIMUM_KEY_LENGTH_IN_BITS) {
+            throw InvalidKeyProvided::tooShort(self::MINIMUM_KEY_LENGTH_IN_BITS, $actualKeyLength);
+        }
+
+        return sodium_crypto_generichash($payload, $key->contents());
+    }
+
+    public function verify(string $expected, string $payload, Key $key): bool
+    {
+        return hash_equals($expected, $this->sign($payload, $key));
+    }
+}
--- a/test/performance/Blake2bBench.php
+++ b/test/performance/Blake2bBench.php
@@ -1,0 +1,35 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT;
+
+use Lcobucci\JWT\Signer\Blake2b;
+use Lcobucci\JWT\Signer\Key;
+use Lcobucci\JWT\Signer\Key\InMemory;
+use PhpBench\Benchmark\Metadata\Annotations\Groups;
+
+/** @Groups({"Hmac"}) */
+final class Blake2bBench extends SignerBench
+{
+    private const ENCODED_KEY = 'b6DNRcX2SFapbICe6lXWYoOZA+JXL/dvkfWiv2hJv3Y=';
+
+    protected function signer(): Signer
+    {
+        return new Blake2b();
+    }
+
+    protected function signingKey(): Key
+    {
+        return $this->createKey();
+    }
+
+    protected function verificationKey(): Key
+    {
+        return $this->createKey();
+    }
+
+    private function createKey(): Key
+    {
+        return InMemory::base64Encoded(self::ENCODED_KEY);
+    }
+}
--- a/test/unit/Signer/Blake2bTest.php
+++ b/test/unit/Signer/Blake2bTest.php
@@ -1,0 +1,98 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Signer;
+
+use Lcobucci\JWT\Signer\Key\InMemory;
+use Lcobucci\JWT\SodiumBase64Polyfill;
+use PHPUnit\Framework\TestCase;
+
+use function hash_equals;
+
+/**
+ * @coversDefaultClass \Lcobucci\JWT\Signer\Blake2b
+ *
+ * @uses \Lcobucci\JWT\Signer\Key\InMemory
+ * @uses \Lcobucci\JWT\SodiumBase64Polyfill::base642bin()
+ */
+final class Blake2bTest extends TestCase
+{
+    private const KEY_ONE                    = 'GOu4rLyVCBxmxP+sbniU68ojAja5PkRdvv7vNvBCqDQ=';
+    private const KEY_TWO                    = 'Pu7gywseH+R5HLIWnMll4rEg1ltjUPq/P9WwEzAsAb8=';
+    private const CONTENTS                   = 'test';
+    private const EXPECTED_HASH_WITH_KEY_ONE = '/TG5kmkav/YGl3I9uQiv4cm1VN6Q0zPCom4G7+p74JU=';
+
+    private const SHORT_KEY = 'PIBQuM5PopdMxtmTWmyvNA==';
+
+    private InMemory $keyOne;
+    private InMemory $keyTwo;
+    private string $expectedHashWithKeyOne;
+
+    /** @before */
+    public function initializeKey(): void
+    {
+        $this->keyOne                 = InMemory::base64Encoded(self::KEY_ONE);
+        $this->keyTwo                 = InMemory::base64Encoded(self::KEY_TWO);
+        $this->expectedHashWithKeyOne = SodiumBase64Polyfill::base642bin(
+            self::EXPECTED_HASH_WITH_KEY_ONE,
+            SodiumBase64Polyfill::SODIUM_BASE64_VARIANT_ORIGINAL
+        );
+    }
+
+    /**
+     * @test
+     *
+     * @covers ::algorithmId
+     */
+    public function algorithmIdMustBeCorrect(): void
+    {
+        $signer = new Blake2b();
+
+        self::assertEquals('BLAKE2B', $signer->algorithmId());
+    }
+
+    /**
+     * @test
+     *
+     * @covers ::sign
+     * @covers ::verify
+     */
+    public function generatedSignatureMustBeSuccessfullyVerified(): void
+    {
+        $signer = new Blake2b();
+
+        self::assertTrue(hash_equals($this->expectedHashWithKeyOne, $signer->sign(self::CONTENTS, $this->keyOne)));
+        self::assertTrue($signer->verify($this->expectedHashWithKeyOne, self::CONTENTS, $this->keyOne));
+    }
+
+    /**
+     * @test
+     *
+     * @covers ::sign
+     *
+     * @uses \Lcobucci\JWT\Signer\InvalidKeyProvided
+     */
+    public function signShouldRejectShortKeys(): void
+    {
+        $signer = new Blake2b();
+
+        $this->expectException(InvalidKeyProvided::class);
+        $this->expectExceptionMessage('Key provided is shorter than 256 bits, only 128 bits provided');
+
+        $signer->sign(self::CONTENTS, InMemory::base64Encoded(self::SHORT_KEY));
+    }
+
+    /**
+     * @test
+     *
+     * @covers ::sign
+     * @covers ::verify
+     */
+    public function verifyShouldReturnFalseWhenExpectedHashWasNotCreatedWithSameInformation(): void
+    {
+        $signer = new Blake2b();
+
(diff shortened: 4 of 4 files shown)


Ocramius (approved):
IMO OK to introduce this - it is still important that:

 1. we document it (not sure which files need adjusting)
 2. we make sure it's documented as in "this is a more niche algorithm, so downstream systems may be unable to verify generated tokens, if they don't support it"

composer.lock
```diff
@@ -8,31 +8,30 @@
     "packages": [
         {
             "name": "lcobucci/clock",
-            "version": "2.0.0",
```
Ocramius: Let's revert all `composer.lock` changes, for now

Ocramius (dismissed):

Ocramius:
Threw it out to pull in some attention: https://twitter.com/Ocramius/status/1511338900043841537

test/unit/Signer/Hmac/Blake2bTest.php
```diff
@@ -0,0 +1,90 @@
...
+     *
+     * @covers ::sign
+     *
+     * @uses \Lcobucci\JWT\Signer\Key\InMemory
+     */
+    public function signMustReturnAHashAccordingWithTheAlgorithm(): string
+    {
+        $hash = sodium_crypto_generichash('test', $this->key);
+
+        self::assertSame($hash, (new Blake2b())->sign('test', InMemory::plainText($this->key)));
```
lcobucci: I know this is a test but I think we should lead by example =p

```suggestion
        self::assertTrue(hash_equals($hash, (new Blake2b())->sign('test', InMemory::plainText($this->key))));
```

lcobucci:
I'm okay too (100% agree with the documentation comments).

test/unit/Signer/Hmac/Blake2bTest.php
Slamdunk: That's copied from https://github.com/lcobucci/jwt/blob/62a0bf15f11493c140b3bf8681380696d23d172e/test/unit/Signer/HmacTest.php#L43

Maybe a better test should hardcode the hash, WDYT?

lcobucci: Yeah, that's me from the past leading by the example (with wrong examples) 😅

lcobucci: I, today, honestly wouldn't split signature creation and verification in different test methods.

On mobile now but will send a suggestion over in a few minutes.

lcobucci: Thinking now I think pre-generated hash makes sense :smile:

test/unit/Signer/Hmac/Blake2bTest.php
```diff
@@ -0,0 +1,90 @@
...
+     * @uses \Lcobucci\JWT\Signer\Key\InMemory
+     */
+    public function verifyShouldReturnFalseWhenExpectedHashWasNotCreatedWithSameInformation(string $expected): void
+    {
+        self::assertFalse((new Blake2b())->verify(
+            $expected,
+            'test',
+            InMemory::plainText(sodium_crypto_generichash_keygen())
+        ));
+    }
```
lcobucci: Maybe this?

```suggestion
    private const CONTENTS = 'test'; 
    private InMemory $key;
    private string $expectedHash;

    /** @before */
    public function initializeKey(): void
    {
        $this->key = InMemory::plainText(sodium_crypto_generichash_keygen());
        $this->expectedHash = sodium_crypto_generichash(self::CONTENTS, $this->key->contents());
    }

    /**
     * @test
     *
     * @covers ::algorithmId
     */
    public function algorithmIdMustBeCorrect(): void
    {
        $signer = new Blake2b();

        self::assertSame('BLAKE2B', $signer->algorithmId());
    }

    /**
     * @test
     *
     * @covers ::sign
     * @covers ::verify
     *
     * @uses \Lcobucci\JWT\Signer\Key\InMemory
     */
    public function generatedSignatureMustBeSuccessfullyVerified(): void
    {
        $algorithm = new Blake2b();
         
        self::assertTrue(hash_equals($this->expectedHash, algorithm->sign(self::CONTENTS, $this->key)));
        self::assertTrue($algorithm->verify($this->expectedHash, self::CONTENTS, $this->key));
    }

    /**
     * @test
     *
     * @covers ::verify
     *
     * @uses \Lcobucci\JWT\Signer\Hmac\Blake2b::sign
     * @uses \Lcobucci\JWT\Signer\Key\InMemory
     */
    public function verifyShouldReturnFalseWhenExpectedHashWasNotCreatedWithSameInformation(string $expected): void
    {
        self::assertFalse((new Blake2b())->verify(
            $this->expectedHash,
            self::CONTENTS,
            InMemory::plainText(sodium_crypto_generichash_keygen())
        ));
    }
```

lcobucci: It's fine to have everything fixed too btw...

lcobucci:
Looks okay to me, would you please add the docs?

Slamdunk:
Sure, tomorrow

jedisct1:
Woah, this is great.

Ocramius:
One consideration to check here: is the generated hash longer or shorter, and how much? A longer hash would mean lower CPU load, but higher I/O load.

I see Blake2B is supposed to produce 64 bytes? If that is the case, it is most likely a big win also for everything I/O related.

EDIT: as for security guarantees, it seems like Blake2b is theorically a lot better than sha256

Slamdunk:
> I see Blake2B is supposed to produce 64 bytes?

It can range between 16 and 64, but default implementation including `libsodium`'s one is 32 bytes, so same output length as `sha256`. the most common JWT sym-algo

jedisct1:
For long-term tokens, a minimum of 24 bytes is required for collision resistance. 16 bytes are fine for short-lived tokens. 32 bytes are safe for all use cases, adds a security margin.

Using 64 bytes would be useless in the case of JWT.

Speed is the same regardless of the output size; what really matters is the block size (how many bytes are processed at one) and it is always 512 bits in BLAKE2b.

So, like @Slamdunk I'd also recommend using 32 bytes output.

jedisct1:
BLAKE2b also accepts an optional "personalization" parameter (sometimes called "context" in other hash functions and constructions). The purpose is to produce a different output in case the same key is used in different contexts, in order to avoid cross-protocol attacks.

Setting it to a constant such as the "JWT" string may not be a bad idea.

Slamdunk:
> BLAKE2b also accepts an optional "personalization" parameter

`libsodium` doesn't expose it on high-level API (see https://libsodium.gitbook.io/doc/hashing/generic_hashing#notes).

I like the idea but I want to stick to a standard implementation as much as possible.

jedisct1:
I know, I'm libsodium's author :)

It's available as `crypto_generichash_blake2b_salt_personal()`.

The parameter is available in all other BLAKE2b implementations as well.

lcobucci:
Some interesting discussions in https://mailarchive.ietf.org/arch/msg/jose/YAQf8peqpqNTNpK2VJFFckbR1P4/. @jedisct1 maybe you can help out?

Ocramius:
From that mail thread:

 > As for process for adding it, I would write Internet-Draft specifying
 > the use in JOSE and then use the I-D as specification for expert
 > review. One might do a comment round through this list first to catch
 > the worst issues...

Sounds like it may take a while, if it ever happens at all.

jedisct1:
> From that mail thread:
> 
> > As for process for adding it, I would write Internet-Draft specifying
> > the use in JOSE and then use the I-D as specification for expert
> > review. One might do a comment round through this list first to catch
> > the worst issues...
> 
> Sounds like it may take a while, if it ever happens at all.

Like all standards. It take years for a proposal to become accepted as a standard, if that ever happens, and by the time it finally happens, the algorithm is already obsolete.

lcobucci:
> Like all standards. It take years for a proposal to become accepted as a standard, if that ever happens, and by the time it finally happens, the algorithm is already obsolete.

Yeap... looking at timelines as https://datatracker.ietf.org/doc/rfc8037/ makes me shiver... I don't have the capacity to create the draft (I'm even struggling to keep up with basic OSS).

We must decide if we want to push this further (just remove references from HMAC as native Blake2b MAC is different from HMAC-Blake2) and accept the usage of a non-standard algorithm or hold this until someone creates the Internet Draft and submit it for review (following [RFC7518's procedure](https://www.rfc-editor.org/rfc/rfc7518.html#section-7)).

> Even if not the case here, sigantures using BLAKE2B as hash would be significantly more problematic: The only space/time efficient signature algorithm allowing variable hashes that is even close to widespread is ECDSA, and ECDSA requires bit order of hash, which BLAKE2 just does not specify.

That bit also came from the thread and makes me feel like the "Designated Experts" might reject BLAKE2b's submission, idk.
I'm honestly not knowledgeable enough on the nitty-gritty details of hashing to understand if not requiring bit order of hash would be a potential problem for Blake2b's adoption - I don't really mind the fact it's not widespread atm.

I'm generally okay with adding a non-standard algorithm if you're all up to it.

jedisct1:
"de facto" standards work.

If common implementations are compatible with each other, this is enough to make something a lot of people can use and take advantage of.

docs/configuration.md
```diff
@@ -70,11 +70,12 @@ $configuration = Configuration::forSymmetricSigner(
...
+| --------- | ------------------ | ------------------------------------ |-----------------|------------------------------------------------------------------------------------------------------------|
+| `HS256`   | HMAC using SHA-256 | `\Lcobucci\JWT\Signer\Hmac\Sha256`   | 32 bytes        |                                                                                                            |
+| `HS384`   | HMAC using SHA-384 | `\Lcobucci\JWT\Signer\Hmac\Sha384`   | 48 bytes        |                                                                                                            |
+| `HS512`   | HMAC using SHA-512 | `\Lcobucci\JWT\Signer\Hmac\Sha512`   | 64 bytes        |                                                                                                            |
+| `BLAKE2B` | Blake2b keyed Hash | `\Lcobucci\JWT\Signer\Hmac\Blake2b`  | 16 bytes        | Not a [JWT standard](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms) |
```
lcobucci: I added the min length required by libsodium... should we align with HMAC algos and bump this to 32?

src/Signer/Hmac/Blake2b.php
```diff
@@ -0,0 +1,28 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Signer\Hmac;
```
lcobucci: Blake2 is a MAC not HMAC, let's move it to `Lcobucci\JWT\Signer`?

src/Signer/Hmac/Blake2b.php
```diff
@@ -0,0 +1,28 @@
...
+use Lcobucci\JWT\Signer\Key;
+
+use function hash_equals;
+use function sodium_crypto_generichash;
+
+final class Blake2b implements Signer
+{
+    public function algorithmId(): string
+    {
+        return 'BLAKE2B';
```
lcobucci: I was thinking that we should make the length visible and open for extension.

```suggestion
        return 'B2-256';
```

src/Signer/Hmac/Blake2b.php
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public function sign(string $payload, Key $key): string
+    {
+        return sodium_crypto_generichash($payload, $key->contents());
```
lcobucci: We should verify the min key size here.

lcobucci (changes requested):

docs/configuration.md
Slamdunk: Sounds reasonable

src/Signer/Hmac/Blake2b.php
Slamdunk: Blake *is* an hash function, so it is a HMAC, doesn't it? :thinking:

src/Signer/Hmac/Blake2b.php
Slamdunk: I don't like this: the numbers seen in other algorithm IDs want to represent the algorithms used, not their output length; see for example `EdDSA`.

`Blake2b` has no variants, so there's only one name.

src/Signer/Hmac/Blake2b.php
lcobucci: AFAIK, HMAC is a MAC algo that is backed up by a hashing algo (ergo hmac-sha*). There are even implementations of hmac-blake2.

Blake2 has a MAC mode (when the key is provided).

src/Signer/Hmac/Blake2b.php
lcobucci: That's fair, I'd argue that it only should be Blake2 (as the 32 bytes output is compatible with Blake2s).

src/Signer/Hmac/Blake2b.php
Slamdunk: Heh, definitions are indeed sharper then reality.

Right, this is not a pure HMAC impl (it's better since it pads the key, and so on) even though it's practically the same impl

src/Signer/Hmac/Blake2b.php
jedisct1: I'd stick with `BLAKE2B`.

No matter what the output length is, BLAKE2B and BLAKE2S would produce different outputs.

docs/configuration.md
lcobucci: Handled :+1:

src/Signer/Hmac/Blake2b.php
lcobucci: Handled :+1:

src/Signer/Hmac/Blake2b.php
lcobucci: Handled :+1:

lcobucci (approved):

PR #833, opened 2022-04-07
Key: permit empty keys only with `::empty()` factory method

Slamdunk:
I consider this a security bug that should be addressed with urgent.

Before this PR, misconfigurations can easily lead to unsecured token issuance under the radar, expecially where creator = consumer.

### DIFF
--- a/src/Signer/Key/InMemory.php
+++ b/src/Signer/Key/InMemory.php
@@ -3,6 +3,7 @@ declare(strict_types=1);
 
 namespace Lcobucci\JWT\Signer\Key;
 
+use Lcobucci\JWT\Signer\InvalidKeyProvided;
 use Lcobucci\JWT\Signer\Key;
 use Lcobucci\JWT\SodiumBase64Polyfill;
 use SplFileObject;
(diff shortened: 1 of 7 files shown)


Ocramius:
I'd still label it as BC break, but better to have a broken system, than a compromised one.

No need for CVE/security issue, since this is mis-configuration on the consumer side, if it happens: instructions on using safe randomly generated keys were already provided.

src/Signer/Key/InMemory.php
```diff
@@ -18,13 +19,21 @@ final class InMemory implements Key
 
     private function __construct(string $contents, string $passphrase)
```
Ocramius, line 21: We should document `$contents` as `@param non-empty-string $contents`, then suppress any raised static analysis issues reported because of it, for now.

Ocramius (changes requested):

Ocramius (approved):
Awesome, thanks @Slamdunk!

src/Signer/Key/InMemory.php
```diff
@@ -16,15 +17,24 @@ final class InMemory implements Key
...
     {
+        if ($contents === '') {
+            throw InvalidKeyProvided::cannotBeEmpty();
+        }
+
         $this->contents   = $contents;
         $this->passphrase = $passphrase;
     }
 
     public static function empty(): self
```
Ocramius, line 31: Should we mark this as `@deprecated`?

Slamdunk: I don't think so: legit usages exist.
What matters is that consumer is aware it's empty.

lcobucci:
IHMO we shouldn't use the baseline to ignore errors we will never fix. That's why we have the annotations to ignore things.

Ocramius:
> IHMO we shouldn't use the baseline to ignore errors we will never fix.

We should probably remove the exception, at some point, and only leave the types

PR #835, opened 2022-04-07
Require minimum key size for HMAC algorithm

Slamdunk:
Hi, as reported to us by @Spomky this library currently lacks a key security requirement of JWT standard reported here: https://www.rfc-editor.org/rfc/rfc7518.html#section-3.2

    A key of the same size as the hash output (for instance, 256 bits for
    "HS256") or larger MUST be used with this algorithm.  (This
    requirement is based on [Section 5.3.4] (Security Effect of the HMAC
    Key) of NIST SP 800-117 [[NIST.800-107], which states that the
    effective security strength is the minimum of the security strength
    of the key and two times the size of the internal hash value.)

Current version of `lcobucci/jwt` allows indeed any weak key like `1234` to be used for HMAC, for example.

While this issue will hopefully be resolved for future algorithms by low level APIs (see `sodium_crypto_sign_detached` and `sodium_crypto_generichash` implementations) I think we should design this library to require any current and future Signer to provide and apply explicit Key validation.

I'm throwing here a PR because, you know, it may be easier to fix something rather than creating it from scratch.
Comments on the code will follow to explain it.

`BC-Break` is intentional even though I know a MAJOR bump is a blocker: I'd like to shape it the correct way, and only after that adapt it to the current `v4` branch.

### DIFF
--- a/src/Signer/Hmac.php
+++ b/src/Signer/Hmac.php
@@ -7,13 +7,13 @@ use Lcobucci\JWT\Signer;
 
 use function hash_equals;
 use function hash_hmac;
-use function mb_strlen;
+use function strlen;
 
 abstract class Hmac implements Signer
 {
     final public function sign(string $payload, Key $key): string
     {
-        $actualKeyLength   = mb_strlen($key->contents(), '8bit');
+        $actualKeyLength   = strlen($key->contents());
         $expectedKeyLength = $this->minimumBytesLengthForKey();
         if ($actualKeyLength < $expectedKeyLength) {
             throw InvalidKeyProvided::tooShort($expectedKeyLength, $actualKeyLength);
--- a/src/Signer/Hmac.php
+++ b/src/Signer/Hmac.php
@@ -7,11 +7,18 @@ use Lcobucci\JWT\Signer;
 
 use function hash_equals;
 use function hash_hmac;
+use function mb_strlen;
 
 abstract class Hmac implements Signer
 {
     final public function sign(string $payload, Key $key): string
     {
+        $actualKeyLength   = mb_strlen($key->contents(), '8bit');
+        $expectedKeyLength = $this->minimumBytesLengthForKey();
+        if ($actualKeyLength < $expectedKeyLength) {
+            throw InvalidKeyProvided::tooShort($expectedKeyLength, $actualKeyLength);
+        }
+
         return hash_hmac($this->algorithm(), $payload, $key->contents(), true);
     }
 
@@ -21,4 +28,7 @@ abstract class Hmac implements Signer
     }
 
     abstract public function algorithm(): string;
+
+    /** @return positive-int */
+    abstract public function minimumBytesLengthForKey(): int;
 }
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -68,6 +68,22 @@ $configuration = Configuration::forSymmetricSigner(
 );
 ```
 
+Currently supported symmetric algorithms:
+
+| Name      | Description        | Class                                    | Key length req. |
+| --------- | ------------------ | ---------------------------------------- | --------------- |
+| `HS256`   | HMAC using SHA-256 | `\Lcobucci\JWT\Signer\Hmac\Sha256`       | 32 bytes        |
+| `HS384`   | HMAC using SHA-384 | `\Lcobucci\JWT\Signer\Hmac\Sha384`       | 48 bytes        |
+| `HS512`   | HMAC using SHA-512 | `\Lcobucci\JWT\Signer\Hmac\Sha512`       | 64 bytes        |
+
+Deprecated symmetric algorithms in `v4`:
+
+| Name      | Description        | Class                                    | Key length req. |
+| --------- | ------------------ | ---------------------------------------- | --------------- |
+| `HS256`   | HMAC using SHA-256 | `\Lcobucci\JWT\Signer\Hmac\UnsafeSha256` | 1 byte          |
+| `HS384`   | HMAC using SHA-384 | `\Lcobucci\JWT\Signer\Hmac\UnsafeSha384` | 1 byte          |
+| `HS512`   | HMAC using SHA-512 | `\Lcobucci\JWT\Signer\Hmac\UnsafeSha512` | 1 byte          |
+
 #### For asymmetric algorithms
 
 Asymmetric algorithms use a **private key** for signature creation and a **public key** for verification.
@@ -90,6 +106,18 @@ $configuration = Configuration::forAsymmetricSigner(
 !!! Important
     The implementation of ECDSA algorithms have a constructor dependency.
     Use the `create()` named constructor to avoid having to handle it (e.g.: `Lcobucci\JWT\Signer\Ecdsa\Sha256::create()`).
+
+Currently supported asymmetric algorithms:
+
+| Name    | Description                     | Class                               |
+| ------- | ------------------------------- | ----------------------------------- |
+| `ES256` | ECDSA using P-256 and SHA-256   | `\Lcobucci\JWT\Signer\Ecdsa\Sha256` |
+| `ES384` | ECDSA using P-384 and SHA-384   | `\Lcobucci\JWT\Signer\Ecdsa\Sha384` |
+| `ES512` | ECDSA using P-521 and SHA-512   | `\Lcobucci\JWT\Signer\Ecdsa\Sha512` |
+| `RS256` | RSASSA-PKCS1-v1_5 using SHA-256 | `\Lcobucci\JWT\Signer\Rsa\Sha256`   |
+| `RS384` | RSASSA-PKCS1-v1_5 using SHA-384 | `\Lcobucci\JWT\Signer\Rsa\Sha384`   |
+| `RS512` | RSASSA-PKCS1-v1_5 using SHA-512 | `\Lcobucci\JWT\Signer\Rsa\Sha512`   |
+| `EdDSA` | EdDSA signature algorithms      | `\Lcobucci\JWT\Signer\Eddsa`        |
 
 #### For no algorithm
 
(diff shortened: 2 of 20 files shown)


src/Signer.php
```diff
@@ -5,11 +5,13 @@
 
 use Lcobucci\JWT\Signer\CannotSignPayload;
 use Lcobucci\JWT\Signer\Ecdsa\ConversionFailed;
-use Lcobucci\JWT\Signer\InvalidKeyProvided;
-use Lcobucci\JWT\Signer\Key;
 
 interface Signer
 {
+    public static function generateKey(): mixed;
```
Slamdunk: Documenting key generation is good, providing key generation is better.

src/Signer.php
```diff
@@ -5,11 +5,13 @@
...
+
+    public function validateAndConsumeKey(string $contents, string $passphrase): void;
```
Slamdunk: 1. Every signer must implement Key validation. Even `None` signer has to have it: explicit code is always better
1. I despise `validateAndConsumeKey` method name, but Signer might need to consume key before validation, see https://github.com/lcobucci/jwt/blob/ae3aac816d4e716e2cd26c5085559409a8432d46/src/Signer/OpenSSL.php#L58-L59

src/Signer.php
```diff
@@ -19,16 +21,14 @@ public function algorithmId(): string;
      * Creates a hash for the given payload
      *
      * @throws CannotSignPayload  When payload signing fails.
-     * @throws InvalidKeyProvided When issue key is invalid/incompatible.
      * @throws ConversionFailed   When signature could not be converted.
      */
-    public function sign(string $payload, Key $key): string;
+    public function sign(string $payload): string;
```
Slamdunk: OpenSSL with encrypted keys requires us to either have a lot of `mixed` typehints or make Signers stateful.
I hate both, but preferred the latter to start this PR.

Signers here have to save the actual key while they validate it in `validateAndConsumeKey`

src/Signer/Key/InMemory.php
```diff
@@ -67,13 +67,17 @@ public static function file(string $path, string $passphrase = ''): self
         return new self($contents, $passphrase);
     }
 
-    public function contents(): string
+    public function sign(Signer $signer, string $payload): string
     {
-        return $this->contents;
+        $signer->validateAndConsumeKey($this->contents, $this->passphrase);
```
Slamdunk: `InMemory` (which should just be renamed as `Key` but I wanted to minimize PR diff) **must** now become a standalone class without interface, so we can enforce this `$signer->validateAndConsumeKey` call

Slamdunk (commented):
The diff below represents a way to be sure that:
1. Signers provide key validation
2. Key are actually validated against it

Everything here can easily be backported in `v4` implementations, waiting `v5` to updated `Signer` interface

src/Signer.php
Ocramius: IMO this opens a can of worms - maintaining this ourselves means more security issues.

Slamdunk: I understand your point. Consider that:

1. It should be a one-liner: `random_bytes(32)`, `openssl_pkey_new(['private_key_type' => $this->keyType())`, `sodium_crypto_generichash_keygen()` etc
2. I would prefer to take this responsability, rather than seeing new developers still copying `uniqid()` output as valid key

Ocramius: I'm on board as long as you are OK with fighting more toxic cryptographers :D

lcobucci:
I need to think more about this but I think this is not the way to go. I'll review it on the laptop later today

src/Signer.php
lcobucci: The consumers of this method are different from the sign/verify.
Although I understand the reasons why you want to provide it, I'd postpone it and offer a different interface.

src/Signer.php
lcobucci: Having the method doesn't guarantee its execution, I don't think we're making the code better by enforcing validation this way.

src/Signer/Key/InMemory.php
lcobucci: That makes no sense from a design PoV... a key doesn't sign anything.
We also remove the extension point for implementing JWK for no good reason.

lcobucci (changes requested):
I believe this PR is going way beyond to solve the issue and by doing so creates an API that doesn't make sense to me (e.g. `Key#sign()`).

We need to rethink this.

Slamdunk:
@lcobucci new day, new RFC :bowtie:

Slamdunk:
Ping :flushed:

Spomky:
Hi,

I see some possibilities on Signer side that could solve this:

1. Each [HMAC](https://github.com/lcobucci/jwt/tree/4.2.x/src/Signer/Hmac) overrides the `sign` method and add an assertion to check the key length before calling the parent function.
2. The creation of an interface (let say `HasKeyConstraintInterface`) with a method that will check the key (`protected function checkKey(Key $key): void`). In signer classes, if `$this` implements that interface, the method is called before computation.

Note that I have not tested these solutions. `1.` will require relaxing on the [`final` statement](https://github.com/lcobucci/jwt/blob/4.2.x/src/Signer/Hmac.php#L13). Solution `2.` seems more convenient and can be applied to the HMAC or any other Signer.

WDYT?

lcobucci:
I believe that introducing an interface to handle the key validation generically might create more confusion on the usage.
I honestly see that process tightly coupled to the algorithm performing the signing.

> Each [HMAC](https://github.com/lcobucci/jwt/tree/4.2.x/src/Signer/Hmac) overrides the sign method and add an assertion to check the key length before calling the parent function.

I think that's the way I'd go, but I'd leverage the capabilities of abstract classes instead of overriding `Hmac#sign()`:

```php
abstract class Hmac implements Signer
{
    final public function sign(string $payload, Key $key): string
    {
        $this->guardAgainstWeakKeys($key);

        return hash_hmac($this->algorithm(), $payload, $key->contents(), true);
    }

    private function guardAgainstWeakKeys(Key $key): void
    {
        if (mb_strlen($key->contents(), '8bit') >= $this->minimumLengthForKeys()) {
            return;
        }

        throw new InvalidKeyProvided();
    }

    abstract public function minimumLengthForKeys(): int;

    final public function verify(string $expected, string $payload, Key $key): bool
    {
        return hash_equals($expected, $this->sign($payload, $key));
    }

    abstract public function algorithm(): string;
}
```

Then, we need to check the trade-offs on the migration path (do we want to introduce classes on the side and mark the existing ones as deprecated, do we want to trigger deprecation errors instead of exceptions on v4.x, how do we avoid breaking BC on the abstract class, ...).

Slamdunk:
> Then, we need to check the trade-offs on the migration path (do we want to introduce classes on the side and mark the existing ones as deprecated, do we want to trigger deprecation errors instead of exceptions on v4.x, how do we avoid breaking BC on the abstract class, ...).

Considering that the current state of art is both **out of standard** and **unsafe**, my last push proposes the following:

1. `Sha256`, `Sha384` and `Sha512` now raise an `InvalidKeyProvided` exception when key doesn't fulfill requirements
2. new classes `UnsafeSha256`, `UnsafeSha384` and `UnsafeSha512` are introduced with minimum key length of `1`, and are already marked as `@deprecated`

After upgrading to the new release:

1. Users with already valid keys notice nothing
2. Users with invalid keys must choose to either update their keys or swap to `UnsafeShaXXX` alternative

As far as I can tell, this approach is both backward compatible and caring for security of end users.

src/Signer/Hmac.php
```diff
@@ -7,11 +7,18 @@
 
 use function hash_equals;
 use function hash_hmac;
+use function strlen;
 
 abstract class Hmac implements Signer
 {
     final public function sign(string $payload, Key $key): string
     {
+        $actualKeyLength   = strlen($key->contents());
```
lcobucci, line 16: Should we worry about `mbstring.func_overload` and use `mb_string($key->contents(), '8bit')` instead? That could affect the number of returned bytes when dealing with multibyte chars.

AFAIK the worst-case scenario would be reporting fewer bytes (considering `UTF-8` as the internal default encoding). @Ocramius @Slamdunk do you know of any other implication?

lcobucci (changes requested):
This is behavioural BC-break, introducing the old behaviour on the side might be alright - especially because people using decent keys should not be affected - but we must document this.

@Slamdunk can you modify the docs to mention HMAC key length requirements?

src/Signer/Hmac.php
Slamdunk: From https://www.php.net/manual/en/mbstring.overload.php:

> This feature has been DEPRECATED as of PHP 7.2.0, and REMOVED as of PHP 8.0.0. Relying on this feature is highly discouraged.

Although not a scientific approach, above statement and https://3v4l.org/KHYVk#v8.1.6 are enough for me

lcobucci: I know the differences... I also know that not everyone uses random binary strings as keys as they should 😢 (some people are crazy enough to use emojis as variable names 🤣) 

I'm not asking for changes, though, just trying to find out the impact.

Slamdunk: emojis are a subset of `random_bytes` as far as I know

lcobucci: Yeah, the length will be reported differently: https://3v4l.org/CTY0s

Slamdunk: > Yeah, the length will be reported differently

???

https://3v4l.org/bfP4f#v8.1.6

lcobucci: I think you're not comparing it correctly, as the override is likely to be disabled for 3v4l.org.

I've added a link on the previous comment.

lcobucci: And the link was wrong 🤦 https://3v4l.org/CTY0s

lcobucci: So, the only implication I can think of is having `strlen()` reporting fewer characters when:

1. the key uses multibyte chars
1. the override is enabled
1. the internal encoding is utf8

This is not a security problem as our verification is stricter. Is there any other implication?

Slamdunk:
@lcobucci I'm sorry I lost where we were here

lcobucci (approved):

src/Signer/Hmac.php
lcobucci: @Slamdunk this was the only pending discussion but I think we're good as is.
Thanks for your help.

PR #855, opened 2022-07-27
Require minimum key size for OpenSSL keys

### LINKED ISSUE #854: Require minimum key size for RSA keys
As stated on [JWA 3.3](https://datatracker.ietf.org/doc/html/rfc7518#section-3.3), RSA keys need to be a least 2048 long:

> Digital Signature with RSASSA-PKCS1-v1_5
>
>    This section defines the use of the RSASSA-PKCS1-v1_5 digital
>    signature algorithm as defined in [Section 8.2 of RFC 3447](https://www.rfc-editor.org/rfc/rfc3447#section-8.2) [[RFC3447](https://www.rfc-editor.org/rfc/rfc3447)]
>    (commonly known as PKCS #1), using SHA-2 [[SHS](https://www.rfc-editor.org/rfc/rfc7518#ref-SHS)] hash functions.
> 
>    A key of size 2048 bits or larger MUST be used with these algorithms.

Slamdunk:
Fixes https://github.com/lcobucci/jwt/issues/854

Few notes to read the PR effectively:

1. `Hmac::minimumBytesLengthForKey` has been renamed to `Hmac::minimumBitsLengthForKey` but it's not a BC-Break since it hasn't been released yet
3. Like in https://github.com/lcobucci/jwt/pull/835#issuecomment-1129792312 main classes now require a safe minimum key length, `Unsafe*` classes behave like the previous one and have already been marked as `@deprecated`
2. Core classes have been copy-pasted, basically, in order to allow an easier *just-delete-deprecated-classes* in next major
4. TestCases have also been copy-pasted except for key length tests
5. `openssl_sign` error test `::signShouldRaiseAnExceptionWhenKeyIsInvalid` used a short key to tests `CannotSignPayload` exception raise, but now that's blocked earlier and only `UnsafeRsaTest` covers those lines of code. In the next major I guess we'll need to just replace the [`if (! openssl_sign`](https://github.com/lcobucci/jwt/blob/d9de187d03c29f65e68c0d092c20e268b3197024/src/Signer/OpenSSL.php#L38) with an assert, maybe?
6. Doc has been updated too

MT fails because I've been a bit too wary:

```diff
 $details = openssl_pkey_get_details($key);
-        assert(array_key_exists('bits', $details) && is_int($details['bits']));
+        assert(array_key_exists('bits', $details) || is_int($details['bits']));
```
Honestly I don't have enough trust in PHP to delete this.


lcobucci:
Thanks @Slamdunk! I'll review it ASAP (I've been dealing with so many things at work this week 😰)

src/Signer/OpenSSL.php
```diff
@@ -109,6 +113,11 @@ private function validateKey($key): void
         if (! array_key_exists('key', $details) || $details['type'] !== $this->keyType()) {
             throw InvalidKeyProvided::incompatibleKey();
         }
+
+        assert(array_key_exists('bits', $details) && is_int($details['bits']));
```
lcobucci: I think we should/could split this into two

docs/configuration.md
```diff
@@ -109,15 +109,26 @@ $configuration = Configuration::forAsymmetricSigner(
...
-| `EdDSA` | EdDSA signature algorithms      | `\Lcobucci\JWT\Signer\Eddsa`        |
+| Name    | Description                     | Class                               | Key length req. |
+| ------- | ------------------------------- | ----------------------------------- |-----------------|
+| `ES256` | ECDSA using P-256 and SHA-256   | `\Lcobucci\JWT\Signer\Ecdsa\Sha256` | 224 bits        |
+| `ES384` | ECDSA using P-384 and SHA-384   | `\Lcobucci\JWT\Signer\Ecdsa\Sha384` | 224 bits        |
+| `ES512` | ECDSA using P-521 and SHA-512   | `\Lcobucci\JWT\Signer\Ecdsa\Sha512` | 224 bits        |
+| `RS256` | RSASSA-PKCS1-v1_5 using SHA-256 | `\Lcobucci\JWT\Signer\Rsa\Sha256`   | 2048 bits       |
+| `RS384` | RSASSA-PKCS1-v1_5 using SHA-384 | `\Lcobucci\JWT\Signer\Rsa\Sha384`   | 2048 bits       |
+| `RS512` | RSASSA-PKCS1-v1_5 using SHA-512 | `\Lcobucci\JWT\Signer\Rsa\Sha512`   | 2048 bits       |
+| `EdDSA` | EdDSA signature algorithms      | `\Lcobucci\JWT\Signer\Eddsa`        | 256 bits        |
```
lcobucci, line 120: Is this correct? AFAIK EdDSA key pairs are 96 bytes long (32 for the public key and 64 for the secret key), so it should be for 512 bits?

lcobucci (changes requested):
Minor comments, I'll rebase your branch to handle them and we can get it merged.
Thanks for your help!

docs/configuration.md
Slamdunk: As far as I know, "key-pair" length isn't a thing, only private key matters in strength and so it is in key length requirement.

And Ed25519 spec within EdDSA uses 32 bytes for private key.
You often see a 64 bytes private key because the public key is appended to the private one (I have no idea why).

lcobucci: I'm talking based on these constants in PHP sodium:

```php
const SODIUM_CRYPTO_SIGN_PUBLICKEYBYTES = 32;
const SODIUM_CRYPTO_SIGN_SECRETKEYBYTES = 64;
const SODIUM_CRYPTO_SIGN_KEYPAIRBYTES = 96;
```

> You often see a 64 bytes private key because the public key is appended to the private one (I have no idea why).

Isn't that to represent the pair?

test/_keys/Keys.php
```diff
@@ -21,25 +20,29 @@ trait Keys
...
-            'public2_ec512'  => LocalFileReference::file(__DIR__ . '/ecdsa/public2_ec512.key'),
+            'private'        => Key\InMemory::file(__DIR__ . '/ecdsa/private.key'),
+            'private-params' => Key\InMemory::file(__DIR__ . '/ecdsa/private2.key'),
+            'public1'        => Key\InMemory::file(__DIR__ . '/ecdsa/public1.key'),
+            'public2'        => Key\InMemory::file(__DIR__ . '/ecdsa/public2.key'),
+            'public-params'  => Key\InMemory::file(__DIR__ . '/ecdsa/public3.key'),
+            'private_ec512'  => Key\InMemory::file(__DIR__ . '/ecdsa/private_ec512.key'),
+            'public_ec512'   => Key\InMemory::file(__DIR__ . '/ecdsa/public_ec512.key'),
+            'public2_ec512'  => Key\InMemory::file(__DIR__ . '/ecdsa/public2_ec512.key'),
+            'private_short'  => Key\InMemory::file(__DIR__ . '/ecdsa/private_ec160.key'),
```
lcobucci, line 44: This fails to load on my machine... OpenSSL says it's unsupported :thinking: 
I'm running `OpenSSL 3.0.5 5 Jul 2022 (Library: OpenSSL 3.0.5 5 Jul 2022)`:

```
openssl
OpenSSL support => enabled
OpenSSL Library Version => OpenSSL 3.0.5 5 Jul 2022
OpenSSL Header Version => OpenSSL 3.0.3 3 May 2022
Openssl default config => /etc/pki/tls/openssl.cnf
openssl.cafile => no value => no value
openssl.capath => no value => no value
Native OpenSSL support => enabled
```

docs/configuration.md
Slamdunk: https://www.rfc-editor.org/rfc/rfc8032.html#section-5.1.5

> The private key is 32 octets (256 bits, corresponding to b) of cryptographically secure random data.

I've always only used 32 bytes to represent private keys in my production for all `libsodium` related apps, even this one.

:shrug:

test/_keys/Keys.php
lcobucci:
```sh
$ openssl ecparam -list_curves
  secp224r1 : NIST/SECG curve over a 224 bit prime field
  secp256k1 : SECG curve over a 256 bit prime field
  secp384r1 : NIST/SECG curve over a 384 bit prime field
  secp521r1 : NIST/SECG curve over a 521 bit prime field
  prime256v1: X9.62/SECG curve over a 256 bit prime field
```

That explains :smile:

Slamdunk: I guess both `shivanmatur/php` and me run OpenSSL v1.

Can you try `secp192k1` then?

```shell
openssl ecparam -name secp192k1 -genkey -noout -out private_ec192.key
```

Slamdunk: > That explains

Indeed I would have loved to check key bits before loading it with OpenSSL, but that's out of the scope of this library.

lcobucci: OpenSSL v3 removed everything < 224 =)
We can skip the tests for OpenSSL >= 3.0

docs/configuration.md
lcobucci: All good then =)

lcobucci (approved):

lcobucci:
I think we're good for 4.2.0 :tada: 
I'll prepare the changelog and release it in the evening.

@Slamdunk thanks for all your help!

PR #864, opened 2022-08-17
Fix ecdsa key size validation

### LINKED ISSUE #704: Invalid signing with SHA256 alg using secp521r1 curve
Generate key with next command

```bash
/: openssl ecparam -name secp521r1 -genkey -noout -out key.pem
```

With package i use sha256 to sign the payload

```php
use Lcobucci\JWT\Signer\Ecdsa\Sha256;
use Lcobucci\JWT\Signer\Key\InMemory;

$payload = 'data';
$signer = Sha256::create();
$key = InMemory::file('/key.pem');

$sign = $signer->sign($payload, $key);

echo mb_strlen(bin2hex($sig), '8bit'); // by python libraries always 264, here can be 260/262/264

echo $signer->verify($sign, $message, $payload); // must be true, exception throws instead that signature has invalid length
```

But interesting thing, when out signature has length 264, python package validates its successfully, when the $signer always failing it

lcobucci:
Fixes https://github.com/lcobucci/jwt/issues/704


Ocramius:
😱

Slamdunk:
🙈

PR #871, opened 2022-08-19
Require PHP 8.1+

lcobucci:
This bumps all dependencies, removes dependency on mbstring, and requires PHP 8.1+.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -15,8 +15,7 @@ jobs:
         dependencies:
           - "locked"
         php-version:
-          - "7.4"
-          - "8.0"
+          - "8.1"
         operating-system:
           - "ubuntu-latest"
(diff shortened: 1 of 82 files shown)


marcovtwout:
@lcobucci Just wondering, why did you bump the minimum required PHP version to 8.1? PHP 8.0 is still supported by PHP until the end of this year. Are you using any 8.1 specific features?

*nevermind, I see in #764b028 that lots of 8.1 specific features are applied as well.

Ocramius:
@marcovtwout used features visible in this diff: we already rely on `readonly`, where possible :) 

Meanwhile, see https://github.com/Ocramius/.github/blob/c77629415fd5a57ef23d9372075f406e7bf98403/version-support.md#dependency-upgrades

You can expect dependency ranges to always be restricted further, as time passes.

PR #880, opened 2022-08-23
PHP Version constraint: stick to tested versions only

Slamdunk:
PHP doesn't follow SemVer, so we should provide support only for tested PHP versions

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,6 +59,9 @@
             "infection/extension-installer": true,
             "ocramius/package-versions": true
         },
+        "platform": {
+            "php": "8.1.99"
+        },
         "preferred-install": "dist",
         "sort-packages": true
     }
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "php": "^8.1",
+        "php": "~8.1.0",
         "ext-hash": "*",
         "ext-json": "*",
         "ext-openssl": "*",
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "php": "~8.1.0",
+        "php": "~8.1.0 || ~8.2.0",
         "ext-hash": "*",
         "ext-json": "*",
         "ext-openssl": "*",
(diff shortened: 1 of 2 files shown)


Ocramius (changes requested):
LGTM, **but** we can add PHP 8.2 at this stage, since it is stable-ish.

Remember to:

 1. [x] define `config.platform.php` to help dependabot/etc in upgrades
 2. [x] update `composer.lock` to reflect the current `composer.json`

Slamdunk:
> 1. define `config.platform.php` to help dependabot/etc in upgrades

Can you elaborate a bit please? Can't find any doc about it :thinking:

composer.json
```diff
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "php": "~8.1.0",
+        "php": "~8.1.0 || ~8.2.0",
```
Ocramius, line 20: 8.2 should also be in CI before merge

Slamdunk: https://github.com/lcobucci/jwt/blob/2fecef3210030c1bfaead33116228118b7ca9cc0/.github/workflows/phpunit.yml#L77

Ocramius:
> > 1. define `config.platform.php` to help dependabot/etc in upgrades
> 
> Can you elaborate a bit please? Can't find any doc about it thinking

If we declare `{"require": {"php": "~8.1.0 || ~8.2.0"}}`, then @dependabot will pick up packages compatible with PHP 8.2.0 **only** (at some point).

To work around this, we can declare `{"config": {"platform": {"php": "8.1.99"}}}` to force Composer's SAT logic to always pick a dependency set that is compatible with PHP 8.1.x, while running `composer update`.

Slamdunk:
> `{"config": {"platform": {"php": "8.1.99"}}}`

Ah, LoL, yes, I was looking for a `config.platform.php` file :facepalm:

Ocramius (approved):
LGTM, letting @lcobucci call the shots though

lcobucci (approved):
Ship it!

PR #931, opened 2022-10-01
Make test assertion more strict

peter279k:
# Changed log

- It should use the `assertSame` to make assert equals strict.


Ocramius (approved):
LGTM 👍

PR #936, opened 2022-10-11
Tighten `string` to `non-empty-string` where possible

Slamdunk:


src/Encoding/JoseEncoder.php
```diff
@@ -23,10 +24,14 @@ final class JoseEncoder implements Encoder, Decoder
     public function jsonEncode(mixed $data): string
     {
         try {
-            return json_encode($data, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE | JSON_THROW_ON_ERROR);
+            $jsonEncoded = json_encode($data, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE | JSON_THROW_ON_ERROR);
         } catch (JsonException $exception) {
             throw CannotEncodeContent::jsonIssues($exception);
         }
+
+        assert($jsonEncoded !== '');
```
Ocramius, line 32: Hmm, I would expect psalm and phpstan to have better type inference here :thinking:

src/Parser.php
```diff
@@ -12,6 +12,8 @@ interface Parser
     /**
      * Parses the JWT and returns a token
      *
+     * @param non-empty-string $jwt
```
Ocramius, line 15: I think a parser should accept an empty string here, with `CannotDecodeContent` as an outcome :thinking:

src/Signer/Ecdsa.php
```diff
@@ -29,6 +32,8 @@ final public function sign(string $payload, Key $key): string
 
     final public function verify(string $expected, string $payload, Key $key): bool
     {
+        assert($expected !== '');
```
Ocramius, line 35: Wait, is the signature not supposed to be non-empty upfront?

test/unit/Token/ParserTest.php
```diff
@@ -41,7 +41,7 @@ public function parseMustRaiseExceptionWhenTokenDoesNotHaveThreeParts(): void
         $this->expectException(InvalidTokenStructure::class);
         $this->expectExceptionMessage('The JWT string must have two dots');
 
-        $parser->parse('');
+        $parser->parse('.');
```
Ocramius, line 44: IMO OK to parse an empty string and get a crash here.

Ocramius (changes requested):

src/Parser.php
Slamdunk: An empty string JWT is always an error, just like a boolean would be, that's why there's a `string` native type ATM.
Why not preventing also empty string at all?

src/Signer/Ecdsa.php
Slamdunk: Not until we allow `\Lcobucci\JWT\Signer\None` to exist.

`\Lcobucci\JWT\Signer::verify` API must take into account all available signers.

src/Parser.php
Ocramius: Need to think about it, but I see a `CannotDecodeContent` exception as quite valuable on an empty string scenario here: forcing further verification on the client side seems wasteful, at this stage.

I agree that a `string&empty` will always result in a crash here: just wondering whether forcing an additional upfront assertion is added value.

In the following (valid) code:

```php
function foo(string $token) {
    Assert::notEmpty($token);
    $parsed = $this->parser->parse($token);

    // ...
```

If a `CannotDecodeContent` is thrown (explaining that the content was empty) is the same value of `AssertionFailed`, except with an additional piece of code for it.

I'm just trying to figure out the ergonomics of the library here:

 * correctness leads to `string&empty -> never`
 * usability leads to `non-empty-string|(string&empty) -> Token|Exception`

Slamdunk: Now I understand better your doubts.

I can only say the sooner the type is safe, the better:

* if the library allows an empty string, my whole app may allow it
* if the library denies an empty string, I have to design the app upfront to get rid of it. This basically means input validation before consumption

Ocramius: There is one good thing coming from this new type restriction: the existence of the token must be verified by the consumer.

Anyway, I'm just presenting the options: I think that the parser (being a parser) is responsible for upcasting what comes in (or throwing), so it's our decision to decide how wide the API is supposed to be.

The parser makes the type safe anyway, so I have doubts on the value of an early assertion, in this specific case.

Ocramius: That said, we can experiment with this, and then later-on relax the type constraint, if we consider it excessive.

src/Encoding/JoseEncoder.php
SvenRtbg, line 32: I'm not too sure whether or not the code should use a function that is affected by at least 6 configuration values in php.ini and can effectively be disabled (I believe this is even recommended for production environments)

Ocramius: It's still a step up from using `/** @var non-empty-string $jsonEncoded */`.

In fact, some coding standards automatically convert this.

Adding another dependency for this is not worth it.

SvenRtbg: My question is more targeted at: Do you know how this is supposed to affect execution, and is everyone fine if the assertion is disabled - or throwing an exception. And why is this considered better than explicitly terminating execution instead of hoping that echoing a warning message (which seems to be the default configuration for assert) will somehow work better.

SvenRtbg: And if you know how this is affecting execution: How do you document this?
- Is it throwing an exception? Which one? 
- Is it not doing anything at all? Ok.
- Warning message? Quite hard to put this into an interface annotation.
- Anything in between I have not considered? Hm...

Ocramius: IMO not part of this PR's discussion: `assert()` is generally OK for type-upcasting, and well understood by PHPStan and Psalm.

As for whether they affect the runtime: doesn't really matter, since assertions of this kind are supposed to be used when things are known to be true upfront (from a developer PoV), and not as a runtime verification mechanism.

Consider an `assert()` the equivalent of an inline `@var`, but with the ability to be a bit stricter, depending on settings.

SvenRtbg: I just updated my knowledge about assertions. :) PHP 8.0 by default throws an exception (a previously poorly documented change that was merged into PHP without too much discussion - and exactly this was triggering me).

Also I now observed this library already uses `assert()` for existing checks, so this is in line with the previous approach. 👍

Ocramius: BTW, `slevomat/coding-standard` automatically converts inline `@var` into `assert()` :+1:

Slamdunk:
Closed in favor of #939

Issue #937, opened 2022-10-11
Remove `\Lcobucci\JWT\Signer\None`


Slamdunk:
While working on https://github.com/lcobucci/jwt/pull/936 I had the wish to remove the `None` signer entirely, so we can require all signatures to be `non-empty-string`.

Can we, please, so we throw away all possible misuses of that ancient specification?

Slamdunk:
Closed by https://github.com/lcobucci/jwt/pull/939

PR #938, opened 2022-10-11
Deprecate empty Signer, empty Key and empty Signature

Slamdunk:
Addresses https://github.com/lcobucci/jwt/issues/937 for `v4`.

This PR deprecates:

1. `\Lcobucci\JWT\Signer\None`
2. `\Lcobucci\JWT\Signer\Key\InMemory::empty`
3. `\Lcobucci\JWT\Token\Signature::fromEmptyData`
4. and as a result, `\Lcobucci\JWT\Configuration::forUnsecuredSigner`

The likelyhood of them being misused are too high.

If someone needs them, they can implement it by themself.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -105,26 +105,6 @@ $configuration = Configuration::forAsymmetricSigner(
 );
 ```
 
-### For no algorithm
-
-!!! Warning
-    This configuration type is **NOT** recommended for production environments.
-    It's only provided to allow people to have a simpler and faster setup for tests, avoiding any kind of signature creation/verification.
-
-```php
-<?php
-declare(strict_types=1);
-
-use Lcobucci\JWT\Configuration;
-
-require 'vendor/autoload.php';
-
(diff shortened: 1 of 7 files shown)


Ocramius (approved):

Slamdunk:
BTW, yes I know what https://www.rfc-editor.org/rfc/rfc7519#section-8 says:

> Of the signature and MAC algorithms specified in JSON Web Algorithms [[JWA](https://www.rfc-editor.org/rfc/rfc7519#ref-JWA)], only HMAC SHA-256 ("HS256") and "none" MUST be implemented by conforming JWT implementations.

Screw that

Ocramius:
That would certainly make @paragonie happy :+1:

lcobucci:
Very well, let's kill this with fire :fire:

~@Slamdunk would you be able to start a new section "v4.x to v5.x" on [upgrading](https://github.com/lcobucci/jwt/blob/4.3.x/docs/upgrading.md) and mention this?~ I'll handle it... need to do it on another PR too

PR #941, opened 2022-10-13
Add .editorconfig file

chapeupreto:
Add .editorconfig file

See: https://editorconfig.org/


Ocramius:
This is just added metadata to the repo: we already enforce CS where relevant, and yml/md files are really not worth this.

PR #960, opened 2022-11-01
Setup Renovate as a replacement for Dependabot

Ocramius:
Including @Ocramius' opinionated renovate configuration
    
This will set up renovate in a fairly opinionated way:
    
 * auto-upgrades
 * widening PHP version support, when tests pass
 * nightly composer.lock upgrades
 * narrowing dependency ranges regularly

In future, it may be a good idea to copy the entire configuration, to avoid upstream shifts causing issues here,
but that is a decision to be taken during code review here


renovate.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+        "local>Ocramius/.github:renovate-config"
```
Ocramius: Note: I try to maintain this file to the best of my maintenance requirements, which are currently quite aligned with @lcobucci's, but we may need to import the whole DSL here, if preferred.

The advantage is more control, the disadvantage is that it's a lot of changes over time, and a lot of little bells and whistles to configure / learn.

Ocramius: Ah, I just realized that renovate doesn't sign commits :thinking: 

Probably need to enable that...

Ocramius: Other detail that I forgot: this action will create PRs until we fix the BC check action (currently failing at every PR, because `5.0.0` is unreleased and full of BC breaks)

lcobucci: That's awesome, @Ocramius!
Thanks :+1: 

> Ah, I just realized that renovate doesn't sign commits thinking
> Probably need to enable that...

We can drop the signature requirements if they can't handle it.

> Other detail that I forgot: this action will create PRs until we fix the BC check action (currently failing at every PR, because 5.0.0 is unreleased and full of BC breaks)

It's failing because of some changes that use a syntax that seems to be unsupported. I'll ignore them for now: https://github.com/lcobucci/jwt/pull/967/commits/fb0597538c353c1ed2f48068a51d827a02232cb6

lcobucci: I've just tested and renovate is signing commits... I've copied over your template to a repo under my control.

I'll update this one and we can merge it.

PR #979, opened 2022-11-17
Builder: make it immutable

### LINKED ISSUE #977: Builder fluent interface is confusing
Currently the Token Builder is confusing:

https://github.com/lcobucci/jwt/blob/602856eb05a0c8e51f0e21925bdc34af372eb582/src/Token/Builder.php#L76-L91

The name of the method resemble an immutable object, but under the hood it's a fluent interface.
All my colleagues get confused too by reading a method like:

```php
class User
{
    public function fillJwtToken(JwtBuilder $jwtBuilder): void
    {
        $jwtBuilder->withClaim('user-id', $this->id);
    }
}
```
Did the developer forget to return the new object?
Or its more a `setClaim` rather than a `withClaim`?

Can we take a direction and forget the other one?

Slamdunk:
Fix https://github.com/lcobucci/jwt/issues/977

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -72,6 +72,55 @@ If you're using it because it's "fast", please look into adoption the non-standa
 +);
 ```
 
+### `Builder` API is now `@immutable`
+
+`\Lcobucci\JWT\Builder` interface alongside with its default implementation `\Lcobucci\JWT\Token\Builder` are now marked `@immutable`.
+
+If you are using it for example with the `JwtFacade` ensure now to use the returned new `Builder` instance:
+
+```diff
+ $token = (new JwtFacade())->issue(
+     new Sha256(),
+     $key,
+     static function (
+         Builder $builder,
+         DateTimeImmutable $now
+     ): Builder {
+-        $builder->issuedBy('https://api.my-awesome-app.io');
+-        $builder->permittedFor('https://client-app.io');
+-        $builder->expiresAt($now->modify('+10 minutes'));
++        $builder = $builder->issuedBy('https://api.my-awesome-app.io');
++        $builder = $builder->permittedFor('https://client-app.io');
++        $builder = $builder->expiresAt($now->modify('+10 minutes'));
+
+         return $builder;
+     }
+ );
+```
+
+Or:
+
+```diff
+ $token = (new JwtFacade())->issue(
+     new Sha256(),
+     $key,
+-    static function (
++    static fn (
+         Builder $builder,
+         DateTimeImmutable $now
+-    ): Builder {
+-        $builder->issuedBy('https://api.my-awesome-app.io');
+-        $builder->permittedFor('https://client-app.io');
+-        $builder->expiresAt($now->modify('+10 minutes'));
+-
+-         return $builder;
+-     }
++    ): Builder => $builder->issuedBy('https://api.my-awesome-app.io')
++        ->permittedFor('https://client-app.io')
++        ->expiresAt($now->modify('+10 minutes'))
+ );
+```
+
 ## v3.x to v4.x
 
 The `v4.0.0` aggregates about 5 years of work and contains **several BC-breaks**.
--- a/src/Token/Builder.php
+++ b/src/Token/Builder.php
@@ -16,6 +16,7 @@ use function array_diff;
 use function array_merge;
 use function in_array;
 
+/** @immutable */
 final class Builder implements BuilderInterface
 {
     /** @var array<non-empty-string, mixed> */
@@ -68,9 +69,10 @@ final class Builder implements BuilderInterface
 
     public function withHeader(string $name, mixed $value): BuilderInterface
     {
-        $this->headers[$name] = $value;
+        $new                 = clone $this;
+        $new->headers[$name] = $value;
 
-        return $this;
+        return $new;
     }
 
     public function withClaim(string $name, mixed $value): BuilderInterface
@@ -85,9 +87,10 @@ final class Builder implements BuilderInterface
     /** @param non-empty-string $name */
     private function setClaim(string $name, mixed $value): BuilderInterface
     {
-        $this->claims[$name] = $value;
+        $new                = clone $this;
+        $new->claims[$name] = $value;
 
-        return $this;
+        return $new;
     }
 
     /**
--- a/tests/Token/BuilderTest.php
+++ b/tests/Token/BuilderTest.php
@@ -13,6 +13,7 @@ use Lcobucci\JWT\Token\RegisteredClaimGiven;
 use Lcobucci\JWT\Token\RegisteredClaims;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
+use SplObjectStorage;
 
 /**
  * @covers \Lcobucci\JWT\Token\Builder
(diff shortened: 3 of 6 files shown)


src/Token/Builder.php
```diff
@@ -16,6 +16,7 @@
 use function array_merge;
 use function in_array;
 
+/** @immutable */
```
Ocramius, line 19: The interface should probably also be immutable

Ocramius (changes requested):

src/Token/Builder.php
Slamdunk: This idea puzzled me for the past hour.

Of course it must be marked `@immutable` as well, otherwise we could end with different implementations and that could be catastrophic for the consumer.

But, on the other hand, what sense does it make to mark the "*statefulness*" for an interface?

:thinking:

Ocramius: This is a bit of a quirk of phpstan vs psalm.

In psalm, immutable means "doesn't change internal state", plus "no side-effects" :) 

As a consumer of an immutable interface, you know that calling methods on it won't lead to an internal state mutation, which is something that PHP doesn't enforce natively, even with PHP 8.2 readonly classes.

In practice, `@immutable` is expected on both levels, not just in the interface: in fact, psalm will even report this as an issue.

Ocramius: Here's an example of why immutability is important from a consumer/contract PoV: 

```php
<?php

/** readonly */ class ImmutableCounter
{
    public function __construct(private readonly int $count) {}

    public function add1(): self { return new self($this->count + 1); }

    public function value(): int { return $this->count; }
}

class MutableCounter extends ImmutableCounter
{

    public function __construct(private int $count) {}

    public function add1(): self
    {
        return new self(++$this->count);
    }

    public function value(): int { return $this->count; }
}

$counter1 = new ImmutableCounter(0);
$counter2 = $counter1->add1();
$counter3 = $counter2->add1();
$counter4 = $counter3->add1();

var_dump([$counter1->value(), $counter2->value(), $counter3->value(), $counter4->value()]);

$mutableCounter1 = new MutableCounter(0);
$mutableCounter2 = $mutableCounter1->add1();
$mutableCounter3 = $mutableCounter2->add1();
$mutableCounter4 = $mutableCounter3->add1();

var_dump([$mutableCounter1->value(), $mutableCounter2->value(), $mutableCounter3->value(), $mutableCounter4->value()]);
```

```
array(4) {
  [0]=>
  int(0)
  [1]=>
  int(1)
  [2]=>
  int(2)
  [3]=>
  int(3)
}
array(4) {
  [0]=>
  int(1)
  [1]=>
  int(2)
  [2]=>
  int(3)
  [3]=>
  int(3)
}
```

https://3v4l.org/3Ocom

Ocramius: Note: I re-used this example to state my point **against** the current state of https://wiki.php.net/rfc/readonly_amendments

Ocramius (approved):

src/Token/Builder.php
bdsl: My understanding is that the RFC wouldn't allow making the `count` property mutable in a child of a readonly class. I think they're just meaning to allow a mutable child of an readonly class to declare **additional** properties, which would be mutable. Not change the ones that are already there. Seems reasonable to me.

Ocramius: @bdsl the properties are completely irrelevant: the behavior changes based on the public API of the object, which is no longer immutable.

I suggest you discuss this further in https://externals.io/message/119007#119016 - here I only explained why the immutability has to be declared at interface level, and respected in all implementations.

bdsl: I wasn't able to get signup for that mailing list working with my email last time I tried, so I'll leave it for now.

lcobucci (commented):
Implementation looks good and it's great no signature had to be changed.

Let's document the change (also on the upgrading guide) and it's good to go. Thanks for helping out!

docs/upgrading.md
```diff
@@ -72,6 +72,32 @@ If you're using it because it's "fast", please look into adoption the non-standa
...
+     static function (
+         Builder $builder,
+         DateTimeImmutable $now
+     ): Builder {
+-        $builder->issuedBy('https://api.my-awesome-app.io');
+-        $builder->permittedFor('https://client-app.io');
+-        $builder->expiresAt($issuedAt->modify('+10 minutes'));
++        $builder = $builder->issuedBy('https://api.my-awesome-app.io');
++        $builder = $builder->permittedFor('https://client-app.io');
++        $builder = $builder->expiresAt($now->modify('+10 minutes'));
```
Slamdunk, line 94: I wouldn't format my code like this, I made it so here to force the user read the diff and think about it.

tests/Token/BuilderTest.php
```diff
@@ -104,4 +105,32 @@ public function getTokenShouldReturnACompletelyConfigureToken(): void
...
+        $builder       = $builder->relatedTo('subject');
+        $map[$builder] = true;
+        $builder       = $builder->permittedFor('test1');
+        $map[$builder] = true;
+        $builder       = $builder->withClaim('test', 123);
+        $map[$builder] = true;
+        $builder       = $builder->withHeader('userId', 2);
+        $map[$builder] = true;
+
+        self::assertCount(10, $map);
```
Ocramius, line 134: Ha, fancy way to check immutability :D

Ocramius (approved):

docs/upgrading.md
lcobucci: Maybe we can add a note about that on the upgrading guide?

lcobucci:
I'll run the BC-break check locally against my PR that implements baseline to it and I believe we can merge this soon.

Since we're making the builder immutable, there's also no reason for having the "builder factory" on the configuration object. I'd say we can easily deprecate it on v5 and remove it on v6. I'll implement that on a separate PR.

docs/upgrading.md
Slamdunk: About what exactly?

lcobucci: That the format only illustrates the immutability factor and we don't expect people to code like that

lcobucci:
> I'll run the BC-break check locally against my PR that implements baseline to it and I believe we can merge this soon.

```sh
$ ../../ocramius/bc-check/bin/roave-backward-compatibility-check --from=origin/5.0.x
## 🇷🇺 Российским гражданам

В Украине сейчас идет война. Силами РФ наносятся удары по гражданской инфраструктуре в [Харькове][1], [Киеве][2], [Чернигове][3], [Сумах][4], [Ирпене][5] и десятках других городов. Гибнут люди - и гражданское население, и военные, в том числе российские призывники, которых бросили воевать. Чтобы лишить собственный народ доступа к информации, правительство РФ запретило называть войну войной, закрыло независимые СМИ и принимает сейчас ряд диктаторских законов. Эти законы призваны заткнуть рот всем, кто против войны. За обычный призыв к миру сейчас можно получить несколько лет тюрьмы.

Не молчите! Молчание - знак вашего согласия с политикой российского правительства.  
**Вы можете сделать выбор НЕ МОЛЧАТЬ.**

---

## 🇺🇸 To people of Russia

There is a war in Ukraine right now. The forces of the Russian Federation are attacking civilian infrastructure in [Kharkiv][1], [Kyiv][2], [Chernihiv][3], [Sumy][4], [Irpin][5] and dozens of other cities. People are dying – both civilians and military servicemen, including Russian conscripts who were thrown into the fighting. In order to deprive its own people of access to information, the government of the Russian Federation has forbidden calling a war a war, shut down independent media and is passing a number of dictatorial laws. These laws are meant to silence all those who are against war. You can be jailed for multiple years for simply calling for peace.

Do not be silent! Silence is a sign that you accept the Russian government's policy.  
**You can choose NOT TO BE SILENT.**

[1]: <https://cloudfront-us-east-2.images.arcpublishing.com/reuters/P7K2MSZDGFMIJPDD7CI2GIROJI.jpg> "Kharkiv under attack"
[2]: <https://gdb.voanews.com/01bd0000-0aff-0242-fad0-08d9fc92c5b3_cx0_cy5_cw0_w1023_r1_s.jpg> "Kyiv under attack"
[3]: <https://ichef.bbci.co.uk/news/976/cpsprodpb/163DD/production/_123510119_hi074310744.jpg> "Chernihiv under attack"
[4]: <https://www.youtube.com/watch?v=8K-bkqKKf2A> "Sumy under attack"
[5]: <https://cloudfront-us-east-2.images.arcpublishing.com/reuters/K4MTMLEHTRKGFK3GSKAT4GR3NE.jpg> "Irpin under attack"

#StandWithUkraine


Using "/home/lcobucci/projects/libs/jwt/.roave-backward-compatibility-check.json" as configuration file
Comparing from a7f5ad6d29ebf86dd639ed4e70112c9da7df5c5e to d53d7a5c4344277f4bb43fd17b496596b7dcd0f0...
Installing dependencies from lock file
Verifying lock file contents can be installed on current platform.
Package operations: 2 installs, 0 updates, 0 removals
  - Installing psr/clock (1.0.0): Extracting archive
  - Installing lcobucci/clock (3.0.0): Extracting archive
1 package you are using is looking for funding.
Use the `composer fund` command to find out more!
No security vulnerability advisories found
Installing dependencies from lock file
Verifying lock file contents can be installed on current platform.
Package operations: 2 installs, 0 updates, 0 removals
  - Installing psr/clock (1.0.0): Extracting archive
  - Installing lcobucci/clock (3.0.0): Extracting archive
1 package you are using is looking for funding.
Use the `composer fund` command to find out more!
No security vulnerability advisories found
No backwards-incompatible changes detected
```

LGTM :+1:

lcobucci (approved):

PR #986, opened 2022-12-19
Allow `lcobucci/clock:^3.0` to be installed with `lcobucci/jwt:^4`

Ocramius:
As per #985 discussions, allowing `lcobucci/clock:^3` in `lcobucci/jwt:^4` will likely ease the upgrade path a bit

### DIFF
--- a/.github/workflows/composer-json-lint.yml
+++ b/.github/workflows/composer-json-lint.yml
@@ -29,7 +29,7 @@ jobs:
           coverage: "none"
           php-version: "${{ matrix.php-version }}"
           ini-values: memory_limit=-1
-          tools: composer:v2, composer-normalize, composer-require-checker, composer-unused
+          tools: composer:v2, composer-normalize:2.28.3, composer-require-checker, composer-unused
 
       - name: Get composer cache directory
         id: composer-cache
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "ext-mbstring": "*",
         "ext-openssl": "*",
         "ext-sodium": "*",
-        "lcobucci/clock": "^2.0"
+        "lcobucci/clock": "^2.0 || ^3.0"
     },
     "require-dev": {
         "infection/infection": "^0.21",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "22fe9a62505921a50cddf66b788a67aa",
+    "content-hash": "22de66f8a38ac3f7a0590a76c72e97b8",
     "packages": [
         {
             "name": "lcobucci/clock",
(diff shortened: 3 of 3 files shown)


Ocramius:
Not sure why `normalize composer.json` fails:

```
Box Requirements Checker
========================

> Using PHP 7.4.33
> PHP is using the following php.ini file:
  /etc/php/7.4/cli/php.ini

> Checking Box requirements:
  E......

                                                                                
Error: ] Your system is not ready to run the application.  
```

lcobucci:
Is it me or this failure makes no sense?

Ocramius:
Perhaps `Box` is running on a newer PHP version, or installing with `--ignore-platform-reqs`?

Ocramius:
Also: how is Box used on this project? :thinking:

lcobucci:
> Also: how is Box used on this project? :thinking: 

Nope... Is it a dependency of that tool?

Slamdunk:
Also, what's going on with [`Backwards compatibility check`](https://github.com/lcobucci/jwt/actions/runs/3735016533/jobs/6337752703#logs)?

```
fatal: detected dubious ownership in repository at '/github/workspace'
To add an exception for this directory, call:

	git config --global --add safe.directory /github/workspace
```

Ocramius:
Ah, we run it in a docker container: that may be a problem, if ownership isn't the same :-\

Since Git 1.25 or such, a lot of added security "noise" has been added to git, so running mixed container setups is very noisy.

This can be mitigated by installing `roave/backward-compatibility-check` as its own tool in its own directory :thinking:

smoench:
`composer-normalize` has dropped support for PHP 7.4. in [the latest release](https://github.com/ergebnis/composer-normalize/releases/tag/2.29.0). The github workflow runs with PHP 7.4 and doesn't use [a fixed version here](https://github.com/lcobucci/jwt/blob/4.3.x/.github/workflows/composer-json-lint.yml#L32).

This line may needs be changed to `tools: composer:v2, composer-normalize:2.8.3, composer-require-checker, composer-unused`.

Slamdunk:
Thank you @smoench, I've locked the [`2.28.3`](https://github.com/ergebnis/composer-normalize/releases/tag/2.28.3) (not `2.8.3` :wink:) and indeed now the build goes green again.

Ocramius:
About the BC checker stuff: should we ignore it?

Slamdunk:
Yes and No: yes for this PR, but we should still fix its run.

Even with BC-Check run fixed it will still fail here, so as `Collaborator`s we can't merge this, we don't have enough permissions.
Need to wait @lcobucci

lcobucci:
> Yes and No: yes for this PR, but we should still fix its run.
> 
> Even with BC-Check run fixed it will still fail here, so as `Collaborator`s we can't merge this, we don't have enough permissions.
> Need to wait @lcobucci

That sucks, it's configured not to be required 😥

Merging...

Slamdunk:
Thanks @lcobucci: can I close [`4.3.0`](https://github.com/lcobucci/jwt/milestone/43) milestone so it gets released?

Ocramius:
Thanks meanwhile, y'all!

lcobucci:
> Thanks @lcobucci: can I close [`4.3.0`](https://github.com/lcobucci/jwt/milestone/43) milestone so it gets released?

Feel free to handle it, my connection is a bit meh

Ocramius:
Done

PR #1001, opened 2023-02-21
Support other PSR-20 implementations

lcobucci:
This promotes better interoperability by allowing users to plug any Clock implementation that follows PSR-20 - we still fallback to lcobucci/clock.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,8 @@
         "ext-json": "*",
         "ext-openssl": "*",
         "ext-sodium": "*",
-        "lcobucci/clock": "^3.0.0"
+        "lcobucci/clock": "^3.0.0",
+        "psr/clock": "^1.0"
     },
     "require-dev": {
         "infection/infection": "^0.26.19",
--- a/composer.json
+++ b/composer.json
@@ -22,11 +22,11 @@
         "ext-json": "*",
         "ext-openssl": "*",
         "ext-sodium": "*",
-        "lcobucci/clock": "^3.0.0",
         "psr/clock": "^1.0"
     },
     "require-dev": {
         "infection/infection": "^0.26.19",
+        "lcobucci/clock": "^3.0",
         "lcobucci/coding-standard": "^9.0",
         "phpbench/phpbench": "^1.2.8",
         "phpstan/extension-installer": "^1.2",
@@ -35,6 +35,9 @@
         "phpstan/phpstan-phpunit": "^1.3.8",
         "phpstan/phpstan-strict-rules": "^1.5.0",
         "phpunit/phpunit": "^10.0.12"
+    },
+    "suggest": {
+        "lcobucci/clock": ">= 3.0"
     },
     "autoload": {
         "psr-4": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,72 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "6213231e73ab90ed5ac862e518dbee2b",
+    "content-hash": "c0d16b3be9c2f7b5b9d81fb835ad6d0f",
     "packages": [
-        {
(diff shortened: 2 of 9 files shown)


composer.json
```diff
@@ -22,7 +22,8 @@
         "ext-json": "*",
         "ext-openssl": "*",
         "ext-sodium": "*",
-        "lcobucci/clock": "^3.0.0"
+        "lcobucci/clock": "^3.0.0",
+        "psr/clock": "^1.0"
```
Slamdunk: I don't like adhering to an interoperability standard AND require one of its implementations.
It sort of defeats the goal of a PSR to me: if every library does this, my project quickly ends up with all the possible implementations installed instead of just the one I prefer.

As an end user, I **like** when a dependency upgrades to a new API that now requires a third dependency previously hidden in an optional argument.

It happens to me a lot of time with the cache PSR. I want all my dependencies to use my cache implementation as it's a really sensitive topic, but some always require it as an optional argument and silently fall back to a custom implementation, so instead of my test suite failing after the upgrade because of a missing required parameter, I have to dig into ALL of them to be sure I'm not letting any other cache implementation slipping in production.
It drives me crazy :roll_eyes: 

We only need it in the `JwtFacade`, I propose to swap the parameter order for this new release and drop the nullability for `Clock`:

```diff
diff --git a/src/JwtFacade.php b/src/JwtFacade.php
index 3a08291..1a1c780 100644
--- a/src/JwtFacade.php
+++ b/src/JwtFacade.php
@@ -5,8 +5,6 @@ namespace Lcobucci\JWT;
 
 use Closure;
 use DateTimeImmutable;
-use Lcobucci\Clock\Clock;
-use Lcobucci\Clock\SystemClock;
 use Lcobucci\JWT\Encoding\ChainedFormatter;
 use Lcobucci\JWT\Encoding\JoseEncoder;
 use Lcobucci\JWT\Signer\Key;
@@ -14,18 +12,16 @@ use Lcobucci\JWT\Validation\Constraint;
 use Lcobucci\JWT\Validation\SignedWith;
 use Lcobucci\JWT\Validation\ValidAt;
 use Lcobucci\JWT\Validation\Validator;
+use Psr\Clock\ClockInterface as Clock;
 
 use function assert;
 
 final class JwtFacade
 {
-    private readonly Clock $clock;
-
     public function __construct(
+        private readonly Clock $clock,
         private readonly Parser $parser = new Token\Parser(new JoseEncoder()),
-        ?Clock $clock = null,
     ) {
-        $this->clock = $clock ?? SystemClock::fromSystemTimezone();
     }
 
     /** @param Closure(Builder, DateTimeImmutable):Builder $customiseBuilder */
```

Slamdunk (changes requested):

composer.json
lcobucci: That's what I want to build up to but I don't want to do both changes in the same major release.

I'll update the commit message to reflect the idea.

Slamdunk: > I don't want to do both changes in the same major release.

May I ask you why not?

Ocramius: @Slamdunk we can remove the dependency in a minor release, by providing a tiny implementation in here, perhaps?

```php
$this->clock = $clock ?? new class implements Clock { ... };
```

lcobucci: > May I ask you why not?

I don't think that the method signature change is worth to be done on this release (I'm aiming at shipping v5.0 this week).
However, I didn't think of @Ocramius' suggestion when I wrote the comment and it does not require changing the signature.

With it, I'm more than happy to move `lcobucci/clock` to `suggested` and fallback to an anonymous class that always returns `new DateTimeImmutable()`.

lcobucci: Alright, eb8bca537e448326698adc11be57cc626e71cb7a keeps it on `require-dev` but now users can decide whether or not `lcobucci/clock` should be on their `vendor`.

Thanks!

Ocramius (approved):
LGTM 👍

Slamdunk (approved):

PR #1002, opened 2023-03-07
Improve benchmarks

### LINKED ISSUE #525: Create benchmarks for token parsing and issuing
Which will be extremely helpful for us to analyse performance impacts.

lcobucci:
Fix #525 

This completely redesigns the benchmark suite and adds new items to analyse the performance impact of token issuing/parsing.


Ocramius (approved):
Overall OK with redesigning this, just unsure whether we'd lose our ability to compare this to 4.x performance.

lcobucci:
> Overall OK with redesigning this, just unsure whether we'd lose our ability to compare this to 4.x performance.
> 
> 

I don't think we lose it, it'd require tinkering with the output, though. 

This can be done in 4.x, the bench suite hasn't changed for quite some time. I'm just not sure if it's worth it

Slamdunk (approved):

PR #1011, opened 2023-04-22
Added support for multiple keys in SignedWith validation

rhertogh:
In order to support key rotation, this pull request adds the possibility to pass multiple keys to the SignedWith validator to verify against.

### DIFF
--- a/src/Validation/Constraint/SignedWith.php
+++ b/src/Validation/Constraint/SignedWith.php
@@ -9,10 +9,19 @@ use Lcobucci\JWT\UnencryptedToken;
 use Lcobucci\JWT\Validation\ConstraintViolation;
 use Lcobucci\JWT\Validation\SignedWith as SignedWithInterface;
 
+use function is_array;
+
 final class SignedWith implements SignedWithInterface
 {
-    public function __construct(private readonly Signer $signer, private readonly Signer\Key $key)
+    private readonly Signer\Key|array $keys;
+
+    public function __construct(private readonly Signer $signer, Signer\Key|array $keys)
     {
+        if (! is_array($keys)) {
+            $keys = [$keys];
+        }
+
+        $this->keys = $keys;
     }
 
     public function assert(Token $token): void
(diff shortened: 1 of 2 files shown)


rhertogh:
@lcobucci Hi, could you please approve the workflows?

Slamdunk (changes requested):
Changing the current constraint would weaken the library, I suggest to propose something like a new `AtLeastOne` constraint which accepts multiple `SignedWith` constraints.

lcobucci:
> Changing the current constraint would weaken the library, I suggest to propose something like a new `AtLeastOne` constraint which accepts multiple `SignedWith` constraints.

Let's also wrap the array. JWK has the concept of key sets, we can use this as opportunity to introduce the it.

lcobucci:
> @lcobucci Hi, could you please approve the workflows?

@rhertogh you may also run everything locally with `make`

rhertogh:
@Slamdunk @lcobucci 
> > Changing the current constraint would weaken the library, I suggest to propose something like a new `AtLeastOne` constraint which accepts multiple `SignedWith` constraints.
> 
> Let's also wrap the array. JWK has the concept of key sets, we can use this as opportunity to introduce the it.

Using a "JWK Set" does indeed make sense.
However, in that case we're passing only 1 thing (the entire set) to the constraint. But when going that path, `AtLeastOne`, which would accept multiple `SignedWith`, sounds contradicting to me.

I see two options:
1. Have `SignedWith` accept `Key|KeySet`
2. Have `SignedWith` only accept `Key` and add a second constraint like `SignedWithOneInSet` which then accepts a `KeySet`

Option 1 makes more sense to me, since it's still validating against either the Key or KeySet. And since the purpose of the KeySet is to be a list of keys (e.g. key rotation and/or different algorithms) there is no risk of confusion or misuse.
 
 @Slamdunk @lcobucci What are your thoughts on this?

lcobucci:
@rhertogh many apologies for my delay... doing OSS has been a little challenging lately.

> And since the purpose of the KeySet is to be a list of keys (e.g. key rotation and/or different algorithms) there is no risk of confusion or misuse.

I'm not sure if I understood what you mean...
A JWK key does point to an algorithm and a key set may contain keys with different algorithms.

Going with option 1 might lead us to abstractions that don't match the real-life possibilities.

Perhaps forgetting about the key set for now and having something like the implementation bellow is indeed the most flexible solution for now:

```php
use Lcobucci\JWT\Validation\SignedWith as SignedWithInterface;

final class SignedWithOneInSet implements SignedWithInterface
{
    /** @var list<SignedWithInterface> */
    private readonly array $constraints;
    
    public function __construct(SignedWithInterface ...$constraints)
    {
        $this->constraints = $constraints;
    }

    public function assert(object $token): void
    {
        // (...)
    }
}
```

lcobucci:
@rhertogh I'll close this one in favour of https://github.com/lcobucci/jwt/pull/1033

Thanks for your help!

PR #1019, opened 2023-08-02
Feature: Add DataSet and Signature interfaces

panigrc:
With this PR the  `Lcobucci\JWT\Token\Dataset` and the `Lcobucci\JWT\Token\Signature` classes will implement the  `Lcobucci\JWT\Dataset` and the `Lcobucci\JWT\Signature` interface respectively.

In that way we will be able to mock the above classes (since they are final), but also implement the interfaces for a custom implementation.

### DIFF
--- a/src/DataSet.php
+++ b/src/DataSet.php
@@ -1,0 +1,18 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT;
+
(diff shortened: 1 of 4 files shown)


Ocramius (changes requested):
There is no need to mock thesr: use the real thing.

PR #1026, opened 2023-09-13
Support for setting claims with an array

hackman:
In order to make my code cleaner I needed a way to set multiple claims by providing them in an array format. 

This implementation expands the withClaim() by providing this interface. 
Any previously defined claims are protected by the same check that withClaim() is doing.

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -74,6 +74,15 @@ interface Builder
     public function withClaim(string $name, mixed $value): Builder;
 
     /**
+     * Configures multiple claim items
+     *
+     * @param non-empty-array $claims
+     *
+     * @throws RegisteredClaimGiven When trying to set a registered claim.
+     */
+    public function withClaim(array $claims): Builder;
+
+    /**
      * Returns a signed token to be used
      *
      * @throws CannotEncodeContent When data cannot be converted to JSON.
--- a/src/Token/Builder.php
+++ b/src/Token/Builder.php
@@ -84,6 +84,18 @@ final class Builder implements BuilderInterface
         return $this->setClaim($name, $value);
     }
 
+	public function withClaims(array $claims): BuilderInterface
+	{
+		$new = clone $this;
+		foreach ($claims as $key => $value) {
+			if (in_array($key, RegisteredClaims::ALL, true)) {
+				throw RegisteredClaimGiven::forClaim($key);
+			}
+			$new->claims[$key] = $value;
+		}
+		return $new;
+	}
+
     /** @param non-empty-string $name */
     private function setClaim(string $name, mixed $value): BuilderInterface
     {


src/Builder.php
```diff
@@ -73,6 +73,15 @@ public function withHeader(string $name, mixed $value): Builder;
...
     public function withClaim(string $name, mixed $value): Builder;
 
+    /**
+     * Configures multiple claim items
+     *
+     * @param non-empty-array $claims
+     *
+     * @throws RegisteredClaimGiven When trying to set a registered claim.
+     */
+    public function withClaim(array $claims): Builder;
```
Ocramius, line 83: Adding methods to an interface is a major breakage, and it must **REALLY** be worth doing. For utility methods, BC breaks are simply not worth pursuing.

This can be done in userland with an `array_reduce()` or similar, if really needed.

Ocramius (changes requested):
This won't be implemented

src/Builder.php
hackman: Can you please give me an example, how one can setup multiple claims with a for loop in case you have this code:
```
    $tokenBuilder   = new Builder(new JoseEncoder(), ChainedFormatter::default());
    $signingKey     = InMemory::file(__DIR__ . $config['api_key']);
    $id                     = getTokenId();
    $now                = new DateTimeImmutable();
    $token             = $tokenBuilder
        ->issuedBy($config['issued_by'])
        ->permittedFor($config['permitted_for'])
        ->identifiedBy($id)
        ->issuedAt($now)
        ->canOnlyBeUsedAfter($now)
        ->expiresAt($now->modify('+1 hour'))
        ->withHeader('module', $module)
        ->withClaims($claims)
        ->getToken(new Sha256(), $signingKey);
```
You can not add them to $tokenBuilder prior or after generating the token. So want to understand how can I implement it without adding withClaims()?

I'm using Signer\Ecdsa\Sha256 if that is of any difference.

Ocramius: pseudo-code, but it's this:


```php
foreach ($claims as $name => $claim) {
    $builder = $builder->withClaim($name, $claim);
}
```

hackman: thank you!

lcobucci: @hackman you can also use our high-level API to simplify your code: https://lcobucci-jwt.readthedocs.io/en/stable/quick-start/#issuing-tokens

```php
$token = (new JwtFacade())->issue(
    new Sha256(),
    InMemory::file(__DIR__ . $config['api_key']),
    static function (Builder $builder, DateTimeImmutable $issuedAt) use ($config): Builder {
        foreach (['claim1' => 'value1'] as $claim => $value) {
            $builder = $builder->withClaim($claim, $value);
        }

        return $builder
            ->issuedBy($config['issued_by'])
            ->permittedFor($config['permitted_for'])
            ->identifiedBy(getTokenId())
            ->expiresAt($issuedAt->modify('+1 hour')) // default is 5 minutes
            ->withHeader('module', $module);
    },
);
```

(the closure can also be replaced by a reference to a function/method using the [first-class callable syntax](https://www.php.net/manual/en/functions.first_class_callable_syntax.php))

PR #1033, opened 2023-11-04
Add support to validate against multiple keys

lcobucci:
This implements the design I suggested on https://github.com/lcobucci/jwt/pull/1011#issuecomment-1657282254 to solve the limitation raised by @rhertogh

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,9 @@
+version: 2
+
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3"
+
+mkdocs:
+  configuration: mkdocs.yml
--- a/docs/rotating-keys.md
+++ b/docs/rotating-keys.md
@@ -1,0 +1,216 @@
+# Rotating Keys
+
+Key rotation consists in retiring and replacing cryptographic keys with new ones.
+Performing that operation on a regular basis is an industry standard.
+
+## Why should I rotate my keys?
+
+Rotating keys allows us to:
+
+1. Limit the number of tokens signed with the same key, helping the prevention of attacks enabled by cryptanalysis
+2. Adopt other algorithms or stronger keys
+3. Limit the impact of eventual compromised keys
+
+## The challenges
+
+After rotating keys, apps will likely receive requests with tokens issues with the previous key.
+If the key rotation of an app is done with a "hard cut", requests with non-expired tokens issued with the old key **will fail**!
+
+Imagine if you were the user who logged in just before a key rotation on that kind of app, you'd probably have to log in again!
+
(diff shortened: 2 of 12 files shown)


.readthedocs.yaml
```diff
@@ -0,0 +1,9 @@
+version: 2
```
lcobucci, line 1: Squeezing this in because I totally missed the [deprecation announcement](https://blog.readthedocs.com/migrate-configuration-v2/)

docs/rotating-keys.md
```diff
@@ -0,0 +1,211 @@
...
++               'GOu4rLyVCBxmxP+sbniU68ojAja5PkRdvv7vNvBCqDQ='
++           ),
++       ),
++       new Constraint\SignedWith(
++            new Signer\Hmac\Sha256(),
++           InMemory::plainText(
++                'a-very-long-and-secure-key-that-should-actually-be-something-else'
++           ),
++       ),
++    ),
```
Slamdunk, line 203: Correct me if I'm wrong, the only foreseeable valid use for `SignedWithOneInSet` is key rotation.

If so, I would like to protect the developer from forgetting to delete the old key once all the tokens signed with that key have expired.

For example, I'd like to see a bond between each key and a time constraint:

1. `SignedWith(Sha256()) ` -> valid until `2023-11-30 00:00:00`
1. `SignedWith(Blake2b())` -> valid until `2025-11-30 00:00:00`

Not sure about the UX and implementation details, but please lets avoid adding code in production that gets forgotten leaving a possible weakness opened forever

Slamdunk (changes requested):

docs/rotating-keys.md
lcobucci: That's a good point, thanks! 
I need to think about the interface here... A possibility would be defining an expiration date to the key but am not sure.

lcobucci: Creating that bond complicates the design a little bit, especially because we don't want to have it as an optional thing.

I'll put an intermediate object to represent that set and see where it takes us.

lcobucci: @Slamdunk I've pushed [26960f1](https://github.com/lcobucci/jwt/pull/1033/commits/26960f17cc8040dea553b3fb9fa6c02d86ea2342) with the experimental design.

What are your thoughts?

Slamdunk: Awesome :muscle: 

Only doc is missing the update, then we can go :+1:

Slamdunk (approved):

docs/rotating-keys.md
lcobucci: Done!

PR #1051, opened 2024-03-21
Make ext-sodium an optional dependency

ausi:
It looks to me like `ext-sodium` is only needed for the two signers `BLAKE2B` and `EdDSA`. 

It would be great if we could move `"ext-sodium": "*"` from `"require"` to `"suggest"` in order to make this library installable even if the Sodium extension is not enabled. This should be fine for all usages of this library that do not use `BLAKE2B` or `EdDSA` which seems to be the case quite often.

Is the 4.3.x branch the correct one for this pull request?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,6 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "ext-openssl": "*",
-        "ext-sodium": "*",
         "lcobucci/clock": "^2.0 || ^3.0"
     },
     "require-dev": {
@@ -37,6 +36,9 @@
         "phpstan/phpstan-strict-rules": "^1.0",
         "phpunit/php-invoker": "^3.1",
         "phpunit/phpunit": "^9.5"
+    },
+    "suggest": {
+        "ext-sodium": "*"
     },
     "autoload": {
         "psr-4": {
--- a/src/Signer/Blake2b.php
+++ b/src/Signer/Blake2b.php
@@ -5,6 +5,7 @@ namespace Lcobucci\JWT\Signer;
 
 use Lcobucci\JWT\Signer;
 
+use function function_exists;
 use function hash_equals;
 use function sodium_crypto_generichash;
 use function strlen;
@@ -20,6 +21,10 @@ final class Blake2b implements Signer
 
     public function sign(string $payload, Key $key): string
     {
+        if (! function_exists('sodium_crypto_generichash')) {
+            throw ExtSodiumMissing::forBlake2b();
+        }
+
         $actualKeyLength = 8 * strlen($key->contents());
 
         if ($actualKeyLength < self::MINIMUM_KEY_LENGTH_IN_BITS) {
--- a/src/Signer/Eddsa.php
+++ b/src/Signer/Eddsa.php
@@ -6,6 +6,7 @@ namespace Lcobucci\JWT\Signer;
 use Lcobucci\JWT\Signer;
 use SodiumException;
 
+use function function_exists;
 use function sodium_crypto_sign_detached;
 use function sodium_crypto_sign_verify_detached;
 
@@ -18,6 +19,10 @@ final class Eddsa implements Signer
 
     public function sign(string $payload, Key $key): string
     {
+        if (! function_exists('sodium_crypto_sign_detached')) {
+            throw ExtSodiumMissing::forEddsa();
+        }
+
         try {
             return sodium_crypto_sign_detached($payload, $key->contents());
         } catch (SodiumException $sodiumException) {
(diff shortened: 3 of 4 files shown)


SvenRtbg:
There is no such thing as an "optional" dependency. The library contains code that needs the Sodium extension. If the extension isn't present, this part of the library will crash if used.

And furthermore, the Sodium extension should be available to the recent PHP versions anyways, it was introduced with PHP 7.2.

Ocramius:
Cleared up by @SvenRtbg

fritzmg:
> There is no such thing as an "optional" dependency.

Sure there is. See [`symfony/password-hasher`](https://github.com/symfony/password-hasher/blob/0eba656c16ecdf5588b3ddd2b2337b06173d839f/Hasher/SodiumPasswordHasher.php#L34-L36) for example. `symfony/password-hasher` has no hard dependency on `ext-sodium`, but you can make use of one of its components in case you have it.

SvenRtbg:
That example illustrates what I said: When trying to use the part needing Sodium, also symfony/password-hasher crashes. It does so using a nicer exception, and it has a nicer message that may be considered helpful, but you have to install Sodium to get access to that part no matter what.

And you will only discover that during runtime! That's the bad part. Composer will complain about the missing extension right away, so you can consider your options before investing quite some time integrating the library.

ausi:
As the library also includes a `SodiumBase64Polyfill` it seemed to me that it is at least somewhat tolerated to be used without the PHP extension. There are also several issue comments that suggest the use of `--ignore-platform-req=ext-sodium` or `config.platform`. So I hoped we could have some leeway here ☺️

> There is no such thing as an "optional" dependency.

I agree. In an ideal world the two signer classes are moved to its own composer package and the extension is only required there, but I think the maintenance burden is not worth it.

> And furthermore, the Sodium extension should be available to the recent PHP versions anyways

I agree as well, but as we maintain an open source CMS we have to deal with the real world of “hosting providers” which is very ugly in large parts, sadly.

Ocramius:
> we have to deal with the real world of “hosting providers” which is very ugly in large parts, sadly.

This is fine, but "we have to deal" does not mean "push the problem down to library maintainers and library consumers" 😁

The restriction is intentional here: less problems we don't want to deal with in first place, and better guarantees for consumers too.

ausi:
> This is fine, but "we have to deal" does not mean "push the problem down to library maintainers and library consumers" 😁

This was not my intention. I just interpreted the existence of `SodiumBase64Polyfill` and comments like _“we have a polyfill for the critical functions”_ as that the maintainers are already handling the case of the extension being missing.

> The restriction is intentional here

That was the point I did not get from reading the code and some issues.
So in that case, this can stay closed and we have to find an alternative solution. 😞

SvenRtbg:
The `SodiumBase64Polyfill` was added with the commit message "Provide Sodium base64 polyfill for libsodium < 1.0.14". Sounds like it is there to allow all possible Sodium versions to work.

leofeyer:
> There is no such thing as an "optional" dependency.

No hard feelings, but this statement is not correct.

| Signer | Uses sodium functions |
| --- | --- |
| BLAKE2B | yes |
| EcDSA | no |
| EdDSA | yes |
| HMAC | no |
| OpenSSL | no |
| RSA | no |

So unless you use the BLAKE2B or the EdDSA signer, your library works **perfectly fine** with the `SodiumBase64Polyfill` and without `ext-sodium`. You could make this even more user-friendly by suggesting either `ext-sodium` or `paragonie/sodium_compat` like [firebase/php-jwt](https://packagist.org/packages/firebase/php-jwt) does:



> and better guarantees for consumers too.

All our users are now forced to have the PHP `ext-sodium` extension, even though we only ever use the HMAC signer. I don‘t see any added value by "better guarantees" here. 🤷‍♂️

Ocramius:
> All our users are now forced to have the PHP `ext-sodium` extension, even though we only ever use the HMAC signer. I don‘t see any added value by "better guarantees" here. 🤷‍♂️

It's "it works" vs "it works because it's declared as a requirement", I suppose.

Just because something works, doesn't mean that it's right :-P 

The fact that you use a portion of the library is certainly a detail: the library is installed, and it should be working in its totality (or be split), which is what @SvenRtbg explained above. Your users may be using more of it.

The requirement for `ext-sodium` is not going away: it's part of PHP itself since 7.2.0, and we're now enforcing PHP 8.1.0 or newer on 5.x.

Unless 4.x is affected by a security issue, leave it alone.

In fact, removing the requirement can cause **more issues** for downstream installations relying on `composer.lock` (heroku-alike deployments and extension discovery), which could potentially fail catastrophically in a minor version upgrade!

What you can do in your downstream applications is `provide: ext-sodium` (or install a polyfill that does that for you), and take that decision **yourself**.

We will not take that decision for your users: it is not what a library should do anyway, and you can get a shot in the foot yourself via `"replace"` and `"provide"` in `composer.json` in your own distributed package, with the added responsibility.

Locking here: it's a waste of time, and thinking about a 5.x upgrade is what you should focus on anyway.

PR #1052, opened 2024-03-22
Require paragonie/sodium_compat instead of ext-sodium

### LINKED ISSUE #717: Composer should only suggest ext-sodium
Composer.json currently has:

`
"require": { 
  "ext-sodium": "*",
}
`

This seems to mean that the package will not install if sodium is missing from PHP.

However, a polyfill was added to provide a fallback if sodium was missing. Therefore should the requirement for ext-sodium not be moved to "suggests" ?  This would allow a user to install the package when sodium is not in PHP, and to use the polyfill ?

leofeyer:
Fixes #717

We still suggest installing `ext-sodium` because it will increase the performance (see https://github.com/paragonie/sodium_compat/blob/master/composer.json#L64). But thanks to the polyfill, `ext-sodium` is no longer a hard dependency. 🎉

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "require": {
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
         "ext-openssl": "*",
-        "ext-sodium": "*",
+        "paragonie/sodium_compat": "^1.20",
         "psr/clock": "^1.0"
     },
     "require-dev": {
@@ -35,7 +35,8 @@
         "phpunit/phpunit": "^10.2.6"
     },
     "suggest": {
-        "lcobucci/clock": ">= 3.0"
+        "lcobucci/clock": ">= 3.0",
+        "ext-sodium": "Better performance, password hashing (Argon2i), secure memory management (memzero), and better security."
     },
     "autoload": {
         "psr-4": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,144 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "5ba08423e19ae56456ff35b3290fc22d",
+    "content-hash": "8e37850d6b636bfbe54c41768e099c31",
     "packages": [
+        {
+            "name": "paragonie/random_compat",
+            "version": "v9.99.100",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/paragonie/random_compat.git",
+                "reference": "996434e5492cb4c3edcb9168db6fbb1359ef965a"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/paragonie/random_compat/zipball/996434e5492cb4c3edcb9168db6fbb1359ef965a",
+                "reference": "996434e5492cb4c3edcb9168db6fbb1359ef965a",
+                "shasum": ""
+            },
+            "require": {
+                "php": ">= 7"
+            },
+            "require-dev": {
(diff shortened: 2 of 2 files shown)


Slamdunk:
As written in https://github.com/lcobucci/jwt/pull/1051#issuecomment-2014597505 we prefer to let the user make this choice by adding the following config in its own `composer.json`:

```json
    "require": {
        "paragonie/sodium_compat": "^1.20"
    },
    "replace": {
        "ext-sodium": "*"
    },
```

Any PR against the documentation for this very topic will be very welcome :muscle:

Ocramius:
`paragonie/sodium_compat` is a polyfill: you install it on the edge, not in a library.

As a general rule, polyfills don't belong in libraries: libraries depend on specific tooling, and that tooling is then polyfilled where unavailable (declaring presence of the tooling, therefore "filling" it).

Therefore, in your downstream project (at the **application** level,  **not** in a library):

```json
{
  "require": {
    "lcobucci/jwt": "^5.2.0",
    "paragonie/sodium_compat": "^1.20.0"
  },
  "replace": {
     "ext-sodium": "*"
  }
}
```

```sh
❯ docker run --rm -ti composer bash
Unable to find image 'composer:latest' locally
latest: Pulling from library/composer
4abcf2066143: Already exists 
4b99432ace8a: Already exists 
d3a73aae1c86: Already exists 
9294d9185d89: Already exists 
da1b90e62964: Already exists 
dae3e00ca24c: Already exists 
1609ff101548: Already exists 
91cbcc61e4f3: Already exists 
d28ea7735d26: Already exists 
5446276905b0: Already exists 
1b585c525ada: Already exists 
b108eda9cf20: Already exists 
253265197c2a: Already exists 
815da127b652: Already exists 
Digest: sha256:63c0f08ca413700adcec721aa425e1247304c98314ed0bc2e5fc3699424e2364
Status: Downloaded newer image for composer:latest
d70f0770e0c3:/app# echo '{"require":{"lcobucci/jwt": "^5.2.0","paragonie/sodium_compat": "^1.20.0"},"replace":{"ext-sodium":"*"}}' > composer.json
d70f0770e0c3:/app# php -n $(which composer) update --dry-run
Loading composer repositories with package information
Updating dependencies
Lock file operations: 4 installs, 0 updates, 0 removals
  - Locking lcobucci/jwt (5.2.0)
  - Locking paragonie/random_compat (v9.99.100)
  - Locking paragonie/sodium_compat (v1.20.0)
  - Locking psr/clock (1.0.0)
Installing dependencies from lock file (including require-dev)
Package operations: 4 installs, 0 updates, 0 removals
  - Installing psr/clock (1.0.0)
  - Installing lcobucci/jwt (5.2.0)
  - Installing paragonie/random_compat (v9.99.100)
  - Installing paragonie/sodium_compat (v1.20.0)
3 package suggestions were added by new dependencies, use `composer suggest` to see details.
No installed packages - skipping audit.

```

EDIT: @Slamdunk you are darn fast :D

leofeyer:
> Any PR against the documentation for this very topic will be very welcome 

My suggestion works for everyone out of the box. No additional configuration or documentation needed. 🤷‍♂️

Ocramius:
@leofeyer libraries should not import polyfills in first place :scream: 

The library just declares what it wants: either your system dependency or a manually imported polyfill provides the requirement, but not the library.

Slamdunk:
> My suggestion works for everyone out of the box.

That's a wrong statement: PR's suggestion **seems** to work out of the box for everyone, buy polyfills are by no mean the exact copy of the original tool; take a look at `paragonie/sodium_compat` README file.

It's up to the developer to make a conscious decision on the tradeoffs, and certainly not up to this library's maintainers to take the corresponding liability.

PR #1053, opened 2024-04-02
Add constraint to verify the existence of a custom claim

freebuu:
Hello!

I want to add a HasClaim constraint that validates a presence of claim in token.

It useful for custom claims with non stable values (device_id etc.)

### DIFF
--- a/src/Validation/Constraint/HasClaim.php
+++ b/src/Validation/Constraint/HasClaim.php
@@ -8,11 +8,16 @@ use Lcobucci\JWT\UnencryptedToken;
 use Lcobucci\JWT\Validation\Constraint;
 use Lcobucci\JWT\Validation\ConstraintViolation;
 
+use function in_array;
+
 final class HasClaim implements Constraint
 {
     /** @param non-empty-string $claim */
     public function __construct(private readonly string $claim)
     {
+        if (in_array($claim, Token\RegisteredClaims::ALL, true)) {
+            throw CannotValidateARegisteredClaim::create($claim);
+        }
     }
 
     public function assert(Token $token): void
--- a/src/Validation/Constraint/HasClaim.php
+++ b/src/Validation/Constraint/HasClaim.php
@@ -1,0 +1,30 @@
+<?php
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Validation\Constraint;
+
+use Lcobucci\JWT\Token;
+use Lcobucci\JWT\UnencryptedToken;
+use Lcobucci\JWT\Validation\Constraint;
+use Lcobucci\JWT\Validation\ConstraintViolation;
+
+final class HasClaim implements Constraint
+{
+    /** @param non-empty-string $claim */
+    public function __construct(private readonly string $claim)
+    {
+    }
+
+    public function assert(Token $token): void
+    {
+        if (! $token instanceof UnencryptedToken) {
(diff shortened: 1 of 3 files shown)


tests/Validation/Constraint/HasClaimTest.php
```diff
@@ -0,0 +1,49 @@
...
+ * @uses \Lcobucci\JWT\Token\Plain
+ * @uses \Lcobucci\JWT\Token\Signature
+ */
+final class HasClaimTest extends ConstraintTestCase
+{
+    /** @test */
+    public function assertShouldRaiseExceptionWhenClaimIsNotSet(): void
+    {
+        $this->expectException(ConstraintViolation::class);
+        $this->expectExceptionMessage('The token does not have the claim "claimId"');
```
Ocramius, line 53: Let's put the expectations right before the failing call, please: don't do any other calls in-between, or we'd have hidden regressions. 

This is for all tests

src/Validation/Constraint/HasClaim.php
```diff
@@ -0,0 +1,30 @@
...
+    {
+    }
+
+    public function assert(Token $token): void
+    {
+        if (! $token instanceof UnencryptedToken) {
+            throw ConstraintViolation::error('You should pass a plain token', $this);
+        }
+
+        $claims = $token->claims();
```
Ocramius, line 29: is this API only present on a plain token?

Ocramius (dismissed):

src/Validation/Constraint/HasClaim.php
```diff
@@ -0,0 +1,30 @@
...
+use Lcobucci\JWT\Token;
+use Lcobucci\JWT\UnencryptedToken;
+use Lcobucci\JWT\Validation\Constraint;
+use Lcobucci\JWT\Validation\ConstraintViolation;
+
+final class HasClaim implements Constraint
+{
+    /** @param non-empty-string $claim */
+    public function __construct(private readonly string $claim)
+    {
```
Slamdunk, line 17: We don't want Registered Claims to be checked here, see

https://github.com/lcobucci/jwt/blob/d6d30cde835c9399219df8496b4993cb1abd2b77/src/Validation/Constraint/HasClaimWithValue.php#L18-L20

Slamdunk (dismissed):

src/Validation/Constraint/HasClaim.php
freebuu: Yes, `claims()` only present in `UnecryptedToken` interface

tests/Validation/Constraint/HasClaimTest.php
freebuu: I get this format from `HasClaimWithValueTest`. If i change it here to this format - i think it must be done in all tests
```
$constraint = new HasClaim('claimId');

$this->expectException(ConstraintViolation::class);
$this->expectExceptionMessage('The token does not have the claim "claimId"');

$constraint->assert($this->buildToken());
```

lcobucci: You're right that it must done in all tests, we'll eventually get there.

It would be highly appreciated, though, if you could help us by not making the backlog bigger.

freebuu: No problem sir, it's just about consistency between tests :)

lcobucci (approved):
Thanks for your contribution 👍

PR #1056, opened 2024-04-14
Use attributes for PHPUnit

lcobucci:


tests/ConfigurationTest.php
```diff
@@ -17,21 +17,16 @@
...
- *
- * @uses \Lcobucci\JWT\Encoding\ChainedFormatter
- * @uses \Lcobucci\JWT\Encoding\MicrosecondBasedDateConversion
- * @uses \Lcobucci\JWT\Encoding\UnifyAudience
- * @uses \Lcobucci\JWT\Signer\Key\InMemory
- * @uses \Lcobucci\JWT\Token\Builder
- * @uses \Lcobucci\JWT\Token\Parser
- * @uses \Lcobucci\JWT\Validation\Validator
- */
+#[PHPUnit\CoversClass(Configuration::class)]
```
SvenRtbg, line 24: I'd argue that summarizing all code execution inside this class is an incorrect generalization given that there are specific method names given that are to be covered.

lcobucci: Given the attributes [only targets the class level](https://github.com/sebastianbergmann/phpunit/blob/main/src/Framework/Attributes/CoversMethod.php#L19), having a list of all the covered methods at the test class level would bring no benefits here.

Sebastian explains the reasoning behind his decisions here: https://github.com/sebastianbergmann/phpunit/issues/4502

Although I like the granularity, I understand the reasons and am fine with the changes in PHPUnit 10 and 11 :smile:

PR #1076, opened 2024-11-04
Mark Builder methods as pure

### LINKED ISSUE #1075: Please mark Builder methods `@pure` to ease understanding that it's immutable
After upgrading to v5, I spent some time figuring out why the code broke, but this wasn't detected by PHPStan or any other tool.

After some tests, I found out that the reason is making the Builder immutable (https://github.com/lcobucci/jwt/pull/979) in the version.

Please consider marking the methods `@pure`, so SCA can report that 'the result is not used'.

Rule check example: https://phpstan.org/r/f638ba16-0ab3-44e2-becd-2927e99ada09

b1rdex:
Fix #1075

### DIFF
--- a/src/Builder.php
+++ b/src/Builder.php
@@ -18,11 +18,15 @@ interface Builder
      * Appends new items to audience
      *
      * @param non-empty-string ...$audiences
+     *
+     * @pure
      */
     public function permittedFor(string ...$audiences): Builder;
 
     /**
      * Configures the expiration time
+     *
+     * @pure
      */
     public function expiresAt(DateTimeImmutable $expiration): Builder;
 
@@ -30,11 +34,15 @@ interface Builder
      * Configures the token id
      *
      * @param non-empty-string $id
+     *
+     * @pure
      */
     public function identifiedBy(string $id): Builder;
 
     /**
      * Configures the time that the token was issued
+     *
+     * @pure
      */
     public function issuedAt(DateTimeImmutable $issuedAt): Builder;
 
@@ -42,11 +50,15 @@ interface Builder
      * Configures the issuer
      *
      * @param non-empty-string $issuer
+     *
+     * @pure
      */
     public function issuedBy(string $issuer): Builder;
 
     /**
      * Configures the time before which the token cannot be accepted
+     *
+     * @pure
      */
     public function canOnlyBeUsedAfter(DateTimeImmutable $notBefore): Builder;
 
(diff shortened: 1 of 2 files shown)


Slamdunk (changes requested):
We can't change the interface as it would be a https://github.com/lcobucci/jwt/labels/BC-break

Please only mark as `@pure` the implementations.

Moreover I think we can mark so the whole class, IIRC

Ocramius:
Possibly worth making a major release for such an interface addition - any PRNG stuff can't be pure though, obviously

Slamdunk:
@Ocramius where is the PRNG here? As far as I know, only `PS256` algorithm involves PRNG, the other ones are all deterministic :thinking:

b1rdex:
> Moreover I think we can mark so the whole class, IIRC

This is not possible, because `@pure` is only for functions/methods.



> We can't change the interface as it would be a [ BC-break](https://github.com/lcobucci/jwt/labels/BC-break)

I'm not sure why could this be considered as a BC break, because this only makes obvious for SCA tools of what was intended to be introduced in the original PR (https://github.com/lcobucci/jwt/pull/979). The behavior is not changed.

> Possibly worth making a major release for such an interface addition - any PRNG stuff can't be pure though, obviously

There is nothing related to PRNG, the methods are just plain "setters".

Ocramius:
> @Ocramius where is the PRNG here? As far as I know, only `PS256` algorithm involves PRNG, the other ones are all deterministic 🤔

`getToken()` interacts with the `Signer`, which uses CSPRNG under the hood

Slamdunk:
> `getToken()` interacts with the `Signer`

I know, and I'll reiterate: there are no random number generation in any currently supported Signer :thinking:

Ocramius:
All signatures are deterministic? They don't even factor in any random offset/salt to make brute forcing harder? 😱

Slamdunk:
> All signatures are deterministic? They don't even factor in any random offset/salt to make brute forcing harder? 😱

All the doc I read state that this is by design: getting secure randomness is tricky, so it's better to only do it while generating the key.

BTW, back to the topic: @b1rdex please update the `Token\Builder` functions signature, excluding the `getToken` just to be sure, and we can :ship: it :muscle:

Slamdunk (approved):

lcobucci:
Given this is marked as a bug fix, I've changed the target branch to 5.4.x.

@b1rdex would please rebase the branch and remove the commits that aren't related to your changes?

Once we do that, we're good for merge & release.

Thanks everyone!

b1rdex:
@lcobucci done

PR #1082, opened 2024-11-07
Add `#[SensitiveParameter]` attribute to sensitive parameters

### LINKED ISSUE #1081: Mark sensitive parameters with `#[SensitiveParameter]`
We noticed yesterday in some of our logs that stack traces coming out of the library included some private key materials that were shipped off in our logs. 

Given that PHP support for PHP 8.1 was dropped earlier this year, adding `#[SensitiveParameter]` to the sensitive parameters will prevent this from happening.

slknijnenburg:
Fixes #1081, adds `#[SensitiveParameter]` to all potentially sensitive parameters, including key material, certificates and passphrases.

I wondered if this would be a BC break, but it seems that that is not the case. Child classes/interface implementations can apparently omit the attribute without any problems.

### DIFF
--- a/src/Signer/Key/InMemory.php
+++ b/src/Signer/Key/InMemory.php
@@ -6,6 +6,7 @@ namespace Lcobucci\JWT\Signer\Key;
 use Lcobucci\JWT\Signer\InvalidKeyProvided;
 use Lcobucci\JWT\Signer\Key;
 use Lcobucci\JWT\SodiumBase64Polyfill;
+use SensitiveParameter;
 use SplFileObject;
 use Throwable;
 
@@ -15,21 +16,33 @@ use function is_string;
 final class InMemory implements Key
 {
     /** @param non-empty-string $contents */
-    private function __construct(public readonly string $contents, public readonly string $passphrase)
-    {
+    private function __construct(
+        #[SensitiveParameter]
+        public readonly string $contents,
+        #[SensitiveParameter]
+        public readonly string $passphrase,
+    ) {
     }
 
     /** @param non-empty-string $contents */
-    public static function plainText(string $contents, string $passphrase = ''): self
-    {
+    public static function plainText(
+        #[SensitiveParameter]
+        string $contents,
+        #[SensitiveParameter]
+        string $passphrase = '',
+    ): self {
         self::guardAgainstEmptyKey($contents);
 
         return new self($contents, $passphrase);
     }
 
     /** @param non-empty-string $contents */
-    public static function base64Encoded(string $contents, string $passphrase = ''): self
-    {
+    public static function base64Encoded(
+        #[SensitiveParameter]
+        string $contents,
+        #[SensitiveParameter]
+        string $passphrase = '',
+    ): self {
         $decoded = SodiumBase64Polyfill::base642bin(
             $contents,
             SodiumBase64Polyfill::SODIUM_BASE64_VARIANT_ORIGINAL,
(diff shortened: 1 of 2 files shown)


Slamdunk (changes requested):
Thank you for this contribution.

As far as I know, `#[SensitiveParameter]` only makes sense in `string` parameters, please update the PR to only target:

1. `\Lcobucci\JWT\Signer\Key\InMemory` methods
2. `\Lcobucci\JWT\Signer\OpenSSL::createSignature`
3. `\Lcobucci\JWT\Signer\OpenSSL::verifySignature`

Also, as this can be considered a https://github.com/lcobucci/jwt/labels/Bug please target the `5.4.x` branch.

src/Signer/Key/InMemory.php
```diff
@@ -15,21 +16,30 @@
 final class InMemory implements Key
 {
     /** @param non-empty-string $contents */
-    private function __construct(public readonly string $contents, public readonly string $passphrase)
-    {
+    private function __construct(
+        public readonly string $contents,
```
Slamdunk, line 21: `$contents` should be protected as well

src/Signer/OpenSSL.php
```diff
@@ -56,15 +59,20 @@ final protected function createSignature(
...
+        string $passphrase,
+    ): OpenSSLAsymmetricKey {
         return $this->validateKey(openssl_pkey_get_private($pem, $passphrase));
     }
 
     /** @throws InvalidKeyProvided */
     final protected function verifySignature(
         string $expected,
         string $payload,
+        #[SensitiveParameter]
```
Slamdunk: I apologize for my previous statement: verification operations only deal with public keys, and by definition public keys don't neet to be protected.

We can revert changes to both `verifySignature` and `getPublicKey`

Slamdunk (changes requested):

src/Signer/Key/InMemory.php
```diff
@@ -15,21 +16,31 @@
...
+        #[SensitiveParameter]
+        public readonly string $passphrase,
+    ) {
     }
 
     /** @param non-empty-string $contents */
-    public static function plainText(string $contents, string $passphrase = ''): self
-    {
+    public static function plainText(
+        string $contents,
```
Slamdunk, line 30: `$contents` should be protected also here

src/Signer/Key/InMemory.php
```diff
@@ -15,21 +16,31 @@
...
         self::guardAgainstEmptyKey($contents);
 
         return new self($contents, $passphrase);
     }
 
     /** @param non-empty-string $contents */
-    public static function base64Encoded(string $contents, string $passphrase = ''): self
-    {
+    public static function base64Encoded(
+        string $contents,
```
Slamdunk, line 42: And here

src/Signer/OpenSSL.php
```diff
@@ -74,8 +81,9 @@ final protected function verifySignature(
     }
 
     /** @throws InvalidKeyProvided */
-    private function getPublicKey(string $pem): OpenSSLAsymmetricKey
-    {
+    private function getPublicKey(
+        string $pem,
+    ): OpenSSLAsymmetricKey {
```
Slamdunk: Revert this change please.

Slamdunk (changes requested):
Awesome, few final fixes and we can :ship: it :muscle:

Slamdunk (approved):

Ocramius (approved):

TimWolla:
> As far as I know, `#[SensitiveParameter]` only makes sense in `string` parameters

This is false. The attribute is agnostic to the parameter type and the Exception handler of your choice might opt to also process objects (e.g. by printing their `__debugInfo()`) output. The attribute should be apply to anything that is sensitive, independent of type.

Slamdunk:
Mmm, by this logic everything is sensitive, if my debug handler dumps the entire memory and filesystem to the logger.

I think handling the default behaviour is enough for us.

PR #1090, opened 2024-11-22
Implement PS256/384/512

### LINKED ISSUE #1074: PS256 algorithm support?
Hello.

PS256 is listed as unsupported on jwt.io for lcobucci. I need this algo for authenticating with one cloud operator. Any plans for it to be supported?

SvenRtbg:
The heavy work is offloaded to phpseclib/phpseclib v3, which is added as a dependency.

Note: Please focus review onto the fact that everyone states that RSASSA-PSS key pairs are somehow special.

My understanding is that PSS is just a different kind of padding that utilizes randomness as a salt, and the signing part is just basic RSA. I have tested with dedicated RSA-PSS key pairs, and the only difference is that the key is explicitly labeled as RSA-PSS, and may contain additional info about the expected hash, mgfhash and salt length. It wasn't noticed by the PHPSecLib implementation, though.

Keep in mind I might have missed an important point here, as I implemented the obvious part, and maybe some non-obvious things, but I wouldn't consider myself the expert here.

Most importantly, I would like to see someone testing against a real-world token use case, as the tests inside are basically verifying that the implementation in the class matches the implementation in the test, which is effectively the same code.

closes #1074

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
         "php": "~8.2.0 || ~8.3.0 || ~8.4.0",
         "ext-openssl": "*",
         "ext-sodium": "*",
+        "phpseclib/phpseclib": "^3.0.36",
         "psr/clock": "^1.0"
     },
     "require-dev": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,235 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "f6da93713398752c6e06805659ca09e4",
+    "content-hash": "d0c70f2f7595443761936bc2ce91d5df",
     "packages": [
+        {
+            "name": "paragonie/constant_time_encoding",
+            "version": "v3.0.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/paragonie/constant_time_encoding.git",
+                "reference": "df1e7fde177501eee2037dd159cf04f5f301a512"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/paragonie/constant_time_encoding/zipball/df1e7fde177501eee2037dd159cf04f5f301a512",
+                "reference": "df1e7fde177501eee2037dd159cf04f5f301a512",
+                "shasum": ""
+            },
+            "require": {
+                "php": "^8"
+            },
+            "require-dev": {
+                "phpunit/phpunit": "^9",
+                "vimeo/psalm": "^4|^5"
+            },
+            "type": "library",
+            "autoload": {
+                "psr-4": {
+                    "ParagonIE\\ConstantTime\\": "src/"
+                }
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "MIT"
+            ],
+            "authors": [
+                {
+                    "name": "Paragon Initiative Enterprises",
+                    "email": "security@paragonie.com",
+                    "homepage": "https://paragonie.com",
+                    "role": "Maintainer"
+                },
+                {
+                    "name": "Steve 'Sc00bz' Thomas",
+                    "email": "steve@tobtu.com",
+                    "homepage": "https://www.tobtu.com",
+                    "role": "Original Developer"
+                }
+            ],
+            "description": "Constant-time Implementations of RFC 4648 Encoding (Base-64, Base-32, Base-16)",
+            "keywords": [
+                "base16",
+                "base32",
+                "base32_decode",
+                "base32_encode",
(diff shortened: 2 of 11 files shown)


composer.json
```diff
@@ -20,6 +20,7 @@
         "php": "~8.2.0 || ~8.3.0 || ~8.4.0",
         "ext-openssl": "*",
         "ext-sodium": "*",
+        "phpseclib/phpseclib": "^3.0.36",
```
Slamdunk, line 23: Here I disagree with @lcobucci https://github.com/lcobucci/jwt/issues/1074#issuecomment-2486585502
While `phpseclib/phpseclib` is indeed good, I don't like adding that entire dependency for such and uncommon and superseded signature algorithm.

lcobucci: We can disagree, and I totally see your point (specially for this use case). 

I don't like adding dependencies but, sometimes, they can help us going further with reduced effort.

Can we quantify things to have a more complete analysis? What are the pros/cons? How much is the lib footprint growing with each approach? How is our maintenability affected?

Slamdunk: My concern is that as a user I have no pratical way to tell if a flaw or security issue on the dependency directly affects (or not) the part of the main library I use.

(Native extensions have been less affected by this problem so far.)

Forcing every user to install such a huge dependency for something they will likely not use is annoying:

```json
    "replace": {
        "symfony/polyfill-ctype": "*",
        "symfony/polyfill-iconv": "*",
        "symfony/polyfill-intl-grapheme": "*",
        "symfony/polyfill-intl-idn": "*",
        "symfony/polyfill-intl-normalizer": "*",
        "symfony/polyfill-mbstring": "*",
        "symfony/polyfill-php70": "*",
        "symfony/polyfill-php72": "*",
        "symfony/polyfill-php73": "*",
        "symfony/polyfill-php80": "*",
        "symfony/polyfill-php81": "*",
        "symfony/polyfill-php82": "*",
        "symfony/polyfill-php83": "*"
    },
```

I have no scientific threshold of (dep weight) to (benefit gained) ratio to respect.

But I can tell you this is the exact case I'd publish a separate `jwt-rssassa-pss` package for.

SvenRtbg: @Slamdunk I wonder what that list of replacements is supposed to say. It does not seem to be anything that is added by PhpSecLib.

Anyways, so far I had some fun doing hands-on work, and I admit I'm not the person promoting this schema, and I'm not pushing it getting merged - so whatever decision is made is fine by me.

Keep in mind though, that some keywords for longstanding feature requests do appear in PhpSecLib, like JWK. And in fact the documentation explicitly uses JWT as an example for signing the tokens, although in quite a low-level approach. It could power all schemas that are to offer here, maybe without Blake2B.

If you decide to move on, I'll start fighting the mutants next week (which didn't appear locally - I was running `make`, no clue what happens behind the scene for now, but will look into it).

Ocramius (changes requested):
Before proceeding with this, I am noticing one interesting (and very good!) detail about this patch: it is all added files only.

I'm wondering if such a patch could sit in `lcobucci/jwt-rsassa-pss` instead.

Such an approach would fully externalize the dependency, and could be re-absorbed in the main `lcobucci/jwt` library after:

1. people started using it
2. upstream PHP has bindings needed for this to work without an added dependency (in SSL or `ext-sodium`)

I fully agree with @Slamdunk that adding a dependency here is a problem.

lcobucci:
Let's pull it to a separate lib, then, and see what the stats show us.

SvenRtbg:
I don't feel I'd be initially involved there, am I? I don't like leaving half-finished work, however I for sure lack admin permissions to create anything within @lcobucci 's projects. :) How to proceed?

Ocramius:
@SvenRtbg given you are alway very active here, I would say that we could totally (pending @lcobucci's opinion, obviously) give you maintainership on the new repo.

lcobucci:
I'll setup the new repo and give access to you folks. I'm just having limited time atm

SvenRtbg:
No need to rush, I'm AFK until Monday.

lcobucci:
https://github.com/lcobucci/jwt-rsassa-pss was created and permissions were granted.

I configured the very basic stuff (branch names, protection rules, keys for the auto-release stuff) and can help further another day!

Slamdunk:
Moved to https://github.com/lcobucci/jwt-rsassa-pss/pull/1

SvenRtbg:
> https://github.com/lcobucci/jwt-rsassa-pss was created and permissions were granted.
> 
> I configured the very basic stuff (branch names, protection rules, keys for the auto-release stuff) and can help further another day!

I'm eager to see what needs to be done next, having prepared an independent variant already - but it's just fragments still, no coherent project.

PR #1091, opened 2024-12-04
Mark all classes as `readonly`

### LINKED ISSUE #1086: Make `Token\Builder` really `readonly`
As reported by PHPStan 2, the current `Token\Builder` implementation is not really `readonly`:

https://github.com/lcobucci/jwt/actions/runs/11784435253/job/32823593891?pr=1085#step:9:50

All attributes must be accepted only in its constructor.
But we can't do that on a public constructor because we need the safeguards provided by its API.

The solution is to add a new `static function new()` method as the public constructor, make the constructor private, and only then apply the `readonly` attribute, hence the https://github.com/lcobucci/jwt/labels/BC-break label here

Slamdunk:
Fix https://github.com/lcobucci/jwt/issues/1086

This PR needs a new `6.0.x` branch

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -15,7 +15,7 @@ use Lcobucci\JWT\Validation\Constraint;
  * Serves like a small DI container to simplify the creation and usage
  * of the objects.
  */
-final class Configuration
+final readonly class Configuration
 {
     private Parser $parser;
     private Validator $validator;
@@ -28,11 +28,11 @@ final class Configuration
 
     /** @param Closure(ClaimsFormatter $claimFormatter): Builder|null $builderFactory */
     private function __construct(
-        private readonly Signer $signer,
-        private readonly Key $signingKey,
-        private readonly Key $verificationKey,
-        private readonly Encoder $encoder,
-        private readonly Decoder $decoder,
+        private Signer $signer,
+        private Key $signingKey,
+        private Key $verificationKey,
+        private Encoder $encoder,
+        private Decoder $decoder,
         ?Parser $parser,
         ?Validator $validator,
         ?Closure $builderFactory,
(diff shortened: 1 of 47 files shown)


src/Configuration.php
```diff
@@ -69,13 +80,26 @@ public static function forSymmetricSigner(
...
             $decoder,
+            null,
+            null,
+            null,
         );
     }
 
     /** @param callable(ClaimsFormatter): Builder $builderFactory */
-    public function setBuilderFactory(callable $builderFactory): void
+    public function setBuilderFactory(callable $builderFactory): self
```
lcobucci: Should we rename those setters to follow the whole `withWhateva()` convention?

src/Configuration.php
```diff
@@ -15,7 +15,7 @@
  * Serves like a small DI container to simplify the creation and usage
  * of the objects.
  */
-final class Configuration
+final readonly class Configuration
```
lcobucci, line 18: We need to open a discussion about this component... its design is a bit too convoluted for me.
I'll post something soon...

src/Token/Builder.php
```diff
@@ -17,16 +17,23 @@
...
+        private array $headers = ['typ' => 'JWT', 'alg' => null],
+        private array $claims = [],
+    ) {
+    }
 
-    public function __construct(private readonly Encoder $encoder, private readonly ClaimsFormatter $claimFormatter)
+    public static function new(Encoder $encoder, ClaimsFormatter $claimFormatter): self
     {
+        return new self($encoder, $claimFormatter);
     }
```
lcobucci: This is something that we'll need to coordinate in v5.5 the migration path for...

lcobucci (approved):
The changes are good, thanks.

I personally don't like the factory method `Builder::new()` but it's a necessary evil to ensure that we don't have people injecting crazy stuff there.

src/Configuration.php
Slamdunk: Done

src/Configuration.php
Slamdunk: I'm curious

Slamdunk:
> I personally don't like the factory method `Builder::new()`

This concern is easy to address with new `Headers`, `Header`, `Claims`, `RegisteredClaims`, `UnregisteredClaim`, `RegisteredClaim` objects and so on, but it would probabily be a bit overkill

src/Token/Builder.php
Slamdunk: Migration API added in https://github.com/lcobucci/jwt/pull/1092

PR #1094, opened 2024-12-10
Local build: rely only on `docker`

Slamdunk:


Makefile
```diff
@@ -1,40 +1,45 @@
...
+	DOCKER_PHP_EXEC := docker compose run --rm php
+endif
 PARALLELISM := $(shell nproc)
 
 .PHONY: all
-all: install phpcbf phpcs phpstan phpunit infection phpbench
+all: phpcbf phpcs phpstan phpunit infection phpbench
 
-.PHONY: install
-install: vendor/composer/installed.json
```
Slamdunk, line 7: There's no need to remember to run `make install`, just add `vendor/composer/installed.json` as a prerequisit for every other target

lcobucci:
Although I see the value in having a standard setup for development environments, I find the benefit very limited for this particular package.

We don't have an exotic set of dependencies nor a big complex set of containers to orchestrate to achieve something.

We also still need to run against multiple PHP versions, which would make things a bit more complex (sure, we can have a build argument and make Dockerfile variable).

So... balancing the complexity of things I honestly lean towards simplicity - unless we're having issues with different PHP configurations between different people.

@Ocramius what are your thoughts?

Slamdunk:
> I find the benefit very limited for this particular package.

The benefit is not for the package, is for the maintainers: I am currently close to 30 packages to maintain, I don't want to adapt the installed packages and aliases on my PC anymore

> We also still need to run against multiple PHP versions, which would make things a bit more complex

How is the `php` alias currently at the beginning of every line any easier than changing the version in the Dockerfile? It has always been a burden to me so far.

For the records, I'm using this PR on this very package for a year. It's a full year right now I have no packages installed on my PC and oh god it's so much easier to do everything.

Ocramius:
FWIW, I'm **very** close to using nix flakes for my stuff myself: I need it pinned like butterflies under a needle.

PR #1095, opened 2024-12-10
CI: simplify workflows with `ramsey/composer-install`

Slamdunk:


.github/workflows/phpunit.yml
```diff
@@ -16,7 +16,6 @@ jobs:
           - "lowest"
           - "highest"
           - "locked"
-          - "development"
```
Slamdunk, line 19: Dropped on purpose: we already build against unstable `php`, we don't need to follow every dependencies' dev step

Slamdunk: At least not anymore thanks to `renovatebot`

PR #1098, opened 2025-01-01
Update LICENSE, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛


Ocramius:
Happy to keep the original date, I'd say. Potentially even remove it?

SvenRtbg:
Copyrightlaw.com states the notice should bear the year of first publishing (and definitely targets printed works), but also adds that the © symbol isn't required by law, but informally adds the fact an infringer willfully is ignoring which may help in court.

In other words: The year most likely isn't to be updated each year.

PR #1102, opened 2025-02-06
Add PS256 etc algorithm support into docs.

SvenRtbg:
I intentionally target the 5.4 branch as the extra package supports this version as a minimum, even though I probably cannot change the docs for the 5.4.2 version that is tagged.

However I don't mind if this should simply go to the main branch instead.


Ocramius:
We can still tag `5.4.3`, if needed

SvenRtbg:
For reference, I also proposed https://github.com/jsonwebtoken/jsonwebtoken.github.io/pull/733 to get jwt.io updated.

lcobucci:
I'll fix the BC-break check (now that we released my patch 🎉)

PR #1105, opened 2025-02-18
Require PHP 8.3

Slamdunk:


.github/workflows/backwards-compatibility.yml
```diff
@@ -32,17 +32,11 @@ jobs:
         uses: "actions/cache@v4.2.0"
         with:
           path: ${{ steps.composer-cache.outputs.composer_cache_dir }}
-          key: "php-8.2-bc-break-check-${{ hashFiles('.github/workflows/backwards-compatibility.yml') }}"
-          restore-keys: "php-8.2-bc-break-check-"
-
-      # Only necessary until https://github.com/Roave/BackwardCompatibilityCheck/pull/737 gets resolved
```
Slamdunk, line 38: PR got resolved and merged, this can now be dropped

composer.json
```diff
@@ -54,9 +54,6 @@
             "ocramius/package-versions": true,
             "phpstan/extension-installer": true
         },
-        "platform": {
-            "php": "8.2.99"
```
Slamdunk, line 58: This configuration hinders us from proper testing against different versions in CI

Ocramius: This is here so Renovate won't upgrade out of range stuff

Slamdunk: At the same time this blocks us to test against newer deps that require higher PHP versions.

We already test against `lowest` and `highest`, it seems safer to me to *not* have this lines

Ocramius: Yes, just saying that you will get invalid lock file updates

Slamdunk: I know: ditching `composer.lock` from committed files is my next PR indeed

PR #1113, opened 2025-06-09
Fix failing build

Slamdunk:
I have no idea how https://github.com/lcobucci/jwt/pull/1111 got auto-merged with a failing build :thinking:


lcobucci:
That's indeed super weird and unexpected

PR #1138, opened 2025-11-10
Allow 1024 bit keys to be the minimum key length

dimegalev:
This change only servers as a way to allow us to upgrade to a higher version of Symfony while we migrate our keys to 2048 bit keys and let our users aquire access/refresh tokens based on the 2048 after their current tokens expire. After a while we will switch back to the official lcobucci/jwt package.

### DIFF
--- a/src/Signer/Rsa.php
+++ b/src/Signer/Rsa.php
@@ -7,7 +7,7 @@ use const OPENSSL_KEYTYPE_RSA;
 
 abstract class Rsa extends OpenSSL
 {
-    private const MINIMUM_KEY_LENGTH = 2048;
+    private const MINIMUM_KEY_LENGTH = 1024;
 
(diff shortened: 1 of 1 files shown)


SvenRtbg:
But how does this work as an official pull request to the official package? I guess you misdirected your PR. No offense, but even if you didn't, this isn't going to be merged.

SvenRtbg:
Migration hint: Using elliptic curve cryptography is way better than still relying on RSA, so you may want to look into ES256 algorithms for JWTs.

dimegalev:
> But how does this work as an official pull request to the official package? I guess you misdirected your PR. No offense, but even if you didn't, this isn't going to be merged.

I apologize, this is indeed a misdirected PR, it was supposed to be on a temporary private fork.

PR #1140, opened 2025-11-27
add magic method, elementary and easy

chokoladis:


Ocramius (changes requested):
No, `__toString()` has way worse implications from a type perspective.

See https://github.com/ShittySoft/symfony-live-berlin-2018-doctrine-tutorial/pull/3#issuecomment-460085493 for a detailed explanation of why an explicit method call is preferable, over a string cast.

PR #1141, opened 2025-11-28
Add `#[\NoDiscard]` attribute

Slamdunk:
In theory we can add the attribute everywhere, but here I tried adding it only where there's an added value.
So the implementation involves:

1. APIs that create something new
2. APIs that compute a result that's not part of the object's internal attributes: `validate`, `format`, `toString` etc

It left aside:

1. Getters
2. APIs with `void` return type (of course)


tests/Benchmark/CreateSignatureBench.php
```diff
@@ -22,6 +22,6 @@ public function initialize(array $params): void
 
     protected function runBenchmark(): void
     {
-        $this->algorithm->sign(self::PAYLOAD, $this->key);
+        $void = $this->algorithm->sign(self::PAYLOAD, $this->key);
```
Slamdunk, line 25: Unfortunately the `(void)` cast doesn't compile on PHP < 8.5

tests/ES512TokenTest.php
```diff
@@ -66,31 +66,31 @@ public function createConfiguration(): void
...
 
         $this->expectException(InvalidKeyProvided::class);
         $this->expectExceptionMessage('It was not possible to parse your key, reason:');
 
-        $builder->identifiedBy('1')
-            ->permittedFor('https://client.abc.com')
-            ->issuedBy('https://api.abc.com')
-            ->withClaim('user', ['name' => 'testing', 'email' => 'testing@abc.com'])
-            ->getToken($this->config->signer(), InMemory::plainText('testing'));
+        $void = $builder->getToken($this->config->signer(), InMemory::plainText('testing'));
```
Slamdunk, line 78: Minor improvement here: all the logics that don't raise the exception yet should go before the `expectException`

Ocramius:
The fact that `#[\NoDiscard]` had to be put in the baseline of `maglnet/composer-require-checker` raises a valid concern: what happens if this codebase is analyzed in any way, at reflection level?

Is downstream supposed to polyfill `#[\NoDiscard]` on their own? Is a crash acceptable at all, in such a scenario?

src/Builder.php
```diff
@@ -81,5 +91,6 @@ public function withClaim(string $name, mixed $value): Builder;
      * @throws InvalidKeyProvided  When issue key is invalid/incompatible.
      * @throws ConversionFailed    When signature could not be converted.
      */
+    #[NoDiscard]
```
Ocramius, line 94: For methods that have `@throws`, I'm not 100% sure we should have `#[NoDiscard]` every time: downstream may be using these only for assertions, which seems valid to me

Slamdunk: Such developer should make that intention clear by `(void)` casting what's returned

Slamdunk:
> The fact that `#[\NoDiscard]` had to be put in the baseline of `maglnet/composer-require-checker` raises a valid concern: what happens if this codebase is analyzed in any way, at reflection level?
> 
> Is downstream supposed to polyfill `#[\NoDiscard]` on their own? Is a crash acceptable at all, in such a scenario?

I asked myself the same question, then I remembered that Attributes have been made safe by default by being comment, that means that a crash only happens when calling `ReflectionAttribute::newInstance`:

https://3v4l.org/D5g7q

Until then, the code is PHP < 8.5 compatible

tests/Benchmark/CreateSignatureBench.php
Ocramius: One last question here: why does our static analysis not pick this up?

Ocramius (approved):

tests/Benchmark/CreateSignatureBench.php
Slamdunk: I'm taking a guess here: because `Signer` is not marked `@immutable`.
Shall we make it so?

Ocramius: I don't think it can be `@immutable`, since it uses environment data (prng) to add signature details?

Slamdunk: Deterministic output is something for `@pure`, not `@immutable`, isn't it?

Ocramius: `@immutable` has always implicitly contained `@pure`, in SA tools.

Slamdunk: I am sorry but what you say differs from what I read online:

https://psalm.dev/articles/immutability-and-beyond

> `@immutable` guarantees that all class properties can be written to just once in the constructor, and all instance methods won’t mutate state outside the class

Depending on a mutable state (like randomness) doesn't mean that the class itself mutates such state

PHPStan has an even more loose definition https://phpstan.org/writing-php-code/phpdocs-basics#immutable-classes

> `@immutable` or `@readonly` on the class can be used to make PHPStan treat every property of that class as being readonly.

Ocramius: Hmm, I must be running on outdated knowledge: means I have to add `@psalm-pure` to my symbols, according to that article.

This specific example used to error, some time ago: https://psalm.dev/r/9ff43ced74 (I'm fairly sure of it, given that I worked on immutability/purity upstream patches for it).

I'd say that enforcing `@immutable` on the `Signer` interface is a good idea then :+1:

lcobucci (approved):

PR #1155, opened 2026-04-13
Require PHP 8.4

lcobucci:
PHP 8.3 entered the security fixes-only phase and, per our policy, should no longer be supported in new versions of the library.


Ocramius:
@lcobucci does this really need to target the next major only though?

lcobucci:
> @lcobucci does this really need to target the next major only though?

I was planning to simply release it this week... 😅

Issue #1161, opened 2026-06-14
Post-Quantum signing/cryptography support 


primordialomegazero:
Hi,

I would to know if you want to convert your jwt to Post-quantum hybrid jwt.

Please let me know the exact details or requirements before I send a Pull request.

Thanks, 

Dan (Primordial Omega Zero)

Ocramius:
JWT is a spec: post-quantum would only be one of the supported signing/encryption algorithms, unless that requires implementing a different spec?

primordialomegazero:
Yes, just additional signing algorithms — no spec changes needed. 
I can send a PR if you're interested.

Slamdunk:
We'd like links to related RFCs first, please

primordialomegazero:
@Slamdunk Here are the relevant NIST PQC standards:

- FIPS 203: ML-KEM (Module-Lattice Key Encapsulation) 
  https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf

- FIPS 204: ML-DSA (Module-Lattice Digital Signature) 
  https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf

- FIPS 205: SLH-DSA (Stateless Hash-Based Signature)
  https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.205.pdf

- RFC 7519: JSON Web Token (JWT) — unchanged
  https://datatracker.ietf.org/doc/html/rfc7519

- NIST PQC Project:
  https://csrc.nist.gov/projects/post-quantum-cryptography

These are the official NIST standards for PQC. 
Falcon-512 and Dilithium-2 are FIPS 204 compliant.

Ocramius:
Closing here too: it seems like #1162 is an attempt at patching this in "as a placeholder", but libraries or extensions for providing cryptographic tooling to implement this aren't yet available.

Please raise a new issue once a reliable and trustworthy PHP abstraction for the necessary cryptographic tooling is available.

PR #1162, opened 2026-06-14
Add Falcon-512 Post-Quantum Signer (FIPS 204)

primordialomegazero:
Add Falcon-512 post-quantum signing algorithm.

- Pure PHP implementation
- Compatible with existing Signer interface
- Uses OpenSSL for signing/verification
- NIST FIPS 204 compliant
- No external dependencies

RFC: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf

### DIFF
--- a/src/Signer/Falcon512.php
+++ b/src/Signer/Falcon512.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Lcobucci\JWT\Signer;
+
+use Lcobucci\JWT\Signer\Key\InMemory;
+use SensitiveParameter;
+use function openssl_sign;
+use function openssl_verify;
+use function hash_hmac;
+use function hash_equals;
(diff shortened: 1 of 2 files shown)


lcobucci (changes requested):
This has no tests and doesn't use the post-quantum algorithm....

primordialomegazero:
@lcobucci Added test suite with 3 tests:
- Sign and verify
- Tamper detection
- Algorithm ID check

Regarding PQC: PHP lacks native liboqs binding. 
Currently uses OpenSSL as signing backend (placeholder for Falcon-512).
For production PQC, we can add shell_exec to liboqs CLI.

Tests passing locally. Ready for re-review.

Ocramius:
Closing here: this is obviously an implementation using SHA512, rather than what was outlined in #1161

PR #1181, opened 2026-07-30
Add ES256K (ECDSA secp256k1) signer support

### LINKED ISSUE #1180: Implement ES256K
Summary:

ES256K uses a specific elliptic curve "secp256k1" and is mostly used in cryptocurrency environments, but is defined for JWTs as well. I came across it when dealing with https://github.com/jsonwebtoken/jsonwebtoken.github.io/pull/938 (updating https://www.jwt.io/libraries).

Details:

OpenSSL is doing all heavy lifting, as dealing with a specific curve is simply handing in an appropriate key when calling `openssl_sign()`, i.e. the elliptic curve key must be "secp256k1".

At the same time, this implies that regular ES256/384/512 keys need to use a different curve, which indeed is made explicit in RFC7518 (see https://www.rfc-editor.org/info/rfc7518/#section-3.4 where ES256 has to use curve "P-256", ES384 uses "P-384" and ES512 uses "P-521" (sic!)). Currently, this library only checks the key length in bits, but does not care about the specified curve.

I'll send a PR right away to allow discussion implementation details.

SvenRtbg:
Resolves #1180 

Adds support for the ES256K algorithm (ECDSA using the secp256k1 curve and SHA-256), as defined by RFC 8812:
https://www.rfc-editor.org/rfc/rfc8812

**Required discussion topic: Curve verification**

The RFC 8812 requires verifying that the key actually uses secp256k1, as any other curve wouldn't be correct. This is implemented by adding `Lcobucci\JWT\Signer\Ecdsa::expectedCurve()` returning null (no expectation) or a string with the curve name that openssl will also return from `openssl_pkey_get_details()` in `$details['ec']['curve_name']`, and the verification code in `Lcobucci\JWT\Signer\OpenSSL`.

The current implementation does trigger infection zombies, as there is no test in the current code base to ensure the `expectedCurve()` method has to be public - infection turns it into `protected` without complaints so far.

On the bigger picture, this triggers my question: Should the other ES signatures maybe implement verifying their curve, too? The RFC 7518 states that there is a defined combination between the signature method annouced in the token, and the curve and hash mechanism, i.e. ES256 MUST use "P-256" (also called "prime256v1") curve and SHA256, for ES384 it is "P-384" or "secp384r1" with SHA384, and ES512 MUST use "P-521" or "secp521r1". The absence of any of the "MUST", "SHOULD" or "MAY" keywords in that section of the RFC might indicate no strict requirement for verification, but there is basically no barrier in the library code to prevent using the wrong curve. In a closed loop, verifying ones own tokens with the wrong curve would work, but they are still incorrect.

**Disclosure**

The code is AI generated by my prompts as an introductory task to gain more knowledge. However it passes my expectations of what would have to be coded, and pretty much looks like something I would have done manually. Still I expect some finishing touches to be applied after your review feedback.


Ocramius:
From a pure code PoV, I didn't see much worth noting (infection already highlights what's worth hardening).

I have no familiarity with the algorithm though: who can we ask for another check?

SvenRtbg:
To me, the algorithm pretty much sounds like "more of the same", with the distinct exception of using a different curve compared to ES256, i.e. all security expectations should align.

Ocramius (approved):

SvenRtbg:
Yesterday, Github was struggling to finish all existing check jobs, so I don't trust the fact that two of two were successful. I'm trying to retrigger them all again, I don't believe all infection issues are resolved (one remaining wasn't addressed until now).

Ocramius:
@SvenRtbg seems solved now, but a force-push of even a single changed byte would help.

Ref: https://www.githubstatus.com/incidents/qcvjkzcs7j74

lcobucci:
Folks, I'm away from a laptop for a few days. I glanced through the changes on the phone and all seems fine to me.

The introduction of the curve check makes sense, though it's a minor BC-break imho.

SvenRtbg:
@lcobucci Enjoy whatever makes you absent from your laptop! I'm aware about the BC break and will add documentation accordingly. @Ocramius added the 6.0.0 label already.

